PR #4, opened 2019-04-12
Pipeline Control Review and Bug Fix

ShawnLess:
When the pipeline stalled by MUL instruction and a remote load returns, the returned result is a load buffer. The forwarding logic in this case should consider the value in load buffer. This might not be the best solution but a workaround.


v/vanilla_bean/hobbit.v
```diff
@@ -54,7 +54,7 @@ module hobbit #(parameter
 
 //localparam trace_lp = 1'b1;
 localparam trace_lp = 1'b0;
-localparam debug_lp = 1'b0;
+localparam debug_lp = 1'b1;
```
vb000: might want to disable debug in the remote...

v/vanilla_bean/hobbit.v
```diff
@@ -1412,16 +1400,16 @@ if(debug_p | debug_lp) begin
...
+        end
+        //// Multiple-divide
+        //$fwrite(pelog, "X%0d_Y%0d.pelog   MD: stall_md=%b md_vlaid=%b md_resp_valid=%b md_result=%0x\n"
+        //         ,my_x_i
+        //         ,my_y_i
+        //         ,stall_md
+        //         ,md_valid
+        //         ,md_resp_valid
+        //         ,md_result
+        //        );
```
vb000: Sorry, I'll replace this with a neat trace logic when I start working on vanilla core. I think this kind of log is important for quick debug without the need for waveform.

tommydcjung: maybe we should have this in a separate module?

v/vanilla_bean/hobbit.v
```diff
@@ -848,11 +832,10 @@ assign id_s = '{
 always_ff @ (posedge clk_i)
 begin
 `ifdef bsg_FPU
-    if (reset_i | net_pc_write_cmd_idle |
-            (flush & (~(stall|fpi_inter.fam_depend_stall | depend_stall )))
-       )
+    if (reset_i | net_pc_write_cmd_idle | flush )
```
tommydcjung: I don't think we should have reset_i in the control logic path.

vb000: I think it's okay for synchronous resets.

v/vanilla_bean/hobbit.v
```diff
@@ -1261,78 +1246,79 @@ if(debug_p | debug_lp) begin
         $fwrite(pelog, "X%0d_Y%0d.pelog %0dns:\n", my_x_i, my_y_i, $time);
 
         // Fetch
-        $fwrite(pelog, "X%0d_Y%0d.pelog   IF: pc=%x instr=%x rd=%0d rs1=%0d rs2=%0d state=%b"
```
tommydcjung: we should have this logger thing in a separate module.

vb000: True if it's the trace. But we need the log for some internal signals to enable quick debug without waveform... :(

A quick diff of internal signal values with a working design can sometime point out the issue without any debug effort also!

vb000: Oh sorry, you mean by referencing internal signals from a module outside?... yeah that's a better way!

tommydcjung: yes, this logging is useful, but we should have a separate module to handle that.

ShawnLess:
Updated.

v/vanilla_bean/hobbit.v
tommydcjung: we can create a module called "pe_logger", which is non_synth, and it can be instantiated inside hobbit.

tommydcjung:
unrelated but I don't think we need jalr_prediction_rr. we are making prediction based on what's stored in jalr_prediction_r, only time jalr_prediction_r gets updated is when jalr is in EXE stage. So I think it's sufficient to compare jalr_addr with jalr_prediction_r to determine whether there was jalr mispredict or not.

taylor-bsg:
Usually _rr means time delay the register by one cycle. Is it trying to
align it in the pipeline?

M

vb000:
Yeah, that's to align in the pipeline. But apparently, it updates even when
there's a stall. The _rr register update is not controlled by usual
pipeline update control signals. Trying to understand it more clearly...

vb000:
In the best case, jalr prediction value and the jalr instruction that used
the prediction would be perfectly aligned. In the worst case, will stalls,
jalr prediction value would reach EXE stage faster than the jalr
instruction that used it. When the instruction eventually reaches EXE
stage, the jalr prediction value would safely be in the register to be
compared against.

There's no issue, but needs complicated analysis to prove that (added
an assembly
test case
<https://github.com/bespoke-silicon-group/bsg_manycore/commit/4d80abb022d764b779290d53e34eda7cbcb8928c>
to test this). In stead, wouldn't it be safer to include
jalr_prediction_r/_rr in pipeline stage registers to make sure the
alignment? Unless we absolutely need to use explicitly instantiated
hardened dffs here
<https://github.com/bespoke-silicon-group/bsg_manycore/blob/b6b2bc9fd14d15363123d22427420641b42a0f48/v/vanilla_bean/hobbit.v#L774>
...

PR #9, opened 2019-04-17
remove jalr_prediction_rr; fix the bug with byte addr comparison with…

tommydcjung:
… word addr for jalr_mispredict


tommydcjung:
coremark score
- before:911217
- after: 907907

PR #14, opened 2019-04-22
Merging dev to fix git history

drichmond:
Re-applying PR #10 after #13


proftaylor:
What was the issue with git history

drichmond:
When PR #10 was merged, it was squashed and merged into master. This means that the history of dev and master look different, and don’t have common ancestors. it makes it more difficult to root cause the issue we’re trying to debug.

In general - my theory of git is that it is best practice to squash and merge PRs into dev because they represent single features. Then merge without squashing from dev to master so you can debug and see the history

taylor-bsg:
So the point is that squashing into master is going to merge together
multiple dev squashes, creating a mega squash that erases the semantic
grouping that we had in dev and making it impossible to track down issues?

drichmond:
Yes, in short.

PR #28, opened 2019-05-12
Tracer

tommydcjung:


vb000:
Thanks Tommy, sorry to have bloated hobbit with all fprintfs! Interesting, I didn't know bind can access internal signals in a clean way! Should we give this a different name to distinguish this from actual hw trace? I used call these processor execution logs :p

taylor-bsg:
Seems good to factor the code, but we should have a greater discussion of what we want for tracing.  The tracer we need for tracking down programmer bugs and concurrency bugs should be simultaneously less verbose (to save disk space) and more informative (for the class of bugs we are tracking down).

I also noticed some of the ring stuff got removed — what is the status of the ring network? Which of the original features are supported?

tommydcjung:
the ring network interface was only being used for remote store into icache.

PR #31, opened 2019-05-21
Merging bsg_cuda_lite_runtime regression test for CUDA API into master

bornaehsani:
This pull request merges all cuda regression tests that are designed to run with the CUDA API on the bsg_f1.

Changes include: 
1. Added a bsg_cuda_lite_runtime.h header file, and moved all the routines into it. cuda regression tests and include the header and simply call it instead of carrying over a large code. (hats off to Max for the idea and his help!)
2. Added the following regression tests inside software/spmd/bsg_cuda_lite_runtime/ directory. Each test is in its separate directory, and all tests share the main.c file inside main directory. (the cuda lite runtime section of the code is shared among all tests).
3. Added the following symbols to bsg_set_tile_x_y.c/h
     - __bsg_tile_group_id_x/y 
     - __bsg_grid_dim_x/y
4. Added the following regression tests: 
     - empty: an empty kernel.
     - empty parallel: similar to empty. 
     - barrier: kernel with a barrier at the end.
     - packet: kernel that sends remote_ptr_io_store packets with the value of  __bsg_x/y, __bsg_id, __bsg_tile_group_id_x/y, __bsg_grp_org_x/y, CSR_TGO_X/Y, __bsg_grid_dim_x/y symbols.
     - stack_load: kernel that takes in 16 arguments (8 go through stack), and stores the sum in DRAM.
     - shared_mem: tiles declare tile group shared memory, store to and load from it with different striping algorithms, and store the sum into DRAM. Tests the tile group shared memory and barrier functionality.
     - shared_mem_load_store: tiles declare tile group shared memory, and load an MxN matrix from DRAM into it block by block, and store to a different location.
     - vec_add/ vec_add_parallel/ vec_add_parallel_multi_grid: tiles perform vector addition on an array in DRAM.
    - matrix_mul: matrix multiplication kernel without shared memory. 
    - matrix_mul_shared_mem: matrix multiplication kernel with tiling and shared memory. Not working.

These tests are not included in the bsg_manycore regression test, as they do not run in a stand alone mode. They are verified by cuda regression tests in the bsg_f1. However, all other regression tests have passed with the modifications to library.

Regression tests passed, rebased, and ready to merge.


drichmond:
Borna, what do we need to do to verify this PR? Or is it stale?

bornaehsani:
It is stale, thanks for pointing it out, I will have a newer version merged when the CUDA API is merged into F1.

PR #36, opened 2019-05-29
fix the dc crash issue

hema0730:


stdavids:
I would move the comment closer to the parameters themselves.

hema0730:
2018.06 is the latest version we have on aoi. The issue might be fixed in
the newer versions by Synopsys but I haven't got a chance to try it yet.

taylor-bsg:
I guessed as much; instead of “prior to”, instead say “at least up to”

PR #39, opened 2019-05-31
Added a Variable that allows cadenv to be ignored in Makefile.verilog

drichmond:
including bsg_cadenv is unnecessary for F1 Cosimulation. Cloning it isn't currently isn't part of the Bladerunner flow.

So, I've added a variable that allows VCS/DVE/Cadenv to be ignored.


drichmond:
@tommydcjung Please don't delete this branch when merged

drichmond:
Yeah, on F1 and in Cosim. You already approved the other one to that effect.

PR #48, opened 2019-06-04
Newlib/LittleFS and Specint

vb000:
- Interfacing code and Makefile flow to run programs with Newlib/LittleFS.
- Tests programs demonstrating Newlib.
- Makefiles to run Specint from raw's greenlight repo.
  - Tiny header to emulate rawlib
- Support for C command line arguments (argc, argv) with Newlib.
- Support for running single threaded programs on spike.
  - Along with patch for spike: with hack to print chars by writing to "tohost"
- Makefile rules to install Newlib and Spike


drichmond:
My only concern is that there is a huge number of commits and merges, and some of those commits have already been merged into master. This causes an unnecessarily large git history - which is a problem when reading the git history for debugging.

In general you should avoid merges in branches. If a branch has a feature you want, you should rebase on it. 

A simple fix would be to squash + merge, but this removes all the git history. A more complex solution is to rebase your changes onto master, removing all the merge commits but forcing you to go through all the merge conflicts by hand.

vb000:
Should I rebase then? I wouldn't want to squash all that work into single commit!

If I do a plain merge, would others' commits I pulled in also get added as new commits? Other than those there are 12 new merge commits the merge would add. Would that be too bad for the history?

PR #51, opened 2019-06-06
Auipc bug

vb000:
A concise test for verifying PC relative DRAM addressing.

### DIFF
--- a/software/spmd/common/asm.ld
+++ b/software/spmd/common/asm.ld
@@ -18,6 +18,7 @@ SECTIONS
   _bsg_dram_start_addr = 0x80000000;
   .text 0x80000000 :  {
           *(.text)
+          *(.dram)
    }
   _bsg_dram_end_addr = .  ;
 }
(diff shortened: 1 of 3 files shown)


software/spmd/common/asm.ld
```diff
@@ -18,6 +18,7 @@ SECTIONS
   _bsg_dram_start_addr = 0x80000000;
   .text 0x80000000 :  {
           *(.text)
+          *(.dram)
```
max-ruttenberg: what does adding this line do? to me it looks like it places .dram after .text, but wasn't that happening anyway?

vb000: To reproduce the bug in smallest possible test case, I chose to write an assembly program. But the assembly linker, spmd/common/asm.ld, didn't include the section for .dram section. Hence I added this to mimic that of current C linker.

max-ruttenberg: Great!

PR #55, opened 2019-06-12
New link script

vb000:
New link script with updated memory layout:
- Separate I-EVA and D-EVA address spaces:
  - I-EVA: 
    - `0x00000000 - 0x00ffffff`
  - D-EVA:
    - `0x00001000 - 0x00001fff` (DMEM)
    - `0x80000000 - 0x80ffffff` (I-cache backing storage in DRAM)
    - `0x81000000 - 0xffffffff` (DRAM data)
- Loader changes to load instructions and data separately to greatly reduce load time.
- Assembly programs can also use the same linker file instead of a separate one.
- Regression passed.
- TODO: Temporarily disabled section overlap checks because I-EVA & D-EVA both start at 0x0 and and a clean way to tell linker to consider them separate may not be available.


max-ruttenberg:
Do all programs in SPMD still compile and run in VCS with this PR?

vb000:
> Do all programs in SPMD still compile and run in VCS with this PR?

All the programs in the regression list, yes.

vb000:
More changes:
- Removed dead code in `crt.S`
- Removed unnecessary files in `spmd/common`
- Added an additional link script for linking data in DRAM by default. Useful for running large programs like benchmarks.
- Renamed link scripts to `link_dmem.ld` and `link_dram.ld`

max-ruttenberg:
All the linker script stuff looks OK to me. @drichmond this will break the regressions in bsg_f1 because the loader needs to be updated. 
 @vb000 I don't totally understand the crt changes in this PR. What's the the story with those? Why are they included in this PR?

vb000:
@max-ruttenberg The crt was initially imported riscv-tests repo. It has lot of redundant stuff. So cleaned it up. Another notable change is, I defined a symbol called `_sp` in linker which is loaded to initialize stack pointer instead of hardcoding it in the crt.

bornaehsani:
@max-ruttenberg no it's not! don't know what's going on, but cuda tests are broken now!

bornaehsani:
Problem solved!

PR #56, opened 2019-06-12
Added a floating point vector addition regression test

bornaehsani:


drichmond:
What are we doing with this PR?

bornaehsani:
I left it here so Tommy would have it merged in when he feels comfortable, since he was in the process of breaking the large fpu commit into smaller ones on a separate branch. if the fpu branch is stable now I can rebase and do another pull request. @tommydcjung what do you think?

tommydcjung:
if it's just a vector addition, we already have a test that's covering that (float_vec), but we always want more testing of course. maybe matrix multiply and 2d convolution, etc

PR #58, opened 2019-06-12
fpu bsg_manycore_proc_vanilla; network_rx/tx

tommydcjung:


vb000:
Hi @tommydcjung! I'm feeling difficulty understanding this without the the knowledge of vanilla core changes. I think it'd be easier to start with a PR just for vanilla core. What do you think?

PR #65, opened 2019-06-19
Added CUDA runtime symbols that are set by host's CUDA API

bornaehsani:
This PR addresses issue #175 on bsg_f1.
Introduced 6 new symbols in the bsg_cuda_lite_runtime.h header file, cuda_kernel_ptr, cuda_argc, cuda_argv_ptr, cuda_finish_signal_addrl, cuda_finish_signal_val and cuda_kernel_not_loaded_val. The symbols are set by the CUDA API after binary is loaded onto tiles. 

**Should only be merged in together with PR #298 on bsg_f1**


drichmond:
@proftaylor Is this OK to merge, and then we branch the two versions from it?

PR #69, opened 2019-06-21
Beebs

vb000:
- Added Makefile flow to run beebs
- Added beebs to regression
- Refactored regression:
  - Separated test lists into groups:
     - Group 0: `rv32ui`
     - Group 1: `rv32um`
     - Group 2: `spmd`
     - Group 3: `coremark`
     - Group 4: `beebs`
     - `make GROUP=<group>` in `software/spmd/` to run individual group
  - Option to run different levels of regression: Level 0 => group 0, Level 1 => group 0, 1...
    - `make LEVEL=<level>` in `software/spmd/` to run different levels of regression.


tommydcjung:
suggestion: I think we should have separate folders for programs with different compilation process (newlib, clang, beebs). right now, we are just tacking on additional stuff onto the existing infrastructure (makefile.builddef). It's not scalable, and it will only get worse. I think this will help us work independently without stepping onto each other's toe.

@soctar

taylor-bsg:
Let’s get stuff into master; tommy perhaps you can file an issue for when we have more time..

tommydcjung:
might wanna add back those tests.

test_case_list+=instr_tests/auipc_rv32
test_case_list+=instr_tests/branch_rv32

PR #75, opened 2019-06-22
[beebs] Recompilation rules for individual runs

vb000:


vb000:
@proftaylor Can we have this merged in and rebased on?

taylor-bsg:
Hi, Bandhav; nice! I merged these changes in, with a few tweaks so that they work for the top-level regression.

PR #76, opened 2019-06-23
Temporary changes for conveniently running rv32uf tests

vb000:
`make old GROUP=rv32uf` for floating point riscv tests

### DIFF
--- a/software/spmd/Makefile
+++ b/software/spmd/Makefile
@@ -40,48 +40,49 @@ all:
 	@echo " Run 'top' to see what else is running to decided how many threads to run"
 
 # mbt: saving this code so that bandhav can scavenge the coverage stuff
-#
-#
-# all: main.regress_test
-# ##############################################
-# #  regression test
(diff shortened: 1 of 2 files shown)


taylor-bsg:
Can you run make RV_TEST_ISA_VARIANTS=rv32uf in bsg_riscv_tests; or if you want, add a rule to that effect to that Makefile?

vb000:
I can, but that floods the terminal with logs and there's no easy way to tell which test is timing out and stalling the regression. Anyway, I'd close this pr and find a way locally for the time being.

taylor-bsg:
Okay, I pushed some mods. The scripts now also capture standard error output to the logs, which helps with compilation logs. Make watch-summary in spmd lives you a live update of which tests timed out, which tests FAIL'd or ERROR'd, and how many SIMV processes are still running. In terms of stalling the regression, it helps to run with -j <threads> so that you have some fault tolerance.

PR #83, opened 2019-06-26
Pipelined reset parameters and reset register fanout.

stdavids:
Adding pipeline reset depth parameters to bsg_manycore and bsg_manycore_link_async_to_wormhole. Adding reset register fanout to all tiles inside of bsg_manycore.


tommydcjung:
lgtm

PR #93, opened 2019-07-08
CUDA Lite runtime dram regression tests refactored

bornaehsani:
Added three new cuda lite runtime dram regression tests.
Related to PR # 336 in bsg_f1.


drichmond:
Does this have the FPU test too? I'm confused why that commit is here

Issue #102, opened 2019-07-22
Change all bsg_global_X/Y and similar symbol assignments to conditional (if not defined before)


bornaehsani:
This is necessary for the Makefiles in bsg_f1 to be able to overwrite these assignments.

https://github.com/bespoke-silicon-group/bsg_manycore/blob/c84d72fab92eaff82f29047de40c13c8536410f1/software/spmd/bsg_cuda_lite_runtime/barrier/Makefile#L7

bornaehsani:
Addressed in PR #103 .

PR #103, opened 2019-07-23
Bug fix in cuda lite runtime regression tests Makefiles

bornaehsani:
Changed default assignment to bsg_global_X/Y, bsg_tiles_X/Y and bsg_grp_org_X/Y in bsg_cuda_lite_runtime kernel Makefiles, so that bsg_f1 Makefile can overwrite symbol values for comipilng the kernel binary.
Related to issue #102.


drichmond:
Does this work in Cosim? Does it resolve Shady's issue?

PR #104, opened 2019-07-23
Refactor 20190723

tommydcjung:
- nbf.py support power of 2 banks
- credit source-limiting spmd loader (fix credit underflow issue)
- add vcache_sets_p parameter to calculate which bit to XOR in hash function (bank of 9)


tommydcjung:
comments are good, but is there anything in specific you want added?

PR #112, opened 2019-07-26
Refactor tile group shared memory to support floating point

bornaehsani:
1. Modified bsg_tile_group_shared_load/store macros to also take in the type (int or float) as input argument. 
2. Added a bsg_tile_group_remote_ptr macro that translates to a bsg_remote_int/float_ptr based on the given argument. 
3. Added a floating point vector addition with tile group shared memory cuda lite runtime regression test.

### DIFF
--- a/software/bsg_manycore_lib/bsg_manycore.h
+++ b/software/bsg_manycore_lib/bsg_manycore.h
@@ -29,6 +29,7 @@ typedef volatile void *bsg_remote_void_ptr;
 
 #define bsg_tile_group_shared_mem(type,lc_sh,size) type lc_sh[((size + ((bsg_tiles_X * bsg_tiles_Y) -1))/(bsg_tiles_X * bsg_tiles_Y))]
 #define bsg_tile_group_shared_load(type,lc_sh,index,val) (  (val) = *(bsg_tile_group_shared_ptr(type,(lc_sh),(index)))	)
+#define bsg_tile_group_shared_load_direct(type,lc_sh,index) (*(bsg_tile_group_shared_ptr(type,(lc_sh),(index))))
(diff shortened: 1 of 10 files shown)


bornaehsani:
@cylinbao I wanted you to see this change, the bsg_tile_group_shared_load/store now takes an extra argument for the type of the input (int/float). With this change the problem that we talked about might be solved, but I wanted to know your opinion on this change. Is it hard to implement this change in tvm's codegen IR or is it too hard?

software/bsg_manycore_lib/bsg_manycore.h
```diff
@@ -28,8 +28,8 @@ typedef volatile void *bsg_remote_void_ptr;
 #define bsg_host_dram_load(addr, val) do { val = *(bsg_host_dram_ptr((addr)));} while (0)
 
 #define bsg_tile_group_shared_mem(type,lc_sh,size) type lc_sh[((size + ((bsg_tiles_X * bsg_tiles_Y) -1))/(bsg_tiles_X * bsg_tiles_Y))]
-#define bsg_tile_group_shared_load(lc_sh,index,val) (  (val) = *(bsg_tile_group_shared_ptr((lc_sh),(index)))	)
-#define bsg_tile_group_shared_store(lc_sh,index,val) (  *(bsg_tile_group_shared_ptr((lc_sh),(index))) = (val)	)
+#define bsg_tile_group_shared_load(type,lc_sh,index,val) (  (val) = *(bsg_tile_group_shared_ptr(type,(lc_sh),(index)))	)
```
cylinbao: It will be helpful for code generation in tvm if you can add a shared_load with two inputs.
It could be like this.
#define bsg_tile_group_shared_load_direct(type,lc_sh,index) (  *(bsg_tile_group_shared_ptr(type,(lc_sh),(index)))	)

bornaehsani: Thanks Chien-Yu! How does it look now?

PR #113, opened 2019-07-29
Toolchain updates

vb000:
- Added `f` extension to target ISA for riscv-gnu-toolchain installation.
- Added `install-clean` and `install-with-builds` to install tools without and with builds preserved respectively.
- Updated readme.

### DIFF
--- a/software/riscv-tools/Makefile
+++ b/software/riscv-tools/Makefile
@@ -105,7 +105,7 @@ configure-riscv-gnu-tools:
 	@echo "Configuring toolchain..."
 	@echo "====================================="
 	cd riscv-gnu-toolchain && \
-		./configure --prefix=$(RISCV) --disable-linux --with-arch=rv32ima --with-abi=ilp32
+		./configure --prefix=$(RISCV) --disable-linux --with-arch=rv32imaf --with-abi=ilp32
 
 build-riscv-gnu-tools: configure-riscv-gnu-tools
 	@echo "====================================="
--- a/software/riscv-tools/Makefile
+++ b/software/riscv-tools/Makefile
@@ -25,6 +25,8 @@
 # assumed already installed: autoconf automake libtool curl gmp gawk bison flex texinfo gperf gcc48 gsed
 #
 #
+.DEFAULT_GOAL := help
+
 TOOLCHAIN_REPO=riscv-gnu-toolchain
 TOOLCHAIN_URL=https://github.com/riscv/$(TOOLCHAIN_REPO)
 DEPENDS_DIR=$(CURDIR)/depends
@@ -51,7 +53,8 @@ export PATH
 export SHELL:=$(SHELL)
 
 
-nothing:
+help:
+	@cat README.md
 
 checkout-repos:
 	mkdir -p $(DEPENDS_DIR);	
@@ -94,7 +97,7 @@ build-deps:
 	for dep_tool in $(DEPENDS_LIST); do 	\
 		tar xf $$dep_tool.tar.xz &&      \
 		cd $$dep_tool	&&	\
-		./configure --prefix=$(DEPENDS_DIR) && make && make install; \
+		./configure --prefix=$(DEPENDS_DIR) && $(MAKE) && $(MAKE) install; \
 		cd ..;	\
 	done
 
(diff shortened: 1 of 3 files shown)


software/riscv-tools/Makefile
```diff
@@ -103,7 +103,7 @@ configure-riscv-gnu-tools:
 	@echo "Configuring toolchain..."
 	@echo "====================================="
 	cd riscv-gnu-toolchain && \
-		./configure --prefix=$(RISCV) --disable-linux --with-arch=rv32ima --with-abi=ilp32
+		./configure --prefix=$(RISCV) --disable-linux --with-arch=rv32imaf --with-abi=ilp32
```
drichmond: Have we been compiling RISC-V Binaries without Floating Point Instructions?

software/riscv-tools/README.md
```diff
@@ -1,3 +1,18 @@
-Typically you will do
-make checkout-all 
-make build-all
+Clean installation of tools:
```
drichmond: I think this is really helpful!

One thing you can also do is put a `help` rule in the top-level makefile and make it the default rule. That way, when people run 'make`, they are automatically given the help prompt. 

Just a thought - you can do it if you'd like, or keep it this way.

software/riscv-tools/Makefile
```diff
@@ -124,13 +124,17 @@ rebuild-newlib:
 	make -C $(TOOLCHAIN_REPO)/build-newlib
 	make -C $(TOOLCHAIN_REPO)/build-newlib install
 
-clean:
-	rm -rf riscv-install
+install-with-builds: checkout-all build-all
+install-clean: install-with-builds clean-builds
```
drichmond: This will be problematic if someone tries to run a parallel build (e.g. -j8). It will run both `install-with-builds` and `clean-builds`.

I don't think anyone runs the build in parallel at the moment. Have you tried?

vb000: This makefile is not intended to run in parallel... this makefile runs the actual riscv toolchain build in parallel though.

But yeah, this is an issue. I'll add a stamp to ensure clean runs strictly after installation! Thanks :)

taylor-bsg:
@tommydcjung @vb000 Aren't there additional flags that need to be provided to the compiler to prevent floating point divides and other unsupported instructions?

taylor-bsg:
@dpetrisko Dan, linking you in. We are modifying the HB manycore RISC-V tools build so that it does a clean after it installs the binaries. This saves about 9 GB per RISC-V tool install across all of our users; Bandhav says RISC-V tools only take about 100 MB after this is done. Can you push into BlackParrot build as well?  You can also offer feedback on Bandhav's implementation here.

dpetrisko:
Yeah, we’ve implemented this in BP by doing git submodule deinit -f to all of the tool build directories (gnu toolchain especially). IIRC, there was some minor installation location problem with Verilator and panicroom, so we may have taken it out, not sure. 

I can put it back in the build once I test it again

dpetrisko:
General feedback is that riscv-tools is super deprecated, basically only used for RocketChip. 

For BlackParrot, we submodule the tools we need directly. Our build script does work with -j, which saves a ton of time

taylor-bsg:
Yep, Dan looks like both you, Farzam, and Mark still have very large risc-v
build directories.

taylor-bsg:
@dpetrisko @vb000 Total space taken on xor's 800 GB SSD for riscv-tools across all users:   234 GB.

Is there a simple command that we can have people type in all of the legacy riscv-gnu-toolchain directories to remove the build files but leave the install files alone?

software/riscv-tools/Makefile
drichmond: So we run our AMI build in parallel from the top-level Makefile. That makefile calls this makefile. This would be an issue for us (I think)

taylor-bsg:
Bandhav, to be more prescriptive, you can move the build-all from the
dependence part of install-with-builds to the rule part, using $(MAKE)
build-all. Similarly for install-clean and clean-builds.

dpetrisko:
Added a target to clean tools in BP, down to ~1GB for all tools.

software/riscv-tools/Makefile
vb000: Got it! I'll fix this...

vb000:
> @tommydcjung @vb000 Aren't there additional flags that need to be provided to the compiler to prevent floating point divides and other unsupported instructions?

They are provided compile time.

vb000:
> Bandhav, to be more prescriptive, you can move the build-all from the dependence part of install-with-builds to the rule part, using $(MAKE) build-all. Similarly for install-clean and clean-builds.

Great, will do! Thanks!

vb000:
> @dpetrisko @vb000 Total space taken on xor's 800 GB SSD for riscv-tools across all users: 234 GB.
> 
> Is there a simple command that we can have people type in all of the legacy riscv-gnu-toolchain directories to remove the build files but leave the install files alone?

`make clean` in riscv-gnu-toolchain clone. The breakout of 9.5G is 4G repository and 5.5G build... so the changes I made removes source repo as well after installation.

proftaylor:
So that would only delete the build files, right? Would that also blow away
their binaries though? I am talking about older checkouts where they may
not want to update to latest.

dpetrisko:
make clean should not remove the installed binaries/libraries

vb000:
> So that would only delete the build files, right? Would that also blow away their binaries though? I am talking about older checkouts where they may not want to update to latest.

Binaries are installed in `software/riscv-tools/riscv-install`. So `rm -rf software/riscv-tools/riscv-gnu-toochain` would be good enough!

taylor-bsg:
Okay great — thanks!

drichmond:
@taylor-bsg Why do you want a series of $(MAKE) commands in the rule, but not dependencies?

dpetrisko:
My understanding is that a series of MAKEs will execute serially with -j, whereas dependencies will execute in parallel. (Each individual MAKE will receive the -j, but finish before the next MAKE)

That’s why I do it, anyway

drichmond:
Yes. Dependencies will execute in parallel, provided they do not have any common dependencies between them.

- If A depends on B and C and B depends on C, C will execute before B. 
- If A depends on B and C and B and C depend on D, D will execute once

software/riscv-tools/Makefile
```diff
@@ -109,23 +112,28 @@ build-riscv-gnu-tools: configure-riscv-gnu-tools
 	@echo "====================================="
 	@echo "Building toolchain..."
 	@echo "====================================="
-	cd riscv-gnu-toolchain && make -j 16 newlib && make install
+	cd riscv-gnu-toolchain && $(MAKE) -j 16 newlib && $(MAKE) install
```
dpetrisko: Somewhat a style choice, but 
`$(MAKE) -C newlib install`
is more concise

Issue #114, opened 2019-07-29
striped_hello time out


tommydcjung:
striped_hello times out for 4x4, but it works fine for 9x15.

tommydcjung:
however, the test passes when the tile group is placed elsewhere (e.g. x=1,y=1)

tommydcjung:
```
F P P x
F F P x
F F P x
x x x x

F = fail
P = Pass
x = N/A
```

tommydcjung:
```
P P P P P P P P x
P ? ? ? ? ? ? ? x
P ? ? ? ? ? ? ? x
P ? ? ? ? ? ? ? x
P ? ? ? ? ? ? ? x
P ? ? ? ? ? ? ? x
P ? ? ? ? ? ? ? x
F ? ? ? ? ? ? ? x
F ? ? ? ? ? ? ? x
F ? ? ? ? ? ? ? x
F ? ? ? ? ? ? ? x
F ? ? ? ? ? ? ? x
F F F P P ? ? ? x
F F F F P P P P x
x x x x x x x x x

?=haven't tried

**Comment by tommydcjung** [comment]
https://github.com/bespoke-silicon-group/bsg_manycore/blob/debug_20190731/software/spmd/striped_minimal/main.c

**Comment by tommydcjung** [comment]
https://github.com/bespoke-silicon-group/bsg_manycore/blob/debug_20190731/software/spmd/striped_minimal/main.c#L15
```
while (A[i][id] != (id * 2) + i * 5);
```
this line is supposed to check if the other tile has written to the striped array.
this corresponds to the instruction with PC=dcc.
```
     dc8: 00159f13            slli  x30,x11,0x1
     dcc: 00062603            lw  x12,0(x12)
     dd0: 19e61a63            bne x12,x30,f64 <main+0x4c8>
```

then it jumps to pc=f64, which is just infinite loop. This is incorrect, because the program is supposed to keep loading the value until it gets the value stored by the other tile. So I suspected that something is wrong with the compilation.


```
     f5c: aca5a823            sw  x10,-1328(x11) # 4000ead0 <_bsg_elf_vcache_size+0x3ffeead0>
     f60: 0000006f            j f60 <main+0x4c4>
     f64: 0000006f            j f64 <main+0x4c8>
```

solution: change the gcc flag from -O2 to -O0.

tommydcjung:
might want to figure out what -O flag we want to use for each type of spmd program.

[...]

PR #124, opened 2019-08-02
Volatile load bugfix

neildryan:
This fixes the issue with the `striped_hello` spmd test timing out (#114) by making the LLVM pass respect the programmer's use of `volatile`, as well as fixing an issue where the test wasn't agnostic to manycore size.

### DIFF
--- a/software/bsg_manycore_lib/bsg_tilegroup.h
+++ b/software/bsg_manycore_lib/bsg_tilegroup.h
@@ -80,7 +80,6 @@ void extern_store_int(int STRIPE *arr_ptr, unsigned elem_size, unsigned offset, 
             (unsigned) arr_ptr, elem_size, offset, val);
 #endif
     int *ptr = get_ptr_val(arr_ptr, elem_size, offset);
-    bsg_printf("store %x %d\n", arr_ptr, __bsg_id);
     *ptr = val;
 }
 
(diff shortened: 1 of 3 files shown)


software/bsg_manycore_lib/bsg_tilegroup.h
```diff
@@ -104,42 +104,62 @@ void extern_store_char(int STRIPE *arr_ptr, unsigned elem_size, unsigned offset,
 }
 
 
-float extern_load_float(int STRIPE *arr_ptr, unsigned elem_size, unsigned offset) {
+float extern_load_float(int STRIPE *arr_ptr, unsigned elem_size, unsigned offset,
```
tommydcjung: just for my understanding, when is extern_load_float used?

neildryan: When loading a float from another tile -- it's effectively the same as extern_load_int, but clang has much stricter typing requirements than GCC

tommydcjung: okay. so are you saying that calling "A[i][id]" gets translated into extern_load_<type>() by the compiler (in layman's term)?

neildryan: If A[i][id] is an rvalue, then yes! If it's an lvalue, A[i][id] translates into extern_store_() by the LLVM pass.

software/spmd/striped_hello/main.c
```diff
@@ -127,10 +128,10 @@ int main()
 
     int bsg_id = bsg_x * bsg_tiles_X + bsg_y;
```
tommydcjung: BTW, shouldn't it be something like int bsg_id = bsg_Y * bsg_tiles_X + bsg_x;

i think bsg_Y, bsg_X are deprecated (replaced by __bsg_Y or something).
I would refer to bsg_set_tiles_x_y.h. I think it also calculate __bsg_id

neildryan: I'll change the test in another commit

PR #128, opened 2019-08-12
separate jump -> jal/jalr

tommydcjung:


v/vanilla_bean/vanilla_core.v
```diff
@@ -16,11 +16,11 @@ module vanilla_core
     , parameter x_cord_width_p="inv"
     , parameter y_cord_width_p="inv"
 
-    , localparam dmem_addr_width_lp=`BSG_SAFE_CLOG2(dmem_size_p)
-    , localparam icache_addr_width_lp=`BSG_SAFE_CLOG2(icache_entries_p)
-    , localparam pc_width_lp=(icache_tag_width_p+icache_addr_width_lp)
-    , localparam reg_addr_width_lp = RV32_reg_addr_width_gp
-    , localparam data_mask_width_lp=(data_width_p>>3)
+    , parameter dmem_addr_width_lp=`BSG_SAFE_CLOG2(dmem_size_p)
```
vb000: A minor detail... shouldn't these be `_p` if these are parameters? A simple vim find and replace should do I think...

tommydcjung: no, the derived parameters are _lp. some tools don't like localparam in module signature.

PR #129, opened 2019-08-13
REFINE the PC increment logic; refactor icache interface

tommydcjung:


v/vanilla_bean/icache.v
```diff
@@ -11,200 +14,231 @@ module icache
...
-    , input [RV32_instr_width_gp-1:0] icache_w_instr_i
-    , output[RV32_instr_width_gp-1:0] instruction_o
+    // ctrl signal
+    , input v_i
+    , input w_i
+    , input flush_i
+
+    // icache write
+    , input [icache_addr_width_lp-1:0] w_addr_i
+    , input [icache_tag_width_p-1:0]  w_tag_i
```
vb000: Isn't it better to hide icache implementation detail (tag and addr widths) by just having something like `w_pc_i` (instead of `w_addr_i` and `w_tag_i`) as input?

PR #132, opened 2019-08-20
Removed spmd_testbench.v from sv.include so that the file list can be used in F1

drichmond:
* Replaced in Makefile


tommydcjung:
this list still contains files that are not used in bsg_f1. maybe you can use the option "-top" when compiling?

drichmond:
You’re absolutely right on both counts. If multiple top levels were the issue then your suggestion would fix it. And yes, there are files in this file list that aren’t used in F1

The issue I’m trying to solve is undefined macros when compiling spmd_testbench. We don’t define the parameter macros on the command line for F1, so compilation fails.

The extra files aren’t simulated if they aren’t used, and compiling verilog files is relatively quick compared to simulating them.

PR #138, opened 2019-08-28
Machine build refactor and EPA for stderr

vb000:
- Refactored machines build flow to automatically rebuild when Verilog source changes.
- **Mapped EPA=`0xEEE0` to stderr to have separate streams for program output and trace**. Also, Newlib std streams could be separated.
- New bsg_manycore_lib macro `bsg_putchar_err` to write char to stderr stream.


tommydcjung:
Refactoring machines build flow seems like an orthogonal change from the other changes. So please separate those two.

Besides, we are using those filelists (core.include,  tb.include, etc) in bsg_f1 as they were. Also, the filenames change seems a bit uncessary?

mrutt92:
I agree with @tommydcjung that these changes should be in two different PRs.

vb000:
I agree! But, machines refactor is a minor change that I needed to make, to make the developemnt of this verilog change easy. 

Cool, I can separate it into a different pr though.

vb000:
> EPA 0xEAE0 overloads the address designated for `bsg_print_int()`. Please look at `software/bsg_manycore_lib/bsg_manycore.h` and choose an EPA not already in use.

Done! Updated it to `0xEEE0`

drichmond:
I like EEE

tommydcjung:
make sure to update the document on the change in EPA mapping

PR #139, opened 2019-08-29
Machines build refactor

vb000:
Refactored machines build flow to automatically rebuild when Verilog source changes.


tommydcjung:
I don't think this is a critical feature. besides what happens if you set the machine dir to somewhere outside bsg_manycore repo? (rtl_hard in bsg_designs)

vb000:
@tommydcjung Previous way, every verilog update, we need to build in machines dir and then run the testcase. I think two steps to run the test after every mod is unnecessary. This pr addresses that inconvenience.

External machines path: thanks for pointing it out, I'll fix that!

machines/Makefile.include
```diff
@@ -0,0 +1,14 @@
+include core_srcs.mk
+include sim_srcs.mk
+include tb_srcs.mk
+
+# Order is important!
+VCS_SRCS := $(CORE_SRCS) $(SIM_SRCS) $(TB_SRCS)
```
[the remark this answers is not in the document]
vb000: Hi Dustin! Aren't variables `CORE_SRCS`, `SIM_SRCS` and `TB_SRCS` doing similar job? Those make variable could be used for different tool flow...?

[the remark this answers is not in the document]
drichmond: What I'm suggesting is that we standardize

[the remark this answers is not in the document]
drichmond: I think this is more important for VCS_INCLUDES and other variabls. VCS_SRCS I'd suggest changing to VSRCS or VSOURCES (like CSOURCES, or CXXSOURCES)

tommydcjung:
can we not change the name of the files? it could create merge conflict for other branches waiting to be merged in.

vb000:
> can we not change the name of the files? it could create merge conflict for other branches waiting to be merged in.

@tommydcjung Later branches rebase on master before merging. So wouldn't be a problem.

machines/Makefile.include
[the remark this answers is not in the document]
vb000: @drichmond Done! Have a look at the latest commit

machines/Makefile
```diff
@@ -1,50 +1,44 @@
-find-dir-with = $(shell /usr/bin/perl -e 'chomp($$_ = `pwd`); while ($$_ ne "" && ! -e "$$_/$(1)") { m:(.*)/[^/]+/??:; $$_ = $$1; } print;')
-
-ifndef BSG_MANYCORE_DIR
-  export BSG_MANYCORE_DIR := $(call find-dir-with,.BSG_MANYCORE_ROOT)
-endif
+.DEFAULT_GOAL = all
 
+export BSG_MANYCORE_DIR := $(shell git rev-parse --show-toplevel)
```
drichmond: I like this. But what happens if bsg_manycore is inside of bsg_bladerunner?

machines/Makefile
```diff
@@ -1,50 +1,44 @@
...
-	$(eval VCS_DEFINE := +define+BSG_MACHINE_GLOBAL_X=${BSG_MACHINE_GLOBAL_X})
-	$(eval VCS_DEFINE += +define+BSG_MACHINE_GLOBAL_Y=${BSG_MACHINE_GLOBAL_Y})
-	$(eval VCS_DEFINE += +define+BSG_MACHINE_VCACHE_SET=${BSG_MACHINE_VCACHE_SET})
-	$(eval VCS_DEFINE += +define+BSG_MACHINE_VCACHE_WAY=${BSG_MACHINE_VCACHE_WAY})
-	$(eval VCS_DEFINE += +define+BSG_MACHINE_VCACHE_BLOCK_SIZE_WORDS=${BSG_MACHINE_VCACHE_BLOCK_SIZE_WORDS})
-	$(eval VCS_DEFINE += +define+BSG_MACHINE_DRAM_SIZE_WORDS=${BSG_MACHINE_DRAM_SIZE_WORDS})
-	$(eval VCS_DEFINE += +define+BSG_MACHINE_DRAM_INCLUDED=${BSG_MACHINE_DRAM_INCLUDED})
-	$(eval VCS_DEFINE += +define+BSG_MACHINE_MAX_EPA_WIDTH=${BSG_MACHINE_MAX_EPA_WIDTH})
-	$(eval VCS_DEFINE += +define+BSG_MACHINE_MEM_CFG=${BSG_MACHINE_MEM_CFG})
+	$(eval VCS_MACRO_DEFS := +define+BSG_MACHINE_GLOBAL_X=${BSG_MACHINE_GLOBAL_X})
```
drichmond: Closer! What I meant was something like: 
```
	VDEFINES += BSG_MACHINE_GLOBAL_Y=${BSG_MACHINE_GLOBAL_Y})
	VDEFINES += BSG_MACHINE_VCACHE_SET=${BSG_MACHINE_VCACHE_SET})
        # etc...
  
       VCS_DEFINES +: $(foreach def,$(VDEFINES),+define+"$(def)")

```

machines/arch_filelist.mk
```diff
@@ -0,0 +1,118 @@
...
+# used in simulation. This has NOT been used in tapeout or any tapeout related
+# activities beyond simulation
+
+ARCH_VINCDIRS =$(BASEJUMP_STL_DIR)/bsg_misc
+ARCH_VINCDIRS+=$(BASEJUMP_STL_DIR)/bsg_cache
+ARCH_VINCDIRS+=$(BASEJUMP_STL_DIR)/bsg_noc
+ARCH_VINCDIRS+=$(BSG_MANYCORE_DIR)/v
+ARCH_VINCDIRS+=$(BSG_MANYCORE_DIR)/v/vanilla_bean
+
+ARCH_VDEFINES =$(BASEJUMP_STL_DIR)/bsg_misc/bsg_defines.v
```
drichmond: What do you mean by ARCH_VDEFINES? Do you mean VSOURCES instead?

machines/Makefile.include
```diff
@@ -0,0 +1,13 @@
+include arch_filelist.mk
+include sim_filelist.mk
+include tb_filelist.mk
+
+VDEFINES  := $(ARCH_VDEFINES) $(SIM_VDEFINES) $(TB_VDEFINES)
+VSOURCES  := $(ARCH_VSOURCES) $(SIM_VSOURCES) $(TB_VSOURCES)
+VINCDIRS  := $(ARCH_VINCDIRS) $(SIM_VINCDIRS) $(TB_VINCDIRS)
+VINCLUDES := $(foreach dir, $(VINCDIRS), $(shell find $(dir) -name '*.vh'))
```
drichmond: Why don't we use VINCLUDES for directories (Since INCLUDES / CINCLUDES is usually directories) and VINCSRCS for .vh files.

I think you can also have VINCLUDES as a prerequisite, and then when a file in that directory changes the makefile will rebuild.

machines/Makefile
```diff
@@ -1,50 +1,44 @@
...
+	$(eval VCS_MACRO_DEFS += +define+BSG_MACHINE_VCACHE_BLOCK_SIZE_WORDS=${BSG_MACHINE_VCACHE_BLOCK_SIZE_WORDS})
+	$(eval VCS_MACRO_DEFS += +define+BSG_MACHINE_DRAM_SIZE_WORDS=${BSG_MACHINE_DRAM_SIZE_WORDS})
+	$(eval VCS_MACRO_DEFS += +define+BSG_MACHINE_DRAM_INCLUDED=${BSG_MACHINE_DRAM_INCLUDED})
+	$(eval VCS_MACRO_DEFS += +define+BSG_MACHINE_MAX_EPA_WIDTH=${BSG_MACHINE_MAX_EPA_WIDTH})
+	$(eval VCS_MACRO_DEFS += +define+BSG_MACHINE_MEM_CFG=${BSG_MACHINE_MEM_CFG})
 	$(VCS) +v2k -sverilog -full64 -o $*/simv \
 		+lint=all,noSVA-UA,noSVA-NSVU,noVCDE,noNS \
 		-timescale=1ps/1ps \
-		-f tb.include $(VCS_INCDIR) $(VCS_DEFINE) | tee -i $*/build.log
+		$(VCS_INCDIRS) $(VCS_DEFINES) $(VCS_SOURCES) $(VCS_MACRO_DEFS) | tee -i $*/build.log
```
tommydcjung: this should be called VCS_DEFINE

drichmond: Why?

machines/arch_filelist.mk
```diff
@@ -0,0 +1,118 @@
...
+ARCH_VSOURCES+=$(BASEJUMP_STL_DIR)/bsg_misc/bsg_circular_ptr.v
+ARCH_VSOURCES+=$(BASEJUMP_STL_DIR)/bsg_misc/bsg_imul_iterative.v
+ARCH_VSOURCES+=$(BASEJUMP_STL_DIR)/bsg_misc/bsg_idiv_iterative.v
+ARCH_VSOURCES+=$(BASEJUMP_STL_DIR)/bsg_misc/bsg_idiv_iterative_controller.v
+ARCH_VSOURCES+=$(BASEJUMP_STL_DIR)/bsg_misc/bsg_buf.v
+ARCH_VSOURCES+=$(BASEJUMP_STL_DIR)/bsg_misc/bsg_buf_ctrl.v
+ARCH_VSOURCES+=$(BASEJUMP_STL_DIR)/bsg_misc/bsg_xnor.v
+ARCH_VSOURCES+=$(BASEJUMP_STL_DIR)/bsg_misc/bsg_nor2.v
+ARCH_VSOURCES+=$(BASEJUMP_STL_DIR)/bsg_misc/bsg_adder_cin.v
+ARCH_VSOURCES+=$(BASEJUMP_STL_DIR)/bsg_misc/bsg_cycle_counter.v
```
tommydcjung: bsg_cycle_counter should be in the testbench list

machines/arch_filelist.mk
```diff
@@ -0,0 +1,118 @@
+# This file contains a list of the *core* manycore architecture files that are
+# used in simulation. This has NOT been used in tapeout or any tapeout related
+# activities beyond simulation
```
tommydcjung: we shouldn't change file names for no good reason

drichmond: Maybe there is a good reason. I believe Bandhav is trying to solve a legitimate problem here, so maybe you should ask him why, if you don't understand

machines/arch_filelist.mk
tommydcjung: bsg_defines.v has to be at the top of the list of files to be compiled.

machines/arch_filelist.mk
tommydcjung: @vb000 I don't understand why these filenames need to change?

machines/Makefile
```diff
@@ -1,50 +1,44 @@
...
 # Sets "VCS" variable
 include $(BSG_MANYCORE_DIR)/../bsg_cadenv/cadenv.mk
 
-MACHINES = 9x15 4x4
-
-SIMV = $(addsuffix /simv, $(MACHINES))
-SIMV_DEBUG = $(addsuffix /simv-debug, $(MACHINES))
-SIMV_DAIDIR = $(addsuffix .daidir, $(SIMV) $(SIMV_DEBUG))
+# Include source lists
+include Makefile.include
```
tommydcjung: why do we need to create another file? can we just have those lines include in this Makefile?

software/mk/Makefile.verilog
```diff
@@ -57,6 +57,10 @@ $(info Machine executable: $(BSG_SIM_EXE))
 
 %.run: NBF_FILE=$*.nbf
 %.run: %.nbf
+# Rebuid the machine if machines dir is internal
+ifeq ($(abspath $(BSG_MACHINE_PATH)/..),$(abspath $(BSG_MANYCORE_DIR)/machines))
```
tommydcjung: $(abspath $(BSG_MACHINE_PATH)/..), isn't this referring to BSG_MANYCORE_DIR?

machines/Makefile
vb000: Shows the toplevel dir name of the repo this Makefile is running in... so still would be correct!

machines/Makefile
vb000: We have two types of `.v` files in basejump_stl. `bsg_defines.v` is one kind with defines and `bsg_dff_reset.v` is the other kind with actual rtl. Hence, I used `VDEFINES` for `.v` with defines and `MACRO_DEFS` for cmdline macro defs. 

I suggest you include this distinction between sources as well in F1 because, this way we can group `VDEFINES` and make sure to compile them first.

machines/arch_filelist.mk
vb000: > We have two types of `.v` files in basejump_stl. `bsg_defines.v` is one kind with defines and `bsg_dff_reset.v` is the other kind with actual rtl. Hence, I used `VDEFINES` for `.v` with defines and `MACRO_DEFS` for cmdline macro defs.
@drichmond 

@tommydcjung! This is one of the reasons I used make variables instead `-f` option of vcs. We can group `bsg_defines.v` & co in `ARCH_VDEFINES` make variable. Files in `ARCH_VDEFINES` are compiled first:
https://github.com/bespoke-silicon-group/bsg_manycore/blob/1f653f7532fbdd19e054c0fc96c942dc97f2d7ac/machines/Makefile#L32

tommydcjung:
probably good for the newcomers to include a README file in machines/ to explain what are the available commands.

machines/Makefile.include
vb000: Oh, didn't realize we can have dirs in prerequisites! Then yes, it make sense to only have `VINCLUDES` listing incidrs.

machines/Makefile
vb000: @tommydcjung I had used `VCS_DEFINES` for listing `bsg_defines.v` kind of files. So, had to change this to macro_defs.

tommydcjung: in that case, would it be more appropriate to call it VERILOG_DEFINES? instead of VCS_DEFINES? Because bsg_defines.v really has nothing to do with VCS.

machines/arch_filelist.mk
vb000: I worked on existing filelist. So, cannot pay attention to this detail:
https://github.com/bespoke-silicon-group/bsg_manycore/blob/b4e4cc7625e8318b20ae66006414a8b07703153e/machines/core.include#L41

I'll move it in the next update.

machines/arch_filelist.mk
vb000: Previously file lists are just text files and now, they are make fragments. So, updated filenames so that people's text editors can detect make syntax.

machines/Makefile
vb000: It's a personal choice. I wanted to separate any "listings" we do in Makefile.include and primarily have rules in Makefile.

software/mk/Makefile.verilog
vb000: https://github.com/bespoke-silicon-group/bsg_manycore/blob/b4e4cc7625e8318b20ae66006414a8b07703153e/software/mk/Makefile.paths#L6

machines/Makefile
vb000: https://github.com/bespoke-silicon-group/bsg_manycore/blob/1f653f7532fbdd19e054c0fc96c942dc97f2d7ac/machines/Makefile.include#L10

Please refer to Dustin's comments about standardizing Make variable names.

tommydcjung: okay, is this standard documented somewhere? or does this rule change whenever someone makes a change?

machines/Makefile
tommydcjung: okay, I think if we are not including Makefile.include in some other places, there is no reason to have this in a separate file. It creates more work for the reader, that now the reader needs to open this file to understand what's going on. It's kind of the same reason why you are making this refactor. People don't like additional steps.

vb000: Yes, include file exclusively is for file listings and can be easily included in other places when the need arises.

machines/Makefile
vb000: @tommydcjung the idea is to make this easy to port, when adding support for other tools. This is a step towards that needing little effort.

vb000:
> probably good for the newcomers to include a README file in machines/ to explain what are the available commands.

just `make` and `clean`.

https://github.com/bespoke-silicon-group/bsg_manycore/blob/master/README.md
See initial steps for running programs section.

vb000:
Addressed all issues... merging this branch.

tommydcjung:
@vb000 hey, you are supposed to squash the commits

tommydcjung:
I don't think anyone approved the changes...

drichmond:
Yeah, I think we were still discussing changes

machines/Makefile
tommydcjung: Sure, but let's not create this file until that need actually arises.

machines/arch_filelist.mk
tommydcjung: that depends on what text editor people use, doesn't it?

drichmond:
Reopened branch

vb000:
> @vb000 hey, you are supposed to squash the commits

@drichmond Is that the usual practice? Will do from next time if so.

drichmond:
I think it depends a lot on your commit style. But in general one commit = 1 feature (once the PR is complete). Either a PR is small (like this), and you can squash it, or it's large (multiple features) and each feature is a commit. 

If you're not sure, you should ask.

vb000:
> I don't think anyone approved the changes...

I thought Dustin did after I made essential changes. And, I voluntarily implemented additional suggested changes.

drichmond:
I approved the initial PR. I thought I retracted it once we started discussing changes.

drichmond:
No worries. I like this PR, but it's worth discussion because I like the changes here enough that I'll consider this *the BSG Way of Doing Makefiles* and propagate it back to F1.

vb000:
I understand, if the discussion is essential for the pr. And we did have relevant discussion in the beginning and I made changes to address it.

But on general note, I don't think we should stall a pr with nitpicks blocking the author from moving on to doing other things. Unfortunately, I based my other branch on this. Otherwise I wouldn't have bothered spending my time arguing _personal preferences_.

drichmond:
On that note, and for discussion (this should go in a google doc), based on @vb000's changes here and my own experience the last few days

<purpose>.mk for included makefiles, where the purpose is obvious, like core_arch.

Makefile for non-included makefiles

`CINCLUDES/CXXINCLUDES` for C/C++ include **directories**
`CSOURCES/CXXSOURCES`,`CFLAGS/CXXFLAGS`, `CDEFINES/CXXDEFINES` etc
`CHEADERS/CXXHEADERS` - A list of header files. Not certain this is necessary, but it is useful sometimes.


`VDEFINES/VHDEFINES` (I know we don't use VHDL in BSG, but we use some VHDL in BSG F1 from Xilinx).
`VINCLUDES/VHINCLUDES` (Directories)
`VHEADERS/VHHEADERS`, I hate that there's a repeated H but we'll never use it so 🤷‍♂ 
`VFLAGS/VHFLAGS` This is probably the most unnecessary one, since there are very rarely, if never, flags that are consistent from toolchain to toolchain.

Since flags and defines usually have hyper tool-specific syntax, let's say that things like `<tool_name>_DEFINES` should be defined as, for example: 

`VCS_DEFINES +: $(foreach def,$(VDEFINES),+define+"$(def)")`

With additional flags (e.g. VCS_FLAGS) defined on subsequent lin

Ditto for FLAGS, and INCLUDES. This will allow us to support Vivado, and Verilator more easily. 

(@mrutt92 @dpetrisko @stdavids for more input)

tommydcjung:
@drichmond  If the file is intended to be included in another file (.mk), would it make sense to not have any targets in the include file? the include file should only contain variables that are being reused. only the main Makefile should have invokable targets.

drichmond:
urgh. Yes, but I've violated this rule in BSG F1 for various reasons.

drichmond:
The tricky part (and where I violated this) is when you have to generate a source file. If we do things this way, I may not have to generate it anymore.

drichmond:
Hmm. I did this a lot in BSG-F1. For example - when we have common rules between cosimulation subdirectories.

drichmond:
@tommydcjung What if we say that makefiles with rules must have a specific suffix? (not .mk)

And how would you feel if we defined a makefile functions that raise a terminal error when an expected variable is undefined? That way we know that we're using includes correctly.

dpetrisko:
@drichmond

(I’m not a make guru, this is just how I do things). 

For BlackParrot, we kind of invert the dependencies. There’s a Makefile which has variables for setup, like what testbench to run, which system configuration to run. Then it includes a Makefile.<tool> which contains targets. That way there’s always a tree of dependencies (no cycles or other problems with hierarchical make). 

For sources / incdirs / defines, we use flists, since the format is the same for DC, VCS and Verilator. This has the advantage of taking things out of makefiles (which is always a plus for me), so you can grep/cat/bash/etc. 

I’m not really familiar with the manycore make flow, so I’ll have to look to catch up and see. If we do have a bsg best practices come out of this, we should write it down somewhere

PR #140, opened 2019-08-29
adds C++ builddefs and an example spmd program

mrutt92:


vb000:
Isn't `c++` too generic for testcase name :p

tommydcjung:
@mrutt92 hey, you are supposed to delete the branch after they are merged or closed.

mrutt92:
Seems like a good thing to do here.

PR #141, opened 2019-08-30
Parameterized HW tracing of Branch & JALR target-PCs

vb000:
- Mapped EPA=0xEEE0 to stderr to have separate streams for program output and trace. Also, Newlib std streams could be separated.
- New bsg_manycore_lib macro bsg_putchar_err to write char to stderr stream.

### DIFF
--- a/machines/4x4/Makefile.machine.include
+++ b/machines/4x4/Makefile.machine.include
@@ -8,3 +8,4 @@ BSG_MACHINE_DRAM_BANK_SIZE_WORDS      = 134217728
 BSG_MACHINE_DRAM_INCLUDED             = 1
 BSG_MACHINE_MAX_EPA_WIDTH             = 28
 BSG_MACHINE_MEM_CFG                   = e_mem_cfg_default
+BSG_MACHINE_BRANCH_TRACE_EN           = 0
--- a/testbenches/common/v/bsg_nonsynth_manycore_monitor.v
+++ b/testbenches/common/v/bsg_nonsynth_manycore_monitor.v
@@ -6,6 +6,10 @@
 `include "bsg_manycore_packet.vh"
 
 module bsg_nonsynth_manycore_monitor
+ 
+  // Import address parameters
+  import bsg_manycore_addr_pkg::*;
+
   #(parameter x_cord_width_p="inv"
     , parameter y_cord_width_p="inv"
     , parameter addr_width_p="inv"
@@ -153,22 +157,22 @@ module bsg_nonsynth_manycore_monitor
     if (~reset_i) begin
       if (v_i & we_i) begin
         if (~addr_i[addr_width_p-1]) begin
-          if (epa_addr == 16'hEAD0) begin
+          if (epa_addr == bsg_finish_epa_gp) begin
             $display("[INFO][MONITOR] RECEIVED BSG_FINISH PACKET from tile y,x=%2d,%2d, data=%x, time=%0t",
               src_y_cord_i, src_x_cord_i, data_i, $time);
             $finish;
           end
-          else if (epa_addr == 16'hEAD4) begin
+          else if (epa_addr == bsg_time_epa_gp) begin
             $display("[INFO][MONITOR] RECEIVED TIME BSG_PACKET from tile y,x=%2d,%2d, data=%x, time=%0t",
               src_y_cord_i, src_x_cord_i, data_i, $time);
           end
-          else if (epa_addr == 16'hEAD8) begin
+          else if (epa_addr == bsg_fail_epa_gp) begin
             $display("[INFO][MONITOR] RECEIVED BSG_FAIL PACKET from tile y,x=%2d,%2d, data=%x, time=%0t",
               src_y_cord_i, src_x_cord_i, data_i, $time);
             $finish;
           end
-          else if (epa_addr == 16'hEADC || epa_addr == 16'hEEE0) begin
-            out_fd = (epa_addr == 16'hEADC) 
+          else if (epa_addr == bsg_stdout_epa_gp || epa_addr == bsg_stderr_epa_gp) begin
+            out_fd = (epa_addr == bsg_stdout_epa_gp) 
                        ? 32'h8000_0001  // 0xEADC => stdout
                        : 32'h8000_0002; // 0xEEE0 => stderr
 
@@ -178,13 +182,13 @@ module bsg_nonsynth_manycore_monitor
               end
             end
           end
-          else if (epa_addr == 16'hEEE4) begin
+          else if (epa_addr == bsg_branch_trace_epa_gp) begin
             // Branch and JALR trace
             $fwrite('h8000_0002, "BSG_BRANCH_TRACE t=%0t x=%0d y=%0d target=%x\n"
                       , $time, src_x_cord_i, src_y_cord_i, data_i
                    );
           end
-          else if (epa_addr == 16'h0D0C) begin
+          else if (epa_addr == bsg_print_stat_epa_gp) begin
             $display("[INFO][MONITOR] RECEIVED PRINT_STAT PACKET from tile y,x=%2d,%2d, data=%x, time=%0t",
               src_y_cord_i, src_x_cord_i, data_i, $time);      
           end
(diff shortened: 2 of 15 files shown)


testbenches/common/v/bsg_nonsynth_manycore_monitor.v
```diff
@@ -166,10 +167,14 @@ module bsg_nonsynth_manycore_monitor
...
-          else if (epa_addr == 16'hEADC) begin
+          else if (epa_addr == 16'hEADC || epa_addr == 16'hEEE0) begin
+            out_fd = (epa_addr == 16'hEADC) 
+                       ? 32'h8000_0001  // 0xEADC => stdout
+                       : 32'h8000_0002; // 0xEEE0 => stderr
+
             for (integer i = 0; i < 4; i++) begin
               if (mask_i[i]) begin
-                $write("%c", data_i[i*8+:8]);
+                $fwrite(out_fd, "%c", data_i[i*8+:8]);
```
tommydcjung: is it possible to color the stderr to some other color like red?

tommydcjung:
looks okay to me

testbenches/common/v/bsg_nonsynth_manycore_monitor.v
drichmond: Yeah... I think we did this somewhere in BSG F1 in Verilog code. There are certainly examples of coloring statements in Makefiles and in C/C++

tommydcjung:
@vb000 Make sure to update the relevant section in the hammerblade doc (e.g. EPA table). if you have done it already, thanks!

vb000:
> @vb000 Make sure to update the relevant section in the hammerblade doc (e.g. EPA table). if you have done it already, thanks!

I couldn't find such table... is there such a table already somewhere? I can add it in HB TRM of not.

testbenches/common/v/bsg_nonsynth_manycore_monitor.v
vb000: Not sure if it's a good idea to color an entire stream. Many times stderr is just used as a way to redirect output to a different fd, and not exclusively for error messages.

drichmond: You could have a BSG_ERR macro to selectively color

vb000: Oh okay! This EPA is for standard error, what do you think of separate EPA for BSG_ERROR?

drichmond: I think the discussion is somewhat orthogonal. Coloring standard error is probably unadvisable, but it may be a good idea to support coloring of arbitrary text from the manycore?

vb000:
@tommydcjung Added a table called "Host Manycore Memory Mapped IO" [here](https://docs.google.com/document/d/1b2g2nnMYidMkcn6iHJ9NGjpQYfZeWEmMdLeO_3nLtgo/edit#heading=h.wf7rr7gvocmv).

taylor-bsg:
Coloring should not be done from Verilog. Instead you should use one of our unique error strings as prefix to use in post-simulation grep —color.

v/bsg_manycore.v
```diff
@@ -90,6 +90,12 @@ module bsg_manycore
   //The IO router row index
   , parameter IO_row_idx_p = 0
 
+  // EPA parameter
+  , parameter bj_trace_epa_p = "inv" // EPA for branch/jalr trace on stderr
```
tommydcjung: Isn't this EPA a fixed value? I don't think we need this parameter.

machines/4x4/Makefile.machine.include
```diff
@@ -8,3 +8,4 @@ BSG_MACHINE_DRAM_BANK_SIZE_WORDS      = 134217728
 BSG_MACHINE_DRAM_INCLUDED             = 1
 BSG_MACHINE_MAX_EPA_WIDTH             = 28
 BSG_MACHINE_MEM_CFG                   = e_mem_cfg_default
+BSG_MACHINE_BJ_TRACE_EN               = 0
```
tommydcjung: can we call this something other than BJ...? I think Branch Trace is just fine.

v/vanilla_bean/vanilla_core.v
```diff
@@ -396,6 +402,21 @@ module vanilla_core
...
+  always_comb
+  begin
+    if (exe_r.decode.is_branch_op) begin
+      exe_pc_target = branch_under_predict
+        ? exe_r.pred_or_jump_addr[2+:pc_width_lp]
+        : exe_r.pc_plus4[2+:pc_width_lp];
+    end else if (exe_r.decode.is_jalr_op)
+      exe_pc_target = alu_jalr_addr;
+    else
+      exe_pc_target = {pc_width_lp{1'b1}}; // invalid pc target
```
tommydcjung: I don't think we need this else statement, and replace the else-if above with else. It's either branch or jalr mispredict.

vb000: That value should only be used when it's a branch/jalr op. Deliberately assigned invalid PC there, so that we'd see an error when some logic uses `exe_pc_target` to update PC incorrectly.

v/bsg_manycore.v
vb000: Firstly, it's more readable if it's a parameter instead of a hex number. Moreover, EPA mapping can change.

v/vanilla_bean/vanilla_core.v
tommydcjung: yes, there is no such condition, where {pc_width_lp{1'b1}} will be actually written. It's just creating extra hardware for no reason.

v/bsg_manycore.v
tommydcjung: make it a macro then? there is no need to propagate this parameter all the way from top module to the tile.

machines/4x4/Makefile.machine.include
tommydcjung: The reason is that you could be offending some people.. gotta be careful when picking acronyms..

vb000: Oh okay! I'll change it to branch tarce.

v/bsg_manycore.v
vb000: I think it should be parameter with the current design. Because that'd create the same variable in different places in the design. It's difficult for someone else to know there is a macro here to be changed when the epa value changes.

v/vanilla_bean/vanilla_core.v
vb000: Okay. But the constant would be greatly optimized by synth tool and extra mux logic wouldn't be significant.

tommydcjung: Here is a better idea: just plug in pc_n into lsu. no changes needed.

vb000: Removed the redundant condition.

v/bsg_manycore.v
tommydcjung: you can define a macro in parameters.vh, since this is an architectural parameter. then it's just one place to modify.

vb000: This is not an architectural parameter. It's a system parameter. It's the EPA manycore monitor, f1 host code looks for. If someone changes what this EPA is in f1 host code, they shouldn't be required to modify parameters.vh.

tommydcjung: It's actually the other way around. EPA is defined, and if it even changes, f1 needs to update. Even though, I don't see a case where this will remap. So I think it's better to define as a macro.

vb000: This implementation is flexible in the sense that VCS env, F1 env or some other env can choose different EPA mappings.

I understand the concern about proliferation of parameters. When that happens, we can package all these parameters in a SV package and import in modules. That way different envs can have different package files having the same flexibility. Right now, I don't think parameters in these modules don't seem too many.

vb000: Actually, I can do it right now. I'll package the EPA mappings in SV package and import it wherever required. It's more cleaner! Sounds good?

tommydcjung: package sounds better.

v/bsg_manycore_addr_pkg.v
```diff
@@ -0,0 +1,29 @@
...
+ *    Manycore address constants
+ */
+
+package bsg_manycore_addr_pkg;
+
+  // NPA prefixes
+  parameter bsg_dram_npa_prefix_gp = 32'h8000_0000;
+  parameter bsg_io_npa_prefix_gp   = 32'h4000_0000;
+
+  // IO EPA
```
tommydcjung: just add a comment whether this is meant to be byte address or word address

vb000: Added the comment. NPAs & EPAs are always word addresses though...

PR #144, opened 2019-09-10
Simple program to test fast heavy printing

vb000:
A simple program to push the IO interface to limits -- fills up fifos in IO interface and network routers and cause a stall in the core. This uncovered a minor issue in bsg_f1's fifo interface where credits are being returned even when the fifo is full.


drichmond:
@sampsyo Would this reproduce the bug you see?

PR #148, opened 2019-09-13
Changed bsg_newlib to dramfs

vb000:
Changed bsg_newlib name to dramfs. Reflecting that change here.

### DIFF
--- a/software/spmd/common/spike.ld
+++ b/software/spmd/common/spike.ld
@@ -33,7 +33,7 @@ SECTIONS
 
   .text : 
   {
-     *crt.o(.text)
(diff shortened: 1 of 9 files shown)


software/spmd/common/spike.ld
```diff
@@ -33,7 +33,7 @@ SECTIONS
 
   .text : 
   {
-     *crt.o(.text)
+     *(.crtbegin)
```
tommydcjung: maybe add a comment why this change was necessary?

vb000: done!

PR #153, opened 2019-09-15
Enable mstatus.FS

vb000:


software/spmd/common/crt.S
```diff
@@ -35,6 +35,16 @@ _start:
   li  x29,0
   li  x30,0
   li  x31,0
+
+  # Enable FPU and clear fcsr.
+  #
+  # These are ignored by manycore hardware but might be
+  # required by execution environments supporting 
+  # exceptions (such as Spike).
+  li t0, 0x00003000 # mstatus.FS
```
tommydcjung: is it possible to surround it with #ifdef and enable only for  spike? 
If all instructions need to match PC by PC, then add bubbles in #else?
clearly `li t0, 0x00003000` is not ignored by manycore hardware.

vb000: `t0` a temporary register. RISC-V calling convention provides no guarantees on values stored in temporary registers. So no program would be expecting any specific value in `t0`.

vb000: I think ifdef is not necessary here and it's better minimize their use. I can add an instruction to initialize t0 back to 0, if that helps?

tommydcjung: if the calling convention ignores it, then it should be fine.
I don't think any instruction that is implemented uses SYSTEM opcode, so I think it's safe.

PR #154, opened 2019-09-17
Corresponding RISCV codes for cuda lite runtime floating point regression tests

bornaehsani:
Adds the following floating point cuda lite runtime regression tests in the bsg_manycore_directory:

- float_vec_add() --> A[i] + B[i] --> C[i]
- float_vec_mul() --> A[i] * B[i] --> C[i]
- float_vec_div() --> A[i] / B[i] --> C[i]
- float_vec_exp() --> exp(A[i]) --> C[i]
- float_matrix_mul
- float_matrix_mul_shared_mem
- float_all_ops() --> Includes all combinations of corner cases (inf, -inf, nan, -nan, 1/0, 0+inf, inf * 0, etc.) and performs all operations (add, sub, mul, div, compare, float to in convert) on a vector.

Related to PR # 410 in bsg_f1. 
@vb000 in the float_vec_exp test, you will notice the added (-filter_out -nostdlib) option from the RISCV linker, which was necessary to be able to include the math.h and get expf() function.


tommydcjung:
just making some comments.. might be interesting to add square-root and logarithmic too, if possible.

bornaehsani:
> 
> 
> just making some comments.. might be interesting to add square-root and logarithmic too, if possible.

that's actually a good idea. I will add these tests, thanks!

bornaehsani:
> 
> 
> just making some comments.. might be interesting to add square-root and logarithmic too, if possible.

Added sqrt and log regression tests. both pass. thanks!

software/spmd/bsg_cuda_lite_runtime/float_vec_sqrt/a.dis
```diff
@@ -0,0 +1,13607 @@
+Toplevel Directory: /home/centos/bsg_bladerunner//bsg_manycore
```
tommydcjung: do we have to add this file??

bornaehsani: No my bad, it was left over. Thanks for pointing it out, good catch!

PR #155, opened 2019-09-17
Specint

vb000:
- Complete specint flow
- Build spike with commit logs enabled
- Directories support for newlib dramfs
- Flow to run specint on host (x86)


software/spmd/common/link_dram.ld
```diff
@@ -196,6 +196,8 @@ SECTIONS
 
     _dram_misc = .;
     *(.*)
+
+    . = ALIGN(4);
```
mrutt92: why is this align necessary? (comments are a good thing)

software/py/func_call_log.py
```diff
@@ -0,0 +1,64 @@
...
+
+if len(sys.argv) != 3:
+  print("Usage: func_call_log.py <pc_trace_file> <elf_file>")
+  sys.exit()
+
+
+###########################################
+# Parse ELF to extract function list
+
+# This works on any ELF
```
mrutt92: Yes, but what does 'work' mean? You should comment what the sed command is doing (what are you searching and replacing).

vb000: Isn't the comment "Parse ELF to extract function list" clear about what it's doing? I can elaborate and yes, that sed needs comments.

software/spmd/specint2000/README.md
```diff
@@ -9,3 +9,11 @@ SPECInt2000 Benchamarks on Manycore
   contains manycore versions of required raw library functions.
 - `Makefile.BenChMaRK`: Makefile to run `BenChMaRK`
 - `Makefile`: Makefile to run all specint benchmarks
+
+Status:
```
tommydcjung: if not all the benchmark compiles or run correctly yet, what's the hurry to get this into the master?
Are you simply looking for feedback on the makefile infrastructure you are building?

software/spmd/common/link_dram.ld
vb000: Oh this, I don't think this line is required. But, without this line loading binaries linked with this script to cosim/f1 was not possible. Thought of discussing with you guys but forgot to, thanks for pointing it out!

software/spmd/specint2000/README.md
vb000: No, I want this on master. It's getting hard to work on branches long term. Especially, when having to run these programs on bsg_bladerunner. 

True, some benchmarks are not compiling because of missing sources and some are failing. But from this repo's point of view, the flow is complete.

software/py/func_call_log.py
mrutt92: That comment is good, but parsing a regex command in a human brain is always confusing. 
I guess it's easy to infer from the `['00000000 _start']` on the next line what the goal is. 
Here's a suggestion: instead of using sed for extracting two fields from stream, you should use awk. That would replace the sed command with `awk {print $2 " " $8}` which in awk speak means "print out the 2nd and 8th space separated word of each line." A little easier to understand than a regular expression.

vb000:
@mrutt92 Added suggested comments.

software/py/func_call_log.py
```diff
@@ -0,0 +1,65 @@
...
+  print("Usage: func_call_log.py <pc_trace_file> <elf_file>")
+  sys.exit()
+
+
+###########################################
+# Parse ELF to extract function list
+
+# Command to parse input ELF to 
+# generate "<start_address> <function_name>" pairs.
+elf_parse_cmd = "readelf -s " \
```
mrutt92: nice.

PR #156, opened 2019-09-18
Minimal test to reproduce expf issue

vb000:
Test to compare manycore's expf result with Spike's result and they are mismatching! Seems to reproduce the FP issue concisely.

@tommydcjung You could use this to debug FP issue:

`make`: run this test on manycore
`make SPIKE=1`: run this Spike.

Fun fact: Spike result matches x86 result (bitwise)! x86 might be implementing ieee754 after all...

### DIFF
--- a/software/spmd/expf/main.c
+++ b/software/spmd/expf/main.c
@@ -1,0 +1,17 @@
+#include <math.h>
+#include <stdio.h>
+
+float data = 3.4;
+
+int main() {
+  union {
+    int i;
+    float f;
+  } res;
+
+  res.f = expf(data);
+
+  printf("%x\n", res.i);
+
+  return 0;
+}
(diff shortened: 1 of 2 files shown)


software/spmd/expf/main.c
```diff
@@ -0,0 +1,17 @@
...
+
+int main() {
+  union {
+    int i;
+    float f;
+  } res;
+
+  res.f = expf(data);
+
+  printf("%x\n", res.i);
```
tommydcjung: the test should be self-checking. maybe add check the value and send the fail packet if it does not match the value expected by x86?

tommydcjung:
I'm not sure what the allegation is. does expf use the FPU hardware or is it software-emulated?

tommydcjung:
also, make sure that the compiler does not generate any fma instructions.

PR #161, opened 2019-09-20
Toolchain newlib patch

vb000:
- Patch riscv-gnu-toolchain to fix a installation issue.
- Added target cflags -- this fixes expf issue!!
https://github.com/bespoke-silicon-group/bsg_manycore/blob/8e93b73a7a87ab11f348b3ec629dd79df9288d6b/software/riscv-tools/Makefile#L32
BSG Manycore's FPU needs a set of gcc flags, because it don't implement fused mul-add, fexp etc. We have those options in our C compilation flow. The issue was, they were not added to the library compilation flow, which caused the math library to be compiled assuming fma (fused mul-add) instructions are legal. Since we don't support fma instructions, expf which had fma instructions in it's compiled binary produces wrong results on our hardware. This addition tells the toolchain build flow to compile library code with provided cflags.

- Added back `--remote` flag to Newlib's `git submodule update` command. Git 1.8 and later versions have this feature of a submodule tracking a branch. This is required because bsg_manycore tracks our custom newlib port.
https://github.com/bespoke-silicon-group/bsg_manycore/blob/8e93b73a7a87ab11f348b3ec629dd79df9288d6b/software/riscv-tools/Makefile#L91

### DIFF
--- a/software/riscv-tools/Makefile
+++ b/software/riscv-tools/Makefile
@@ -81,7 +81,7 @@ replace-newlib:
 	cd $(TOOLCHAIN_REPO) && git config --file=.gitmodules submodule.riscv-newlib.branch $(BSG_NEWLIB_BRANCH)
 	cd $(TOOLCHAIN_REPO) && git apply ../$(BSG_NEWLIB_PATCH)
 	cd $(TOOLCHAIN_REPO) && git submodule sync
-	cd $(TOOLCHAIN_REPO) && git submodule update --init --recursive riscv-newlib
+	cd $(TOOLCHAIN_REPO) && git submodule update --remote riscv-newlib
 
 checkout-spike:
 	@echo "====================================="
--- a/software/riscv-tools/Makefile
+++ b/software/riscv-tools/Makefile
@@ -29,10 +29,12 @@
 
 TOOLCHAIN_REPO=riscv-gnu-toolchain
 TOOLCHAIN_URL=https://github.com/riscv/$(TOOLCHAIN_REPO)
+TOOLCHAIN_VERSION=a03290eab661e2aa58288ad164f908bbbcc2169c
 DEPENDS_DIR=$(CURDIR)/depends
 
 BSG_NEWLIB_URL=https://github.com/bespoke-silicon-group/bsg_newlib_dramfs.git
 BSG_NEWLIB_BRANCH=dramfs
+BSG_NEWLIB_PATCH=newlib.patch
 
 SPIKE_REPO=riscv-isa-sim
 SPIKE_URL=https://github.com/riscv/riscv-isa-sim.git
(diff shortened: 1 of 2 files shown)


drichmond:
This fixes the fma issue? What flag fixes it?

vb000:
@drichmond 
https://github.com/bespoke-silicon-group/bsg_manycore/blob/8e93b73a7a87ab11f348b3ec629dd79df9288d6b/software/riscv-tools/Makefile#L32

drichmond:
Sorry - I should have been more clear. How do those flags fix it?

Channeling max - a comment would be nice

(I saw the flags, just mis-worded my question)

tommydcjung:
I'm going to test this out of box

vb000:
@drichmond bsg_manycore's FPU needs a set of gcc flags, because we don't support fused mul-add etc. We have those options in our C compilation flow. The issue was, they were not added to the library compilation flow, which caused the math library to be compiled with fma (fused mul-add) instructions. Since we don't support fma instructions, expf which had fma instructions in it's compiled binary produces wrong results on our hardware.

software/riscv-tools/Makefile
```diff
@@ -27,16 +27,25 @@
 #
 .DEFAULT_GOAL := help
 
-TOOLCHAIN_REPO=riscv-gnu-toolchain
-TOOLCHAIN_URL=https://github.com/riscv/$(TOOLCHAIN_REPO)
-DEPENDS_DIR=$(CURDIR)/depends
+TARGET_ARCH       := rv32imaf
+TARGET_ABI        := ilp32 # TODO: Test ilp32f for potential performance improvement
```
tommydcjung: create an issue instead.

vb000: Good idea!

drichmond:
> @drichmond bsg_manycore's FPU needs a set of gcc flags, because we don't support fused mul-add etc. We have those options in our C compilation flow. The issue was, they were not added to the library compilation flow, which caused the math library to be compiled assuming fma (fused mul-add) instructions are legal. Since we don't support fma instructions, expf which had fma instructions in it's compiled binary produces wrong results on our hardware.

This makes sense! Two questions - do we have a single source of truth for the flags we pass to both builds? And can you add this comment (or similar) describing the flags?

vb000:
I'll add the comment, thanks! Some flags could be different -- it'll not always be the case that program flags and toolchain flags are the same. It's better to not add additional flags to toolchain flow unless absolutely required in cases like this.

drichmond:
@bornaehsani is going to test this on F1 just to make sure

drichmond:
Unless you already have

bornaehsani:
@vb000 @drichmond Should there be any changes done to the Makefile of the manycore riscv binary as well? Because I rebased the cuda_fp_tests branch on top of the master that includes this fix, and re-ran the float_vec_exp test. However, I'm seeing the same results, and the binary still contains f.madd instructions.

drichmond:
Did you rebuild the tools in bsg_manycore?

bornaehsani:
Just occurred to me that I haven't :stupid. I will try that now.

bornaehsani:
@drichmond @vb000 I rebuilt and re-ran. All floating point tests passed including expf, and the biggest relative error in expf was 1e-07. Thanks Bandhav!

PR #163, opened 2019-09-20
Blood graph2

tommydcjung:


tommydcjung:
blood

tommydcjung:
fft

blood

mrutt92:
wow.

bornaehsani:
wow indeed, this is great work Tommy, thanks to both you and Bandhav! I will start trying it out soon!

drichmond:
Fascinating

tommydcjung:
also, refactored post_process_trace.py using the reusuable parsers.

taylor-bsg:
@tommydcjung Now that you know what kind of stalls there are a lot of of,
it makes sense for you to choose a color scheme that will better illuminate
what is going on. As you can tell, most of the common stalls currently have
a reddish, pink, or purple color. I would go select colors that are more
easy to differentiate. And also, I would make it so that programmer
correctable stalls stand out more than non-correctable ones.

It's also worth looking at the blood graph and making sure the processor is
doing what you expect it to in each of the phases of program execution. =)

PR #165, opened 2019-09-21
add enable signal for vcache tracer

tommydcjung:


drichmond:
Why is it no longer needed? (otherwise, looks good to me)

tommydcjung:
we used to have this to get the list of all the accessed address, so we can minimize the nbf file by reducing the tag initialization step. But since we moved on to DRAM mode, this is really not needed. maybe I should just give an option to disable this, since this might come useful for debugging later on.

tommydcjung:
But we want to be able to disable the tracing to speed up the simulation.

PR #167, opened 2019-09-30
Modified blood_graph to include lr_aq_stall and also fp_executed

bornaehsani:
- There are cases in the vanilla.log where both int_pc and fp_pc show BUBBLE, however a stall reason exists, such as LR_AQ and LOADWB stall. 
- Modified the code to first search for a stall reason, and if that doesn't exist it can search for int_pc and fp_pc and if none exist it can then depict BUBBLE.
- fp_executed wasn't included so I added that. 
- Made the LR_AQ stall color a little brighter (it was too heavy on the eyes)

### DIFF
--- a/software/py/blood_graph.py
+++ b/software/py/blood_graph.py
@@ -18,16 +18,17 @@
 #   {timestep}    time step in picosecond
 # 
 #   Color Code:  
-#   executed  = ffffff (white)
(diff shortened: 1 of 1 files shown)


bornaehsani:
Went Stale. New PR will come soon.

PR #168, opened 2019-09-30
Added parser to read through vanilla_stats 

bornaehsani:
And generate a execution_stats.log file that prints out total execution time, number and fraction of each instruction executed, number and fraction of stalls and different types of misses.

### DIFF
--- a/software/py/generate_stats.py
+++ b/software/py/generate_stats.py
@@ -20,6 +20,29 @@ import re
 from enum import Enum
 
 
+instructions_list = ['instr','fadd','fsub','fmul','fsgnj','fsgnjn',\
+                     'fsgnjx','fmin','fmax','fcvt_s_w','fcvt_s_wu',\
+                     'fmv_w_x','feq','flt','fle','fcvt_w_s','fcvt_wu_s',\
+                     'fclass','fmv_x_w','local_ld','local_st',\
+                     'remote_ld','remote_st','local_flw','local_fsw',\
+                     'remote_flw','remote_fsw','lr','lr_aq',\
+                     'swap_aq','swap_rl','beq','bne','blt','bge','bltu',\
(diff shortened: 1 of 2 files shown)


software/py/generate_stats.py
```diff
@@ -0,0 +1,227 @@
...
+    self.print_stats_timing()
+    self.print_stats_instructions()
+    self.print_stats_stalls()
+    self.print_stats_miss()
+
+
+
+
+  # default stats generator
+  def generate_stats(self, input_file):
```
tommydcjung: I think there is a csv module for python.
https://docs.python.org/3.6/library/csv.html

software/py/generate_stats.py
```diff
@@ -0,0 +1,227 @@
...
+
+stalls_list = ['stall_fp_remote_load','stall_fp_local_load',\
+               'stall_depend','stall_depend_remote_load',\
+               'stall_depend_local_load','stall_force_wb',\
+               'stall_ifetch_wait','stall_icache_store',\
+               'stall_lr_aq','stall_md,stall_remote_req','stall_local_flw']
+
+
+
+class Stats:
```
tommydcjung: I would call it something more descriptive like VanillaStatsParser.
the filename should match the name of the python class in the file.

software/py/generate_stats.py
```diff
@@ -0,0 +1,227 @@
...
+#   ex) python3 generate_stats.py 4 4 vanilla_stats.log
+#
+#   {manycore_dim_y}  Mesh Y dimension of manycore
+#   {manycore_dim_x}  Mesh X dimension of manycore
+
+
+
+import sys
+import os
+import re
```
tommydcjung: 're' not used?

bornaehsani:
Stale. Replaced by PR #181 .

PR #172, opened 2019-09-30
Branch trace csr

vb000:
Implemented branch trace CSR hardware and related nbf changes.

### DIFF
--- a/software/mk/Makefile.builddefs
+++ b/software/mk/Makefile.builddefs
@@ -47,6 +47,9 @@ else
   endif
 endif
 
+# Enable branch trace during boot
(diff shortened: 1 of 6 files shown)


tommydcjung:
please make sure that all the regression (including beebs) passes before making any hardware changes.

tommydcjung:
I wouldn't merge this change yet, because we are still waiting to see how it affects the critical path and area  of the tile.

vb000:
Closing this as there's no near-term use for this extra hardware to support run-time toggling of branch trace. Besides, there's a compile time option to do the same.

PR #175, opened 2019-10-03
stall_trace.log; fixed bug with stall_depend counting

tommydcjung:


software/spmd/hello_float/Makefile
```diff
@@ -1,46 +1,14 @@
-#########################################################
-# Network Configutaion
-# If not configured, Will use default Values
-#	bsg_global_X ?= $(bsg_tiles_X)
```
drichmond: Just curious - why was this Makefile changed?

tommydcjung: I was using it to generate vanilla_stall_trace.log. I would have used coremark or fft, if they worked, but both seem to be broken now.

software/py/blood_graph.py
```diff
@@ -45,14 +33,21 @@ def __init__(self, start_time, end_time, timestep):
...
 
-    self.parser = VanillaTraceParser()
-
   
   # main public method
   def generate(self, input_file):
-    # parse vanilla.log
-    traces = self.parser.parse(input_file)
-
+    # parse vanilla_stall_trace.log
```
vb000: Think script help message should be updated...

PR #176, opened 2019-10-03
Fix LLVM installation

drichmond:
I've removed the dependence that LLVM_DIR be a globally accessible directory. Now, it is built with the RISC-V tools on a clean installation. 

This makes bsg_manycore immediately portable to kk9, and simplifies the aws build process (a bit)

This also updates to cmake3 (instead of cmake) which means we no longer need to download and build cmake.

### DIFF
--- a/software/mk/Makefile.llvminstall
+++ b/software/mk/Makefile.llvminstall
@@ -7,10 +7,6 @@ ifndef RISCV_INSTALL_DIR
 endif
 
 llvm-install:
-	# Install cmake
-	wget https://github.com/Kitware/CMake/releases/download/v3.14.0-rc4/cmake-3.14.0-rc4-Linux-x86_64.sh -O cmake-install.sh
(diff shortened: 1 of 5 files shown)


software/mk/Makefile.llvminstall
```diff
@@ -7,10 +7,6 @@ ifndef RISCV_INSTALL_DIR
 endif
 
 llvm-install:
-	# Install cmake
-	wget https://github.com/Kitware/CMake/releases/download/v3.14.0-rc4/cmake-3.14.0-rc4-Linux-x86_64.sh -O cmake-install.sh
-	chmod +x cmake-install.sh && sudo ./cmake-install.sh --skip-license --prefix=/usr/local
-	rm cmake-install.sh
 	mkdir -p $(LLVM_DIR)/llvm-build && mkdir -p $(LLVM_DIR)/llvm-install
```
drichmond: We no longer need to build cmake3. We have a sufficient version on kk9

drichmond:
This is relatively high priority, since it unblocks people working on KK9

PR #177, opened 2019-10-07
Equivalent spike bins

vb000:
Mods to compile very similar binaries for spike and manycore.


drichmond:
I don't fully understand the the problem that this PR is addressing. Can you explain more?

vb000:
Binaries compiled for Spike are different from those compiled for Manycore, because both run different pieces of code to print to console, for example. This difference makes it tough for us to compare traces produced by those binaries against each other. To solve this problem, this pr adds a mode to compile very similar binaries which produce very similar traces. This is achieved by having a global variable called `spike_run`, which is used to conditionally run different pieces of code at runtime -- as opposed to producing different binaries at compile time.

drichmond:
Great description! Makes sense! Can you add this in the commit description when you merge?

PR #180, opened 2019-10-21
Vanilla core profiler update and bloodgraph refactor

bornaehsani:
1. Modified vanilla_core_profiler.v to output the specific type of instruction being executed instead of printing (no_stall)
2. Modified bloodgraph to draw all stall types and all instructions (instructions are drawn the same color for now, with the exception of integer, fp, and unknown instructions having their own color.)
3. Bloodgraph now has two modes, abstract and detailed. In abstract, most insignificant stall types are colored the same and focus is on ALU/FPU utilization instead of stall types.
4. Now using argparse for input arguments 
5. Thanks to @mrutt92 who added a python script to depict a visual color key for bloodgraph


testbenches/common/v/vanilla_core_profiler.v
```diff
@@ -701,7 +701,178 @@ module vanilla_core_profiler
...
+            else if (fcvt_wu_s_inc)
+              $fwrite(fd2,"%0t,%0d,%0d,%s", $time, my_x_i, my_y_i, "fcvt_wu_s");
+            else if (fclass_inc)
+              $fwrite(fd2,"%0t,%0d,%0d,%s", $time, my_x_i, my_y_i, "fclass");
+            else if (fmv_x_w_inc)
+              $fwrite(fd2,"%0t,%0d,%0d,%s", $time, my_x_i, my_y_i, "fmv_x_w");
+
+
+             else if (instr_inc | fp_instr_inc)
+              $fwrite(fd2,"%0t,%0d,%0d,%s", $time, my_x_i, my_y_i, "unknown");
```
tommydcjung: does this ever happen?

testbenches/common/v/vanilla_core_profiler.v
```diff
@@ -701,7 +701,178 @@ module vanilla_core_profiler
           else if (stall_local_flw)
             $fwrite(fd2,"%0t,%0d,%0d,%s", $time, my_x_i, my_y_i, "stall_local_flw");
           else
-            $fwrite(fd2,"%0t,%0d,%0d,%s", $time, my_x_i, my_y_i, "no_stall");
+          begin
+
+            if (local_ld_inc)
+              $fwrite(fd2,"%0t,%0d,%0d,%s", $time, my_x_i, my_y_i, "local_ld");
```
tommydcjung: lots of repeating code here! let me suggest a way to make it little better.

make a task like this.
```
 task print_trace(integer fd, string op);
    $fwrite(fd, "%0t,%0d,%0d,%s\n", $time, my_x_i, my_y_i, op);
  endtask
```

which then can be called like
```
print_trace(fd2, "stall_md");
```

testbenches/common/v/vanilla_core_profiler.v
tommydcjung: might happen if floating-point csr instruction is used in one of fdiv or fsqrt function.

software/py/blood_graph_key.py
```diff
@@ -0,0 +1,48 @@
...
+HEIGHT = 256
+DEFAULT_MODE = "detailed"
+
+def parse_args():
+    parser = argparse.ArgumentParser(description="Depicts a visual key for bloodgraph")
+    parser.add_argument("--mode", default=DEFAULT_MODE, type=str,
+                        help="Type of bloodgraph - abstract / detailed")
+    args = parser.parse_args()
+
+    if (not (args.mode == "abstract" or args.mode == "detailed")):
```
drichmond: Fun fact, you can write custom parsers so that you don't need to do this. 

https://docs.python.org/3.6/library/argparse.html#action

software/py/blood_graph_key.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+if __name__ == "__main__":
+    args = parse_args()
+
+    bg = BloodGraph(0,1,1, args.mode)
+    img = Image.new("RGB", (WIDTH, HEIGHT), "black")
+    draw = ImageDraw.Draw(img)
+    font = ImageFont.load_default()
+    # get the size of the font
```
drichmond: This comment seems misplaced.

software/py/blood_graph_key.py
```diff
@@ -0,0 +1,48 @@
...
+
+    return args
+
+
+
+if __name__ == "__main__":
+    args = parse_args()
+
+    bg = BloodGraph(0,1,1, args.mode)
+    img = Image.new("RGB", (WIDTH, HEIGHT), "black")
```
drichmond: Why are WIDTH and HEIGHT constants?

software/py/blood_graph_key.py
```diff
@@ -0,0 +1,48 @@
...
+    draw = ImageDraw.Draw(img)
+    font = ImageFont.load_default()
+    # get the size of the font
+    yt = 0
```
drichmond: What does `yt` do? A more descriptive name would help

software/py/blood_graph_key.py
```diff
@@ -0,0 +1,48 @@
...
+    return args
+
+
+
+if __name__ == "__main__":
+    args = parse_args()
+
+    bg = BloodGraph(0,1,1, args.mode)
+    img = Image.new("RGB", (WIDTH, HEIGHT), "black")
+    draw = ImageDraw.Draw(img)
```
drichmond: does the draw object allow you to draw shapes in the image?

software/py/blood_graph_key.py
```diff
@@ -0,0 +1,48 @@
...
+        sys.exit()
+
+    return args
+
+
+
+if __name__ == "__main__":
+    args = parse_args()
+
+    bg = BloodGraph(0,1,1, args.mode)
```
drichmond: Does BloodGraph automatically generate an output image?

software/py/blood_graph_key.py
```diff
@@ -0,0 +1,48 @@
...
+    img = Image.new("RGB", (WIDTH, HEIGHT), "black")
+    draw = ImageDraw.Draw(img)
+    font = ImageFont.load_default()
+    # get the size of the font
+    yt = 0
+    # for each color in stalls...
+    for (key,color) in chain(bg.stall_bubble_color.iteritems(),
```
drichmond: Seems odd to name your iteration variable `key`, when you're drawing the key. Maybe cause would be better if you are iterating through the stall causes

software/py/blood_graph_key.py
```diff
@@ -0,0 +1,48 @@
...
+                             [("unified_instr"    ,bg.unified_instr_color),
+                              ("unified_fp_instr" ,bg.unified_fp_instr_color),
+                              ("unknown"          ,bg.unknown_color)]):
+        (font_height,font_width) = font.getsize(key)
+        yb = yt + font_width
+        #color = bg.stall_bubble_color[key]
+        draw.rectangle([0, yt, 64, yb], color)
```
drichmond: I don't understand what 0, 64, yt, and yb are for

software/py/blood_graph.py
```diff
@@ -3,48 +3,230 @@
...
-  def __init__(self, start_time, end_time, timestep):
+  def __init__(self, start_time, end_time, timestep, mode):
 
     self.start_time = start_time
     self.end_time = end_time
     self.timestep = timestep
+    self.mode = mode
+
+    # List of types of stalls incurred by the core 
+    self.stall_list    = ["stall_depend",
```
drichmond: I'd call this `stalls` and make it a set

software/py/blood_graph.py
```diff
@@ -3,48 +3,230 @@
...
+                          "stall_ifetch_wait",
+                          "stall_icache_store",
+                          "stall_lr_aq",
+                          "stall_md",
+                          "stall_remote_req",
+                          "stall_local_flw" ]
+
+
+    # List of types of integer instructions executed by the core 
+    self.instr_list    = ["local_ld",
```
drichmond: Same here. `instructions` and make it a set

software/py/blood_graph.py
```diff
@@ -101,50 +285,52 @@ def mark_trace(self, trace):
...
+  parser.add_argument("--start", default=DEFAULT_START_TIME, type=int,
+                      help="Starting cycle of bloodgraph")
+  parser.add_argument("--end", default=DEFAULT_END_TIME, type=int,
+                      help="Ending cycle of bloodgraph")
+  parser.add_argument("--timestamp", default=DEFAULT_TIMESTAMP, type=int,
+                      help="Distance between each trace in cycles")
+  args = parser.parse_args()
+
+  if (args.start > args.end):
+    parser.error("start cycle cannot be larger than end time.")
```
drichmond: https://docs.python.org/3.6/library/argparse.html#action

software/py/blood_graph.py
```diff
@@ -101,50 +285,52 @@ def mark_trace(self, trace):
...
+    sys.exit()
+
+  if (not (args.mode == "abstract" or args.mode == "detailed")):
```
drichmond: https://docs.python.org/3.6/library/argparse.html#action

drichmond:
In general, I think this is good. But sense a good portion of this is Python, I think it should be more intuitive and easier to comprehend. Most of my comments are to this effect.

I do have some questions about why the format and print methods are combined. In general, it's bad form to parse, format and print in the same method because it makes your code less reusable.

I don't understand why we need separate scripts for generating the key, and graph -- or why you need to create BloodGraph object during key generation. Do you need to access the class variables? Then those variables should be class variables, not instance variables.

Can discuss later today if need be.

software/py/blood_graph_key.py
mrutt92: Changing the font characteristics (i.e. the font height and width) looked like more trouble than it was worth. Given that, I picked a width and height that looked reasonable.

software/py/blood_graph_key.py
mrutt92: yes

software/py/blood_graph_key.py
mrutt92: No

mrutt92: You need to call generate() to do that.

software/py/blood_graph_key.py
drichmond: I'd argue that it should be called "drawer"

software/py/blood_graph_key.py
drichmond: Still want to know why creating a BloodGraph object is necessary

mrutt92: it was to harvest it for its internal members; moot now though as I've unified the scripts.

software/py/blood_graph_key.py
mrutt92: I like draw

testbenches/common/v/vanilla_core_profiler.v
bornaehsani: Yes I've seen it pop up in multiple regression test operation logs. I've named it unknown for now, until we can identify it later.

software/py/blood_graph_key.py
bornaehsani: Yeah I used argparse in the blood_graph.py, was too lazy to do the same for blood_graph_key.py, it's merged into blood_graph.py now (thanks to Max) so we're good.

testbenches/common/v/vanilla_core_profiler.v
bornaehsani: Great point! Done, thanks!

software/py/blood_graph_key.py
bornaehsani: changed it to "operation".

software/py/blood_graph.py
bornaehsani: Thanks for pointing it out! I added a TimeAction to take in time as start_time@end_time and check if start time is less than end time.

software/py/blood_graph.py
bornaehsani: Good point, modified the input argument to a --abstract argument which is false by default. Thanks!

bornaehsani:
Deprecated. Replaced by PR #198.

PR #183, opened 2019-10-21
Double div bug

tommydcjung:


drichmond:
I’m actually super curious about this. Can you add a bit of a description in the Pr? Even if it’s a copy paste

tommydcjung:
Software FP div or sqrt routine reads the floating-point rounding mode CSR at the beginning and stores in some register (e.g. rd = x18), using the frrm instruction. vanilla core treats as frrm as a bubble (no side effect), so whatever was previously written in x18 is used as the rounding mode. The fix is to decode frrm, and write zero into rd, so that the default mode can be chosen (round to nearest even). We just assume for now that the HW will only support the default mode.

tommydcjung:
Why do we see this bug now, and not before? according to @vb000, this bug should have emerged when we started compiling risc-v toolchain in RV32F mode, so that those software routine actually invokes frrm.

tommydcjung:
alternative solution would be to compile in the old way without frrm being generated. I would prefer that, because that wouldn't add extra hw.

dpetrisko:
> alternative solution would be to compile in the old way without frrm being generated. I would prefer that, because that wouldn't add extra hw.

I think that’s not “RISC-V compliant”. Unless it’s a huge hardware cost, probably better to fix in the new way

taylor-bsg:
I think Bandhav was playing with the floating point settings in GCC?

I think if all we need to do is write 0 to a register, it’s better to be
that much more RISC-V compliant, so we don’t lose time on debugging as more
and more libraries are used by the manycore. Can we add an assert that
prints the PC to check if unhandled cases exercised as well, so we can
track these issues down faster?

M

tommydcjung:
sounds good to me. I will add the assertion.

vb000:
Two improvements are made to the toolchain flow to better use our FPU hardware:
- Use FP hardware instructions (as opposed to integer emulations) for library implementation.
- Use FP abi (`ilp32f`) to utilize FP registers in the calling convention.

Both improvements are very much required for obvious code size and performance improvements. This particular issue has to do with the first improvement, where software emulation of fdiv and fsqrt uses floating point instructions, as opposed to integer instructions. Incidentally, fdiv and fsqrt emulations use `frrm` instruction to read the rounding mode, which is handled by HW as a `nop`. This failure could occur with any floating point program that explicitly use the rounding mode, so I think it's a right thing to implement it as a consistent 0 instead of leaving it as an x.

tommydcjung:
I don't think library implementation of div and sqrt actually makes use of floating-point add or multiply. It's that adding rv32f includes that floating-point rounding mode csr, and it gives an option to choose which mode to do div and sqrt.

tommydcjung:
That being said, I think it's worthwhile to build area-efficient fsqrt and fdiv, since these routines tend to be large. Or using fused-multiply-add with enough precision to do sqrt and fdiv in software. That would be pretty cool.

vb000:
> I don't think library implementation of div and sqrt actually makes use of floating-point add or multiply. It's that adding rv32f includes that floating-point rounding mode csr, and it gives an option to choose which mode to do div and sqrt.

Toolchain has to be compiled with rv32imaf not only because we would have performance improvements in instruction emulations, but mainly because library implementations like expf, sine, cosine etc. would be faster and smaller.

v/vanilla_bean/parameters.vh
```diff
@@ -181,6 +182,8 @@
 `define RV32_FLW_S `RV32_Itype(`RV32_LOAD_FP, 3'b010)
 `define RV32_FSW_S `RV32_Stype(`RV32_STORE_FP, 3'b010)
 
+`define RV32_FRRM {12'b10, 5'b00000, `RV32_CSRRS_FUN3, 5'b?????, `RV32_SYSTEM}
```
vb000: Would it be possible to define "csr" instruction class like this below, to be consistent with the style of our overall instruction decoding macros?
https://github.com/bespoke-silicon-group/bsg_manycore/blob/7c4059cee15e101f2455282d40c7ab7fad244a7b/v/vanilla_bean/parameters.vh#L70-L74

vb000:
@tommydcjung 
This fix considers only `frrm` instruction which makes me ask following question as I couldn't understand the reason...

What is the case for not having an FP csr? I can imagine a user level program checking exception flags. Would compiler avoid generating CSR instructions?

dpetrisko:
To be compliant with the RISC-V spec, we need to support the FPCSR with the
atomic read-modify-write operation and exception aggregation.

User-level libraries may check FP exception flags, especially something
math-y.

taylor-bsg:
I think the general wisdom is that fp exceptions, rounding modes etc are
pretty rare and only found in assembly FP routines, since this stuff was
not (is not?) standardized or portable in C. Basically, only non-portable
library code uses it.  Hence why I am suggesting having some assertions
that serve as “trip-wires” that tell us what features software is using.

M

vb000:
> I think the general wisdom is that fp exceptions, rounding modes etc are pretty rare and only found in assembly FP routines, since this stuff was not (is not?) standardized or portable in C. Basically, only non-portable library code uses it. Hence why I am suggesting having some assertions that serve as “trip-wires” that tell us what features software is using.

May be error packets? So that huge programs running only on fpga can be flagged...?

taylor-bsg:
Interesting idea.

I guess this is also a property that can be checked statically?

vb000:
Possible false negatives with static checks -- a legacy library with a function with unsupported instructions that'll never be called.

taylor-bsg:
That’s generally considered to be a feature not a bug — we want to know not
just if the software triggers a bug when we run it, but if it could for the
end user...

vb000:
Fair enough.

taylor-bsg:
I have generally been fond of the idea of having tiles package up
exceptions to be handled by other accelerator cores or the host though. You
could even imagine floating point divide or sqrt being handled that way.
But the 3 day waiting period on adding new hardware always applies. :-)

tommydcjung:
Added the assertion as requested. The full CSR support is deferred since the current FP hardware does not fully support exception cases / rounding modes. I'm currently planning on a major refactor using hardfloat modules.

tommydcjung:
actually, we might need to support exception flags now. I will provide a fix soon.

drichmond:
We should merge this PR at some point, right? So we can track it when hardfloat is added?

tommydcjung:
resolved with hardfloat

PR #184, opened 2019-10-21
Added a new enum value, e_mem_cfg_f1_dram

drichmond:
In F1 this enumeration value switches between the default memory model
(e_mem_cfg_default), with an AXI memory model, and the DRAM IP (with
more accurate timing)

### DIFF
--- a/testbenches/common/v/bsg_mem_cfg_pkg.v
+++ b/testbenches/common/v/bsg_mem_cfg_pkg.v
@@ -12,6 +12,7 @@ package bsg_mem_cfg_pkg;
   typedef enum bit [lg_max_cfgs-1:0] {
     e_mem_cfg_default    = 0
     , e_mem_cfg_infinite = 1
+    , e_mem_cfg_f1_dram  = 2
   } bsg_mem_cfg_e;
 
 endpackage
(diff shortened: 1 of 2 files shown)


tommydcjung:
Maybe, memory_system module is not the correct abstraction. We kinda decided to share this memory_system module to reuse in both places, maybe that was the bad idea?

In bsg_manycore, we need
1) infinite mem
2) vcache+AXI4
3) vcache+bsg_dmc
In F1,
1) infinite mem (cosim)
2) vcache+AXI4

AXI4 can be the fake memory, HBM, or F1 DRAM.

taylor-bsg:
This is worth thinking through. Being able to easily switch between memory
systems is very valuable. It is likely that we will support AIB (the
chiplet interface) as well. I think it’s good to get rid of a default value
in the enum since the emulation default is different than the architectural
default. Is it some kind of combination of cache X dram? Although AXI is
the interface and determines what Verilog is synthesized, it does not
specify what model to use for the ram?

M

drichmond:
I agree - this is kind of an ugly patch since it crosses repositories. But it is important, and time critical with Chun and Max both bringing memory system changes

PR #187, opened 2019-10-23
Toolchain updates

vb000:
- Added a configure option to Newlib for installing our dramfs bsp.
- Updated the compiler ABI to ilp32f with a patch to riscv-gcc
- All RTL regression tests except nbody are passing; failure is debugged and is being fixed in https://github.com/bespoke-silicon-group/bsg_manycore/pull/183

### DIFF
--- a/software/riscv-tools/Makefile
+++ b/software/riscv-tools/Makefile
@@ -84,6 +84,9 @@ checkout-repos:
 	git clone $(TOOLCHAIN_URL);
 	cd $(TOOLCHAIN_REPO) && git checkout $(TOOLCHAIN_VERSION);
 	cd $(TOOLCHAIN_REPO) && git submodule update --init --recursive -j 16;
+
+	# This patch is necessary if target abi is ilp32f because of the issue
+	# https://github.com/riscv/riscv-gnu-toolchain/issues/530
 	cd $(TOOLCHAIN_REPO)/riscv-gcc && git apply $(CURDIR)/riscv-gcc.patch
 
 replace-newlib:
--- a/software/riscv-tools/Makefile
+++ b/software/riscv-tools/Makefile
@@ -83,12 +83,14 @@ checkout-repos:
 	@rm -rf $(TOOLCHAIN_REPO);
 	git clone $(TOOLCHAIN_URL);
 	cd $(TOOLCHAIN_REPO) && git checkout $(TOOLCHAIN_VERSION);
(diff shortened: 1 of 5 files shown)


software/riscv-tools/Makefile
```diff
@@ -28,7 +28,7 @@
 .DEFAULT_GOAL := help
 
 TARGET_ARCH       := rv32imaf
-TARGET_ABI        := ilp32 # TODO: Test ilp32f for potential performance improvement
+TARGET_ABI        := ilp32f # TODO: Test ilp32f for potential performance improvement
```
drichmond: Seems like this todo might be unnecessary. Can you file an issue for it, if it still is?

vb000: Forgot to remove it.

vb000:
> Aside from one comment, looks good to me
> 
> Have you run it through regression on F1?

No, I only ran RTL regression in this repo.

vb000:
I can run f1 regression, but that'd set a precedent for every pr to bsg_manycore. I think it's not reasonable to be required to run on f1 for every bsg_manycore's commit because:
1. f1 hardware might not be same as bsg_manycore's hardware
2. bsg_manycore's changes don't go into bladerunner immediately.

Open to here your thoughts...

drichmond:
In general, I completely agree. I argue that this is a special case, because the tool chain is so critical to the functioning of F1 that it can’t be considered manycore specific.

RTL, yes. SPMD tests, yes. You’re absolutely right on those.

vb000:
This particular scenario, yes, I can run regression on f1. 

But in general, toolchain, rtl, bsg_manycore_lib and spmd tests are inter-dependent. For example, f isa extension in toolchain requires f  isa extension in hw. My thought is that since bladerunner is protected by precise commit hashes, it's better to keep bsg_manycore's development independent, as testbed to quickly try out new things. If somethings doesn't work, it just won't make it to bladerunner until fixed!

drichmond:
Yep, your reasoning is exactly mine. This is a special case, so it merits a special request (and special thanks).

drichmond:
To be clear though. I’m just asking for cosimulation regression

tommydcjung:
I think this is exactly why we need continuous integration. I got something started with help of @dpetrisko, but nowhere near fully operational yet. But I think it's worth an investment.

vb000:
Cosim regression passed!

You're right @tommydcjung, especially because we have two different environments.

Issue #122, opened 2019-08-01
naming


taylor-bsg:
https://github.com/bespoke-silicon-group/bsg_manycore/blob/56b7e4894e29fe8ded697541bd70d9173b1ddcc3/testbenches/common/v/memory_system.v#L8

missing bsg_nonsynth

taylor-bsg:
When an issue is closed — the manner of resolution should be written in the
issue.

Issue #121, opened 2019-08-01
rename enum for readability


taylor-bsg:
https://github.com/bespoke-silicon-group/bsg_manycore/blob/56b7e4894e29fe8ded697541bd70d9173b1ddcc3/testbenches/common/v/bsg_mem_cfg_pkg.v#L13

rename to something more descriptive, maybe _vcache instead of _default.

tommydcjung:
we will name this one e_mem_cfg_vcache_axi. and when we have a non-blocking cache ready, we will add e_mem_cfg_nbcache_dmc with bsg_dmc.

PR #190, opened 2019-10-23
Bsg manycore mem cfg

tommydcjung:
Basically, a refactor to divide up the configurable memory system into 3-levels of hierarchy to give flexibility to easily switch between different memory setups.

Refer to bsg_manycore_mem_cfg_pkg.v for more info.


tommydcjung:
https://github.com/bespoke-silicon-group/bsg_manycore/issues/121
https://github.com/bespoke-silicon-group/bsg_manycore/issues/122

v/bsg_manycore_vcache.v
```diff
@@ -0,0 +1,136 @@
+/**
+ *  bsg_manycore_vcache.v
+ *
+ */
+
+`include "bsg_manycore_packet.vh"
+
+module bsg_manycore_vcache
```
drichmond: How do you feel about calling this bsg_manycore_blocking_vcache? (That's correct, right?)

tommydcjung: That is correct. our coding style guides us to put the adjective after the noun, so bsg_manycore_vcache_blocking would be more appropriate.

drichmond: Sounds good to me

tommydcjung:
Ran regression for both infinite_mem and vcache_blocking test.

v/bsg_manycore_vcache.v
drichmond: In case it wasn't clear, I think you should rename this.

PR #192, opened 2019-10-26
non-blocking cache integration

tommydcjung:
- refactored bsg_manycore_endpoint.
- user can configure memory system between
1) e_infinite_mem
2) e_vcache_blocking_axi4_nonsynth_mem
3) e_vcache_non_blocking_axi4_nonsynth_mem

all regression pass.


machines/Makefile
```diff
@@ -42,6 +41,7 @@ define set_machine_variables
 	$(eval VCS_DEFINES += +define+BSG_MACHINE_MAX_EPA_WIDTH=${BSG_MACHINE_MAX_EPA_WIDTH})
 	$(eval VCS_DEFINES += +define+BSG_MACHINE_MEM_CFG=${BSG_MACHINE_MEM_CFG})
 	$(eval VCS_DEFINES += +define+BSG_MACHINE_BRANCH_TRACE_EN=${BSG_MACHINE_BRANCH_TRACE_EN})
+	$(eval VCS_DEFINES += +define+den2048Mb+sg5+x16+FULL_MEM)
```
drichmond: What does this do? Is this something I need to add in F1? What does it do?

tommydcjung: we don't need that in bsg_f1. this is for the lpddr4 simulation model. If we are including mobile_ddr.v to the filelist, these macros have to be defined, otherwise the tool will crash.

https://github.com/bespoke-silicon-group/basejump_stl/blob/master/testing/bsg_dmc/lpddr_verilog_model/mobile_ddr.v#L109

drichmond: Can you add a comment describing this in the makefile here?

drichmond: Thanks!

drichmond:
Do we want to change all the machine defaults to the non-blocking cache?

tommydcjung:
I think we want to change default to non-blocking cache. So people can use it more and find out the bug.

PR #195, opened 2019-10-29
Fix kernel_softmax and implement kernel_log_softmax

save-buffer:


save-buffer:
Passes on both cosim and F1

bornaehsani:
Looks good, great code! Cosim for softmax passed. I only have two comments: 
1. Could you add some comment at the beginning of the kernel codes, explaining what the kernel does? Not too detailed but a high-level overview would be useful for later use of this code.
2. For the log_softmax kernel, it'd be good to have a corresponding bsg_f1 host side regression test for it, just like the softmax. 
Other than that the codes looks good to me!

software/spmd/bsg_cuda_lite_runtime/log_softmax/kernel_log_softmax.c
```diff
@@ -0,0 +1,137 @@
+
+/*!
+ * This kernel performs softmax
```
bornaehsani: More detailed comment here would be useful.  Same with kernel_softmax.c

PR #196, opened 2019-10-31
Adding CI script for manycore regression

dpetrisko:
This CI setup builds the toolchain manually and runs the manycore regression.

Things missing (may or may not be important before merging):
- Checking success of toolchain build (One can manually trigger the build, but it will not fail, even if tools are not present)
- Add Badge to project README (to check status of builds, the CI status should still show up in PRs)

I've also only run this on a sucessful run.  I don't know now to easily force a CI fail, but it theoretically should fail if any test has BSG_FAIL or BSG_TIMEOUT and _no_ test has BSG_PASS (to detect build failures)

Before merging, we should restrict this to run on certain branches only.  This depends on the manycore development model.  Obviously we want to run on master, but probably not on all development branches, if they are frequently pushed.  If the model is people work in their own fork and then PR when ready, then running on all branches should be fine.  Since this is dependent on the manycore team, I'll leave it up to y'all to decide.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,0 +1,74 @@
+variables:
+  GIT_STRATEGY: none
+  GIT_SUBMODULE_STRATEGY: none
+
+stages:
+  - prereq
+  - check
+  - test-short
+  - test-long
+
+build-toolchain:
+  when: manual
+  stage: prereq
+  tags:
+    - bsg
+  before_script:
+    - rm -rf *
+    - echo "Cloning Manycore"
+    - git clone -b $CI_COMMIT_REF_NAME https://github.com/bespoke-silicon-group/bsg_manycore.git
+    - echo "Copying cadenv from local directory"
+    - cp -r $BSG_CADENV_DIR bsg_cadenv
+  script:
+    - echo "Building Manycore Toolchain."
+    - cd bsg_manycore
+    - make checkout_submodules
+    - make -C software/riscv-tools clean-all install-clean
+  cache:
+    key: $CI_COMMIT_REF_SLUG
+    paths:
+      - $CI_PROJECT_DIR/bsg_manycore/imports/
+      - $CI_PROJECT_DIR/bsg_manycore/software/riscv-tools/
+    policy: push
+
+
+test-regress:
+  stage: test-long
+  tags:
+    - bsg
+    - vcs
+  before_script:
+    - echo "This before_script is a HACK.  The cache automatically restores to bsg_manycore/*"
+    - echo "So we copy the cached tools, delete the dependencies, then restore the cache..."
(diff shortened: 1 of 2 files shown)


vb000:
Would a convention like pulling a new branch named `ci_<branch>` when `<branch>` is mature enough makes sense?

dpetrisko:
> Would a convention like pulling a new branch named ci_<branch> when <branch> is mature enough makes sense?

That's definitely one approach.  I'd suggest maybe reverse filtering like 'branches without dev'. 
 There's also a way to make it so that CI only triggers on PRs (but I haven't looked hard enough into it).

.gitlab-ci.yml
```diff
@@ -26,8 +25,14 @@ build-repos:
       - $CI_PROJECT_DIR/bsg_cadenv
       - $CI_PROJECT_DIR/basejump_stl
       - $CI_PROJECT_DIR/bsg_manycore
-      - $CI_PROJECT_DIR/bsg_f1
+      - $CI_PROJECT_DIR/bsg_bladerunner
     policy: push
+  only:
+    refs:
+      - merge_requests
```
dpetrisko: I don't think this will work.  Merge requests means gitlab merge requests, not github pull requests

.gitlab-ci.yml
```diff
@@ -3,72 +3,66 @@ variables:
...
+    - echo "Building toolchain"
+    - cd bsg_manycore/software/riscv-tools
+    - make clean-all install-clean
   cache:
     key: $CI_COMMIT_REF_SLUG
     paths:
-      - $CI_PROJECT_DIR/bsg_manycore/imports/
-      - $CI_PROJECT_DIR/bsg_manycore/software/riscv-tools/
+      - $CI_PROJECT_DIR/bsg_cadenv
+      - $CI_PROJECT_DIR/basejump_stl
```
dpetrisko: Do you want to cache basejump_stl?

.gitlab-ci.yml
```diff
@@ -3,72 +3,66 @@ variables:
...
+      - $CI_PROJECT_DIR/bsg_manycore
```
dpetrisko: This means that you need to rebuild the toolchain on every regression.  Is that what you want?

.gitlab-ci.yml
vb000: Oh! Worth a try :) I'm trying out few things...

dpetrisko: Yeah, it's a pain.  I think the way you have to do it is setting up a webhook on the GitHub where it pings the GitLab when a PR is created

vb000:
Closing this as these changes and more are incorporated in [#197](https://github.com/bespoke-silicon-group/bsg_manycore/pull/197).

PR #197, opened 2019-11-01
Manycore CI on master & branches starting with ci_

vb000:
**Outline**:

This flow does following things on every commit to master and any branch named starting with `ci_`:
- Builds toolchain (riscv-gcc and llvm) whenever something in `software/spmd/riscv-tools` is changed.
- Runs SPMD regression.
- Provides a way to manually run cosim regression, which is not currently working because aws-fpga repo is cloned using ssh url. But this should be easy to fix I believe.

**Suggested usage**:  

If you're developing on a branch called `mybranch`, please create a new branch called `ci_mybranch` based on `mybranch` to run CI and `mybranch`. You could keep working on `mybranch` for incremental updates and rebase `ci_mybranch` on `mybranch` when it's ready for another CI run.

**Improvements**:

Tweaking the flow is as easy as modifying any bash script. Following links would be helpful in understanding basic CI:
- [Intro](https://docs.gitlab.com/ee/ci/introduction/index.html)
- [CI script quick start](https://docs.gitlab.com/ee/ci/quick_start/README.html)
- [CI script reference](https://docs.gitlab.com/ee/ci/yaml/README.html)

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -3,72 +3,114 @@ variables:
   GIT_SUBMODULE_STRATEGY: none
 
 stages:
-  - prereq
-  - check
-  - test-short
-  - test-long
+  - checkout
+  - build
+  - test
 
-build-toolchain:
-  when: manual
-  stage: prereq
+build-repos:
+  stage: checkout
   tags:
     - bsg
-  before_script:
-    - rm -rf *
-    - echo "Cloning Manycore"
-    - git clone -b $CI_COMMIT_REF_NAME https://github.com/bespoke-silicon-group/bsg_manycore.git
-    - echo "Copying cadenv from local directory"
-    - cp -r $BSG_CADENV_DIR bsg_cadenv
   script:
-    - echo "Building Manycore Toolchain."
-    - cd bsg_manycore
-    - make checkout_submodules
-    - make -C software/riscv-tools clean-all install-clean
+    - echo "Cloning repos..."
+    - rm -rf bsg_cadenv basejump_stl bsg_manycore
+    - cp -r $BSG_CADENV_DIR bsg_cadenv
+    - git clone https://github.com/bespoke-silicon-group/basejump_stl.git
+    - git clone --recursive -b $CI_COMMIT_REF_NAME https://github.com/bespoke-silicon-group/bsg_manycore.git
+    - >
+      if [ -e cache/ ]; then
+        echo "Pulling toolchain installation from cache...";
+        cp -r cache/riscv-install bsg_manycore/software/riscv-tools/;
+        cp -r cache/llvm bsg_manycore/software/riscv-tools/;
+      else
+        echo "Building toolchain...";
+        make -C bsg_manycore/software/riscv-tools clean-all install-clean;
+        echo "Updating cache with latest toolchain";
+        mkdir -p cache/;
+        cp -r bsg_manycore/software/riscv-tools/riscv-install cache/;
+        cp -r bsg_manycore/software/riscv-tools/llvm cache/;
+      fi;
   cache:
     key: $CI_COMMIT_REF_SLUG
     paths:
-      - $CI_PROJECT_DIR/bsg_manycore/imports/
-      - $CI_PROJECT_DIR/bsg_manycore/software/riscv-tools/
+      - $CI_PROJECT_DIR/bsg_cadenv
+      - $CI_PROJECT_DIR/basejump_stl
+      - $CI_PROJECT_DIR/bsg_manycore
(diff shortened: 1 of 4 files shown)


drichmond:
This is great. I really like it. Do you have some time to sit down on Monday and teach me about how this all works?

vb000:
@drichmond Okay

Btw, this branch's CI run is expected to fail because of the FP issue. This is evident from the regression summary which can be seen here:
https://gitlab.com/dpetrisko/bsg_manycore/-/jobs/339855020

To open this: click ci "details" below to open the pipeline run, and then click "test-spmd" job on the pipeline page.

tommydcjung:
did fft pass?

dpetrisko:
One piece of advice would be to split out regression jobs by test group I.e. unit tests, fpu regression, benchmarks... Then you can tell which tests are causing the failure / allow certain test groups to fail for known issues

vb000:
Good point @tommydcjung; pushed a mod to see if we can report what happened to that!

@dpetrisko Each test in the regression is running independently. As is summarized in the `test-spmd` CI job log, we can see what tests failed...

vb000:
I believe now we have complete information in the CI log -- regression summary with lists of tests that finished successfully, did not finish successfully, finished with explicit fail packet and those that timed out. Also, all the error/fail messages are displayed about this summary.

Interestingly, there were few silent failure those were not showed on previous regress summary. I raised a separate issue to discuss them.

dpetrisko:
> Each test in the regression is running independently. As is summarized in the test-spmd CI job log, we can see what tests failed...

It's personal taste.  This way, CI is "all or nothing" pass/fail.  Since there are a set of failing tests right now, that means that people may just tune out of the results and not check them, so further failures are missed.  

If you have separate CI targets for test sets, then you can selectively disable them from the CI runs while fixes are developed.  Another benefit is that you can have short tests which run as smoke tests and then long tests which handle correctness regression and then long-long tests which do performance regression, so that you don't have to wait for the whole suite to finish figure out there's a bug.

vb000:
Yes, I agree. In fact I had implemented regression flow with different levels in one of the older versions of manycore. But that required the person adding a test to classify the test properly. And as a result there were many tests that were completely forgotten as they weren't updated in regression list.

To avoid those @proftaylor has implemented the current regression flow in which all the tests available in spmd directories are recursively run eliminating the possibility of missing some tests. I wouldn't call this "all or nothing" strategy -- each test is run in a separate make job and a result, individual failure doesn't affect other tests. As can be seen in the [ci log](https://gitlab.com/dpetrisko/bsg_manycore/-/jobs/339979075), we can see which tests passed and which ones failed; single failure doesn't stop the regression.

vb000:
And reviewers can see which tests caused CI failure and determine if those failures are okay for the branch to be merged now. But we would certainly want CI pass (the green tick mark) to mean all tests available in the repo run successfully.

dpetrisko:
First, let me say I definitely understand the motivation. Totally fine to keep it this way. 

> And reviewers can see which tests caused CI failure and determine if those failures are okay for the branch to be merged now.

Fair, although adding humans in loops is always troublesome. 

> But we would certainly want CI pass (the green tick mark) to mean all tests available in the repo run successfully.

Yeah so this is the all-or-nothing part. It’s totally valid to say the manycore development strategy is that all tests should always pass in regression before merging.  

But say there’s a known bug in the FPU which causes a single test to fail. We want to keep developing on this branch. Now, every time CI is run this forces a human in the loop to decide whether the red is caused by the existing bug or a newly introduced bug.  Whereas otherwise you can disable the set of tests known to fail in the development branch while still getting “green” on CI. Moreover, we can enforce a policy of never merging red. Of course, locally we can still run the all-recurse target. 

That said, even BlackParrot doesn’t do all this. This is just how I’ve seen it set up at companies. Food for thought!

vb000:
A nit-picky argument... if I understood correctly, I see two choices here:
1. Developer can remove acceptable failures from regress list to maintain the invariant that only green gets merged in.
2. Reviewer review case-by-case, which reds can be merged in.

Both need human in the loop. But failures in option 1 are hidden inside CI script, whereas failures in option 2 are wide open in "red". So, shouldn't we prefer 2?

dpetrisko:
Yep, that’s about right. It depends on if you trust individual developers or the devops =)

Usually the allowable failures are in some kind of a ‘waiver’ list, also checked in

vb000:
@drichmond Can we merge this if everything seems okay? We can take advantage of ci as soon as this merged!

vb000:
I rebased this on master... waiting for the ci to trigger for one last time before merging.

PR #198, opened 2019-11-01
Vanilla core profiler update and bloodgraph refactor - Ver 2.

bornaehsani:
Ver. 2 of PR #180.

- Modified vanilla_core_profiler.v to output the specific type of instruction being executed instead of printing (no_stall).
- Modified vanilla_core_profiler.v to distinguish between remote load stores to DRAM, Global tile (tile outside of tile group including host), and group tile. The same with remote load stalls.
- Modified bloodgraph accordingly. 
- Now using argparse for input arguments.
- Bloodgraph now has two modes, --detailed (which is default) and --abstract.
- @mrutt92  Added --generate-key option to generate key for the specific mode requested, within the same application.


software/py/blood_graph.py
```diff
@@ -3,48 +3,246 @@
...
+                          "fmv_x_w" }
+
+    # List of unknown operation by the core 
+    self.unknown_list  = {"unknown"}
+
+
+    # Coloring scheme for different types of operations
+    # For detailed mode 
+    # i_cache miss is treated the same is stall_ifetch_wait
+    self.detailed_stall_bubble_color = { "stall_depend"                           : (0x00, 0x00, 0x80), ## navy blue
```
drichmond: A thought - but hex colors aren't very intuitive. What if we used the Python Colour package (https://github.com/vaab/colour) or the ImageColor package from PIL (https://pillow.readthedocs.io/en/3.1.x/reference/ImageColor.html#color-names), and use the color names defined there. 

That would allow something like: 

```
stall_depend :       Color("NavyBlue")
stall_depend_local_load: Color("Cyan")
```
 or

```
stall_depend :       ImageColor.getrgb("NavyBlue")
stall_depend_local_load: ImageColor.getrgb("Cyan")
```

software/py/blood_graph.py
```diff
@@ -74,18 +302,19 @@ def get_tg_dim(self, traces):
     
     self.xdim = self.xmax-self.xmin+1
     self.ydim = self.ymax-self.ymin+1
+    return
 
 
-  # private function
+  # private method
   # initialize image
   def init_image(self):
```
drichmond: If something is private, you can add the prefix __ to the name, and make it (effectively) private.

e.g. `def __init_image`

testbenches/common/v/vanilla_core_profiler.v
```diff
@@ -285,16 +307,23 @@ module vanilla_core_profiler
...
-  // float_sb[1] : remote load
-  // float_sb[0] : local_load
+  // float_sb[3]: remote dram load
+  // float_sb[2]: remote global load
+  // float_sb[1]: remote group load
+  // float_sb[0]: local_load
   //
-  logic [31:0][1:0] int_sb_r;
-  logic [31:0][1:0] float_sb_r;
+  logic [31:0][3:0] int_sb_r;
```
drichmond: Is there not a package where these field widths are defined?

software/py/blood_graph.py
bornaehsani: Yeah in general it is more graspable, and I thought about using coloring names at first too. But the problem is there are too many colors, I needed between about 16 highly distinguishable colors and it was too hard to pick those using just your eye. So Michael came up with the idea of dividing the range of 0x00 to 0xff for each of the R,G,B colors into however many shades of that color we want, and generate the combination of those colors. I've done the same in the bloodgraph more or less, or at least tried to come close to it. I've used two shades of Red and Blue (0x00 & 0xff) and four shades of Green (0x00, 0x55, 0xaa, 0xff) to generate the colors. So the colors don't necessarily have a name, and I need to fiddle with the hex code too often to generate better colors so I'd rather it stay as hex coding. Why four shades of green you might wonder, because human eye can distinguish between more shades of Green than any other color :D

software/py/blood_graph.py
bornaehsani: Good point, I fixed all private methods with this. Thanks!

testbenches/common/v/vanilla_core_profiler.v
bornaehsani: Yes good point, fixed!

testbenches/common/v/vanilla_core_profiler.v
```diff
@@ -285,16 +307,23 @@ module vanilla_core_profiler
...
-  // float_sb[1] : remote load
-  // float_sb[0] : local_load
+  // float_sb[3]: remote dram load
+  // float_sb[2]: remote global load
+  // float_sb[1]: remote group load
+  // float_sb[0]: local_load
   //
-  logic [31:0][1:0] int_sb_r;
-  logic [31:0][1:0] float_sb_r;
+  logic [data_width_p-1:0][3:0] int_sb_r;
```
tommydcjung: This is not equal to data_width_p. Rather this line should be added to parameter.vh, and used instead.

localparam RV32_reg_els_gp = 32;

bornaehsani: My bad! and great catch. Fixed, thanks a lot!

software/py/blood_graph.py
```diff
@@ -3,68 +3,292 @@
 #
 #   vanilla_core execution visualizer.
 # 
-#   input: vanilla_stall_trace.log
+#   input: vanilla_operation_trace.log
```
drichmond: .log or .csv?

software/py/blood_graph.py
```diff
@@ -74,77 +298,99 @@ def get_tg_dim(self, traces):
...
+    elif trace["operation"] in self.instr_list:
+      self.pixel[col,row] = self.unified_instr_color
+    elif trace["operation"] in self.fp_instr_list:
+      self.pixel[col,row] = self.unified_fp_instr_color
+    elif trace["operation"] in self.unknown_list:
+      self.pixel[col,row] = self.unknown_color
+    else:
+      print ("Error: invalid operaiton in operation log: " + trace["operation"])
+      sys.exit()
+    return
```
drichmond: Raise an exception instead of calling sys.exit()

testbenches/common/v/vanilla_core_profiler.v
```diff
@@ -628,7 +732,7 @@ module vanilla_core_profiler
   // file logging
   //
   localparam logfile_lp = "vanilla_stats.log";
-  localparam tracefile_lp = "vanilla_stall_trace.log";
+  localparam tracefile_lp = "vanilla_operation_trace.log";
```
drichmond: .csv?

software/py/blood_graph.py
drichmond: or .trace

software/py/blood_graph.py
```diff
@@ -3,68 +3,292 @@
...
+import argparse
+from PIL import Image, ImageDraw, ImageFont
+from itertools import chain
 
 
-class BloodGraph:
+DEFAULT_START_CYCLE = 0
+DEFAULT_END_CYCLE   = 500000
+DEFAULT_MODE       = "detailed"
+DEFAULT_INPUT_FILE = "vanilla_operation_trace.log"
```
drichmond: I think you can make these class variables of the argument parser

v/vanilla_bean/parameters.vh
```diff
@@ -203,6 +203,7 @@
 
 localparam RV32_instr_width_gp    = 32;
 localparam RV32_reg_data_width_gp = 32;
+localparam RV32_reg_els_gp        = 32;
```
drichmond: Is this just the number of RISC-V registers?

software/py/blood_graph.py
bornaehsani: good point, thanks, done!

testbenches/common/v/vanilla_core_profiler.v
bornaehsani: changed to csv, thank!

software/py/blood_graph.py
bornaehsani: changed to csv, to be aligned with the vanilla_stats.csv. Thanks!

software/py/blood_graph.py
bornaehsani: Good idea, thanks a lot for pointing it out! Done!

v/vanilla_bean/parameters.vh
bornaehsani: yes it's used for the socreboard inside vanilla core profiler.

drichmond:
This doesn’t seem like a parameter. The RISC-v spec is only defined for 32 gp registers. Therefore my feeling is that it shouldn’t be in parameters.vh, because it isn’t configurable

Tommy? Thoughts?

v/vanilla_bean/parameters.vh
drichmond: This seems odd here - but I guess it's fine. (Parameters *to me* imply variables that can be changed, whereas constants do not)

taylor-bsg:
Interesting nuance.  I updated the style guide to address this question. I
think ordinarily we would call them constants, but SystemVerilog has
aliased the term constant to parameter.  _gp (=global parameter) means they
have global scope and are not localized to the module. Like in C, global
scope constants are usually fairly dangerous to change...

m

v/vanilla_bean/parameters.vh
dpetrisko: Nit: RV32E has 16 registers. It’s not inconceivable that vanilla core would support RV32E as an extension

PR #199, opened 2019-11-01
Vanilla Stats Parser Ver.2

bornaehsani:
Vanilla Stats Parser Ver.2 replacement of PR #181 .
The vanilla stats parser takes the vanilla_stats.log file as input and prints out per tile and total statistics in separate files. 
The VanillaStatsParser class returns timing stats, aggergate manycore stats and per tile stats dictionaries. 
Stats include:
- Execution time, for each tile group and aggregated total
- Instruction count and ratio for each tile and total for all types of instructions
- Stall cycle count and ratio for each tile and total for all types of stalls
- Miss count and ratio for each tile and total for all types of miss
Useful when precise numbers and ratios are needed, compared to bloodgraph's high-level view.


bornaehsani:
@drichmond regarding the use of sets vs. lists, I needed to print stats in the same order as defined in the program, so I went with lists here.

software/py/vanilla_stats_parser.py
```diff
@@ -0,0 +1,472 @@
...
+    self.manycore_dim = manycore_dim_y * manycore_dim_x
+    self.per_tile_stat = per_tile_stat
+
+
+    self.max_tile_groups = 1024
+    self.num_tile_groups = 0
+
+    self.timing_start_list = [0] * self.max_tile_groups 
+    self.timing_end_list =   [0] * self.max_tile_groups
+#    self.timing_stat = []
```
drichmond: Remove

software/py/vanilla_stats_parser.py
```diff
@@ -0,0 +1,472 @@
...
+
+    self.total_execution_time = 0
+    self.total_instr_cnt = 0
```
drichmond: if you want counters, there's a cool dict subclass called Counter (`from collections import Counter`)

software/py/vanilla_stats_parser.py
```diff
@@ -0,0 +1,472 @@
...
+    self.total_stall_cnt = 0
+    self.total_miss_cnt = 0
+
+    self.tile_stat_dict = [[dict() for x in range(self.manycore_dim_x)] for y in range(self.manycore_dim_y)]
+    self.manycore_stat_dict = dict()
+
+
+
+  # private method
```
drichmond: See comment in #198

drichmond: Regarding private methods

software/py/vanilla_stats_parser.py
```diff
@@ -0,0 +1,472 @@
...
+
+import sys
+import argparse
+import os
+import re
+from enum import Enum
+import csv
+
+# These values are used by the manycore library in bsg_print_stat instructions
+# They are added to the tag value to detremine wether the message is from the
```
drichmond: typos in `determine and wether`

software/py/vanilla_stats_parser.py
```diff
@@ -0,0 +1,472 @@
...
+
+
+  # private method
+  # Sum up all other stats for all tiles based on the last bsg_print_stat instr
+  def generate_stats_all(self):
+      # Generate timing stats 
+      for trace in self.traces:
+        self.generate_stats_timing(trace)
+
+      self.timing_stat = [0] * self.num_tile_groups
```
drichmond: Why are you initializing a object member in a non-init method?

software/py/vanilla_stats_parser.py
```diff
@@ -0,0 +1,472 @@
...
+  # print execution timing for the entire manycore 
+  def print_manycore_stats_timing(self, stat_file):
+    # For total execution time, we only sum up the execution time of origin tile in tile groups
+    # The origin tile's relative coordinates is always 0,0 
+    stat_file.write("Timing Stats\n")
+    stat_file.write(stats_timing_header_format.format("tile group", "exec time", "share (%)"))
+    stat_file.write("{}".format(separating_line))
+
+    # Sum up execution time for all tile groups 
+    for i in range (0, self.num_tile_groups):
```
drichmond: I think you can just call `sum(self.timing_stat)`

software/py/vanilla_stats_parser.py
```diff
@@ -0,0 +1,472 @@
...
+      self.total_execution_time += self.timing_stat[i]
```
drichmond: Similar to above - why are you setting an instance variable inside of a non_init method? Would it be better to return it, and then use it in other methods?

software/py/vanilla_stats_parser.py
```diff
@@ -0,0 +1,472 @@
...
+# Default input values
+DEFAULT_MANYCORE_DIM_Y = 4
+DEFAULT_MANYCORE_DIM_X = 4
+DEFAULT_MODE = "total"
+DEFAULT_INPUT_FILE = "vanilla_stats.log"
+
+
+
+# formatting parameters for aligned printing
+name_length = 35
```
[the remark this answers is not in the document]
bornaehsani: Good point! I made a function to use a formatting dictionary based on the stats type to print. How does it look now?

software/py/vanilla_stats_parser.py
bornaehsani: Because I didn't know the size of the object I needed at init time, but I'll go with the max size possible, it's not like I'm saving that much space anyway. Thanks!

drichmond: You can also use default_dict, or something similar for lists.

software/py/vanilla_stats_parser.py
bornaehsani: Good point!

bornaehsani:
@drichmond  Thanks for your detailed review! I tried to take care of minor comments as best as I could. On the bigger issue, I understand your view on the whole program flow and implicit argument passing, I made a couple changes and introduced new functions for explicit argument passing in the main body of the generate_stats() main method. How does it look now?

software/py/vanilla_stats_parser.py
bornaehsani: Good point, I fixed it along with the program flow. How does it look now?

software/py/vanilla_stats_parser.py
```diff
@@ -0,0 +1,640 @@
...
+# they are added to the tag value to determine the tile group that triggered the stat
+# and also the type of stat (stand-alone stat, start, or end)
+# the value of these paramters should match their counterpart inside 
+# bsg_manycore/software/bsg_manycore_lib/bsg_manycore.h
+# Formatting for bsg_cuda_print_stat instructions
+# < stat_type >
+# Section                 Stat type  -   y cord   -   x cord   -    tile group id   -        tag
+# of bits                <----2----> -   <--6-->  -   <--6-->  -   <------10----->  -   <-----8----->
+# Stat type value: {"stat":0, "start":1, "end":2}
+bsg_STAT_TAG_BITS   = 8
```
drichmond: A more object-oriented way to do this is to define a CudaTag class. When you read in the CSV, read each tag value and initialize it as a CudaTag object.

Specifically, these variables become class fields (i.e. not initialized in __init__). 

That way, CudaTags can be queried for different properties. For example tag.isStart() can be used to determine if a tag is a start tag. Ditto for finish, or just a normal tag, or the tile group ID.

software/py/vanilla_stats_parser.py
```diff
@@ -0,0 +1,640 @@
...
+bsg_STAT_TAG_MASK   = ((1 << bsg_STAT_TAG_BITS) - 1)
+bsg_STAT_TG_ID_MASK = ((1 << bsg_STAT_TG_ID_BITS) - 1)
+bsg_STAT_X_MASK     = ((1 << bsg_STAT_X_BITS) - 1)
+bsg_STAT_Y_MASK     = ((1 << bsg_STAT_Y_BITS) - 1)
+bsg_STAT_TYPE_STAT  = 0
+bsg_STAT_TYPE_START = 1
+bsg_STAT_TYPE_END   = 2
+
+
+bsg_TILE_GROUP_ORG_X = 0
```
drichmond: Are these still used?

software/py/vanilla_stats_parser.py
```diff
@@ -0,0 +1,640 @@
...
+bsg_TILE_GROUP_ORG_Y = 1
+bsg_ORG_X = 0
+bsg_ORG_Y = 1
+
+# Default input values
+DEFAULT_MANYCORE_DIM_Y = 4
+DEFAULT_MANYCORE_DIM_X = 4
+DEFAULT_INPUT_FILE = "vanilla_stats.log"
```
drichmond: Can you switch to vanilla_stats.csv

software/py/vanilla_stats_parser.py
```diff
@@ -0,0 +1,640 @@
...
+
+  # print a line of stat into stats file based on stat type
+  def __print_stat(self, stat_file, stat_type, *argv):
+    stat_file.write(self.print_format[stat_type].format(*argv));
+    return
+
+
+  # Decodes the tag value of the stat to determine the type of 
+  # stat, the tile group id of sending tile, and the tag value 
+  def __decode_stat_val(self, stat_val):
```
drichmond: This would be replaced by CudaTag (if you decide to do it)

software/py/vanilla_stats_parser.py
```diff
@@ -0,0 +1,640 @@
...
+        if (item.startswith('instr_')):
+          if (not item == 'instr_total'):
+            self.instr_list += [item]
+        elif (item.startswith('miss_')):
+          self.miss_list += [item]
+        elif (item.startswith('stall_')):
+          self.stalls_list += [item]
+        else:
+          self.stats_list += [item]
+      self.all_ops_list = self.stats_list + self.instr_list + self.miss_list + self.stalls_list
```
drichmond: To be honest, I'd do all of this in __init__, since it's O(1), and done every time. Then you don't have to worry about incorrect object initialization/object-function-call ordering

software/py/vanilla_stats_parser.py
```diff
@@ -0,0 +1,640 @@
...
+    stat_tag    = (stat_val) & bsg_STAT_TAG_MASK
+    return (stat_type, stat_y, stat_x, stat_tg_id, stat_tag)
+    
+
+
+  # go though the input traces and extract start and end stats  
+  # for each tile, and each tile group 
+  # return number of tile groups, tile group timing stats, and the tile stats
+  # this function only counts the portion between two print_stat_start and end messages
+  # in theory, this excludes the time tiles are waiting to be loaded, etc.
```
drichmond: "In theory"?

software/py/vanilla_stats_parser.py
```diff
@@ -0,0 +1,640 @@
...
+    tile_stat       = [[Counter() for x in range(self.manycore_dim_x)] for y in range(self.manycore_dim_y)]
+
+    tile_group_stat_start = [Counter() for tg_id in range(self.max_tile_groups)] 
+    tile_group_stat_end   = [Counter() for tg_id in range(self.max_tile_groups)] 
+    tile_group_stat       = [Counter() for tg_id in range(self.max_tile_groups)] 
+
+    for trace in traces:
+      y = trace["y"]
+      x = trace["x"]
+      relative_y = y - bsg_ORG_Y
```
drichmond: Is it worth worrying about relative x/y coordinates? Why not just print absolute X/Y coordinates?

software/py/vanilla_stats_parser.py
```diff
@@ -0,0 +1,640 @@
...
+bsg_STAT_TYPE_END   = 2
+
+
+bsg_TILE_GROUP_ORG_X = 0
+bsg_TILE_GROUP_ORG_Y = 1
+bsg_ORG_X = 0
+bsg_ORG_Y = 1
+
+# Default input values
+DEFAULT_MANYCORE_DIM_Y = 4
```
drichmond: Why not make these a required argument?

software/py/vanilla_stats_parser.py
```diff
@@ -0,0 +1,640 @@
...
+        trace = {}
+        for stat in self.stats_list:
+          trace[stat] = int(row[stat])
+        for instr in self.instr_list:
+          trace[instr] = int(row[instr])
+        for stall in self.stalls_list:
+          trace[stall] = int(row[stall])
+        for miss in self.miss_list:
+          trace[miss] = int(row[miss])
+        self.traces.append(trace)
```
drichmond: Can you explain what's going on here?

testbenches/common/v/vanilla_core_profiler.v
```diff
@@ -1016,25 +1032,28 @@ module vanilla_core_profiler
           $fclose(fd2);
         end
     
-
-        if (~reset_i & print_stat_v_i) begin
+        // only print stat if core's x,y coordinates matches the ones in the
+        // print_stat_tag_i section of the print_stat
+        if (~reset_i & print_stat_v_i
+             & my_y_i == print_stat_tag_i[STAT_TAG_Y_START_BIT:STAT_TAG_Y_END_BIT]
+             & my_x_i == print_stat_tag_i[STAT_TAG_X_START_BIT:STAT_TAG_X_END_BIT]) begin
```
drichmond: What's this for?

software/py/vanilla_stats_parser.py
drichmond: @tommydcjung Do we have a standard naming for bit-width fields? 

@bornaehsani I feel that the better naming here is _width

testbenches/common/v/vanilla_core_profiler.v
```diff
@@ -1141,7 +1149,18 @@ module vanilla_core_profiler
...
+            stat.beq_miss,
+            stat.bne_miss,
+            stat.blt_miss,
+            stat.bge_miss,
+            stat.bltu_miss,
+            stat.bgeu_miss,
+            stat.jalr_miss
+          );
+     
           $fwrite(fd, "%0d,%0d,%0d,%0d,%0d,%0d,%0d,%0d,%0d,%0d,%0d,%0d,%0d,%0d,",
```
drichmond: There's an extra comma on this line that shouldn't be here. It's messing with CSV parsers

testbenches/common/v/vanilla_core_profiler.v
```diff
@@ -14,6 +14,10 @@ module vanilla_core_profiler
     , parameter reg_addr_width_lp = RV32_reg_addr_width_gp
     , parameter reg_els_lp = RV32_reg_els_gp
     , parameter dmem_addr_width_lp=`BSG_SAFE_CLOG2(dmem_size_p)
+    , parameter STAT_TAG_Y_START_BIT = 29
```
drichmond: See the BSG Style guide for parameter naming - https://docs.google.com/document/d/1xA5XUzBtz_D6aSyIBQUwFk_kSUdckrfxa2uzGjMgmCU/edit

testbenches/common/v/vanilla_core_profiler.v
```diff
@@ -734,6 +738,7 @@ module vanilla_core_profiler
   localparam logfile_lp = "vanilla_stats.log";
```
drichmond: Change to `vanilla_stats.csv`

testbenches/common/v/vanilla_core_profiler.v
```diff
@@ -734,6 +738,7 @@ module vanilla_core_profiler
...
   localparam tracefile_lp = "vanilla_operation_trace.log";
 
+
```
drichmond: Line added

software/bsg_manycore_lib/bsg_manycore.h
```diff
@@ -138,4 +138,40 @@ inline void bsg_fence()      { __asm__ __volatile__("fence" :::); }
 #define bsg_print_stat(tag) do { bsg_remote_int_ptr ptr = bsg_remote_ptr_io(IO_X_INDEX,0xd0c); *ptr = tag; } while (0)
 
 
+
+#define BSG_CUDA_PRINT_STAT_ID         0
+#define BSG_CUDA_PRINT_STAT_START_ID   1
+#define BSG_CUDA_PRINT_STAT_END_ID     2
+#define bsg_cuda_print_stat(tag) do {  int tg_id = __bsg_tile_group_id_y * __bsg_grid_dim_x + __bsg_tile_group_id_x;  \
+                                       int val = ( (BSG_CUDA_PRINT_STAT_ID << 30)                |                    \
```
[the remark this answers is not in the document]
bornaehsani: Done!

software/py/vanilla_stats_parser.py
bornaehsani: No, removed, thanks for pointing it out!

software/py/vanilla_stats_parser.py
bornaehsani: Done, thanks!

testbenches/common/v/vanilla_core_profiler.v
bornaehsani: Great catch! Fixed, thanks!

testbenches/common/v/vanilla_core_profiler.v
bornaehsani: done, thanks!

testbenches/common/v/vanilla_core_profiler.v
bornaehsani: fixed, thanks!

software/py/vanilla_stats_parser.py
bornaehsani: I need relative because of indexing into an array. If dimensions are 4x4, I'm instantiating a 4x4 array, but if the absolute X/Y is between 1 to 5, then it goes out of index.

software/py/vanilla_stats_parser.py
bornaehsani: done!

bornaehsani: Done!

software/py/vanilla_stats_parser.py
bornaehsani: It's reading the trace for every type of operations (instruction, miss, stall, stats) and putting them into a trace line. But I made it shorter, how does it look now?

testbenches/common/v/vanilla_core_profiler.v
bornaehsani: I incorporated the X,Y coordinates of the tile that is calling bsg_cuda_print_stat into the tag message, and the profiler checks it's own X/Y coordinates against the one incorporated in the message, and only prints a stat line if they match. This way, I can ensure only the tile that calls the print_stat message actually prints a stat, not all the tiles in the manycore. I needed to do this to identify the tiles and which tile group they belong to.

bornaehsani: See the comment at the top of vanilla_stats_parser for how I incorporate the X/Y coordinates, and the tile group ID into the tag message.

software/py/vanilla_stats_parser.py
bornaehsani: Changed it to in practice :D

testbenches/common/v/vanilla_core_profiler.v
bornaehsani: Good pointer, thanks! I added parameters to the parameters.vh file, and added local parameters trying to stick to the naming rules inside the vanilla_core_profiler. Let me know how it looks, thanks!

software/py/vanilla_stats_parser.py
bornaehsani: The entire method is called inside the init() though, do you mean just scrap the function and move it's body inside the init()? that'd be a bit dirty wouldn't it? I can make the method private though, to make sure nobody calls it from outside to disrupt the function call ordering, what do you think?

v/vanilla_bean/parameters.vh
```diff
@@ -235,4 +235,18 @@ localparam RV32_Jimm_width_gp     = 20;
 
 
 
+/////////////////////////////////////////////////////
+// The CUDA Print stat parameters
+// The cuda print stat tag message incorporates within itself the x,y coordiantes
+// and the tile group id of the tile that triggerd the bsg_cuda_print_stat instruction
+// <stat_type>   -   <y cord>   -   <x cord>   -   <tile group id>   -   <tag>
+localparam RV32_cuda_print_stat_tag_width_gp    = 8;
```
tommydcjung: parameters.vh is only for hardware parameters related to vanilla core, I wouldn't put this here.
Also, instead of having these lists of parameters, how about just defining a fixed-width struct?

bornaehsani: That's a good idea, I'm not sure how to do it correctly though. Let's discuss this tomorrow, thanks!

software/py/vanilla_stats_parser.py
bornaehsani: That's a great idea, I created a CudaStatTag class to be used this way. I also changed instances of _BITS to _WIDTH in parameters. Let me know what you think!

software/py/vanilla_stats_parser.py
```diff
@@ -0,0 +1,656 @@
...
+
+    tile_group_stat_start = [Counter() for tg_id in range(self.max_tile_groups)] 
+    tile_group_stat_end   = [Counter() for tg_id in range(self.max_tile_groups)] 
+    tile_group_stat       = [Counter() for tg_id in range(self.max_tile_groups)] 
+
+    for trace in traces:
+      y = trace["y"]
+      x = trace["x"]
+      relative_y = y - BSG_ORG_Y
+      relative_x = x - BSG_ORG_X
```
drichmond: Let's use ORIGIN instead of ORG.

software/py/vanilla_stats_parser.py
```diff
@@ -0,0 +1,656 @@
...
+    return ((self.stat_val >> (BSG_STAT_TAG_WIDTH + BSG_STAT_TG_ID_WIDTH + BSG_STAT_X_WIDTH)) & BSG_STAT_Y_MASK)
+  def get_type(self):
+    return ((self.stat_val >> (BSG_STAT_TAG_WIDTH + BSG_STAT_TG_ID_WIDTH + BSG_STAT_X_WIDTH + BSG_STAT_Y_WIDTH)))
+
+  def isStart(self):
+    return (self.get_type() == BSG_STAT_TYPE_START)
+  def isEnd(self):
+    return (self.get_type() == BSG_STAT_TYPE_END)
+  def isStat(self):
+    return (self.get_type() == BSG_STAT_TYPE_STAT)
```
drichmond: If you really want to go full Python, see the @property decorator:  https://docs.python.org/3/library/functions.html#property

Then you can do cst.tag, csg.tg_id, cst.x, cst.y, cst.isStart, etc.

drichmond: (This isn't necessary, but is a useful feature sometimes)

software/py/vanilla_stats_parser.py
```diff
@@ -0,0 +1,656 @@
...
+BSG_STAT_X_WIDTH     = 6
+BSG_STAT_Y_WIDTH     = 6
+BSG_STAT_TYPE_WIDTH  = 2
+BSG_STAT_TAG_MASK   = ((1 << BSG_STAT_TAG_WIDTH) - 1)
+BSG_STAT_TG_ID_MASK = ((1 << BSG_STAT_TG_ID_WIDTH) - 1)
+BSG_STAT_X_MASK     = ((1 << BSG_STAT_X_WIDTH) - 1)
+BSG_STAT_Y_MASK     = ((1 << BSG_STAT_Y_WIDTH) - 1)
+BSG_STAT_TYPE_STAT  = 0
+BSG_STAT_TYPE_START = 1
+BSG_STAT_TYPE_END   = 2
```
drichmond: These should be class variables of CudaStatTag

software/py/vanilla_stats_parser.py
drichmond: I mean, parsing the field names. 

To summarize what I find a weird is that your main "public" method is called in init and does initialization. Those are two distinct operations (initialization, and processing). It sets instance variables (self.tile_stat, self.tile_group_stat), and returns the same values. 

I think it should be broken down to be less monolithic - the functionality may be the same, but it'll be easier to understand what initialization means. 

For example, one way to break this down is to initialize all variables inside of __init__ (you do this now), and then call __parse_ops_list (which returns the parsed CSV header) and returns (not sets) the list of ops (which are set inside of __init__). Then you call __parse stats/__generate_tile stats, which returns (not sets) the tile stats, etc...


To summarize what's weird, is that this function sets or modifies instance variables, but isn't an init method, and isn't a method prepended with __init (to indicate initialization)

drichmond:
Also, if you can use four spaces instead of two that would be 👍

drichmond:
See the (not working) demo_parser branch to see some examples of what I described here

machines/sim_filelist.mk
```diff
@@ -2,6 +2,7 @@
 # simulation. These augment the sythesizable files in core.include.
 
 VHEADERS += $(BSG_MANYCORE_DIR)/testbenches/common/v/bsg_manycore_mem_cfg_pkg.v
+VHEADERS += $(BSG_MANYCORE_DIR)/testbenches/common/v/bsg_manycore_profile_pkg.v
```
drichmond: You'll need to file a separate PR for these in F1 or include them in arch_filelist.mk

bornaehsani: Yes thanks for letting me know, will do once we plan to move forward with this PR.

v/vanilla_bean/parameters.vh
bornaehsani: Fixed, thanks to Tommy!

software/py/vanilla_stats_parser.py
bornaehsani: I store your code, thanks! :D

software/py/vanilla_stats_parser.py
bornaehsani: Stole your code, thanks! :D

bornaehsani:
> 
> 
> Also, if you can use four spaces instead of two that would be 👍

Done!

software/spmd/bsg_cuda_lite_runtime/float_matrix_mul_shared_mem/kernel_float_matrix_mul_shared_mem.c
```diff
@@ -99,6 +99,8 @@ void __attribute__ ((noinline)) subblock_shmem_matrix_mul_xposed (float *sh_A, f
 int  __attribute__ ((noinline)) kernel_float_matrix_mul_shared_mem(float *A, float *B, float *C, int M, int N, int P, int block_size_y, int block_size_x) {
 
 	
+	bsg_cuda_print_stat_start(0);
+
```
drichmond: Did you mean to include these?

v/vanilla_bean/parameters.vh
```diff
@@ -233,6 +233,4 @@ localparam RV32_Jimm_width_gp     = 20;
 `define         CSR_TGO_X              1
 `define         CSR_TGO_Y              2
 
-
-
```
drichmond: Removed lines?

software/spmd/bsg_cuda_lite_runtime/float_matrix_mul_shared_mem/kernel_float_matrix_mul_shared_mem.c
bornaehsani: no just didn't want to add them every time to test. Removed now. But it wouldn't be the worst thing to have on them by default, would it?

drichmond: No, but it would be an overhead in regression. If we have them in all kernels, I would request that we have an on/off flag in the build flow.

v/vanilla_bean/parameters.vh
bornaehsani: fixed. thanks!

bornaehsani:
Deprecated. Replaced with PR #206.

PR #209, opened 2019-12-13
bsg_manycore_atomic

tommydcjung:


mrutt92:
does this require a specific version of basejump?

tommydcjung:
> does this require a specific version of basejump?

https://github.com/bespoke-silicon-group/basejump_stl/pull/196

v/bsg_manycore_pkg.v
```diff
@@ -15,79 +16,130 @@ package bsg_manycore_pkg;
...
-    , e_remote_swap_aq
-    , e_remote_swap_rl
+    , e_remote_amo
   } bsg_manycore_packet_op_e;
 
-
-  typedef enum logic {
+  typedef enum logic [1:0] {
     e_return_credit
-    , e_return_data
```
mrutt92: is this supposed to be replaced by e_return_credit?

tommydcjung: there are other return types.

mrutt92: nm

mrutt92:
can you please update bsg_manycore_link_sif_async_buffer.v? it is used by external projects.

v/bsg_manycore_pkg.v
```diff
@@ -15,79 +16,130 @@ package bsg_manycore_pkg;
...
     e_remote_load
     , e_remote_store
-    , e_remote_swap_aq
-    , e_remote_swap_rl
+    , e_remote_amo
   } bsg_manycore_packet_op_e;
 
-
-  typedef enum logic {
+  typedef enum logic [1:0] {
```
mrutt92: please add comments.

v/bsg_manycore_pkg.v
```diff
@@ -15,79 +16,130 @@ package bsg_manycore_pkg;
...
     e_return_credit
-    , e_return_data
+    , e_return_int_wb
+    , e_return_float_wb
+    , e_return_ifetch
   } bsg_manycore_return_packet_type_e;
 
+  typedef struct packed {
```
mrutt92: please add comments.

v/bsg_manycore_pkg.v
```diff
@@ -15,79 +16,130 @@ package bsg_manycore_pkg;
...
+    logic float_wb;
+    logic icache_fetch;
+    logic is_unsigned_op;
+    logic is_byte_op;
+    logic is_hex_op;
```
mrutt92: what is a hex_op?

mrutt92: I would add a comment stating that this is a "half" op.

dpetrisko: Any reason not to rename to half? That’s the Risc-v parlance

tommydcjung: That's how it was. bsg_manycore parlance

dpetrisko: I don’t have a horse in this race, but there are a heck of a lot more people who speak Risc-v than bsg_manycore

taylor-bsg: 😼

machines/4x4/Makefile.machine.include
```diff
@@ -8,5 +8,5 @@ BSG_MACHINE_DRAM_SIZE_WORDS           = 536870912  # 2^29 words (2GB)
 BSG_MACHINE_DRAM_BANK_SIZE_WORDS      = 134217728
 BSG_MACHINE_DRAM_INCLUDED             = 1
 BSG_MACHINE_MAX_EPA_WIDTH             = 28
-BSG_MACHINE_MEM_CFG                   = e_vcache_non_blocking_axi4_nonsynth_mem
```
mrutt92: does this work with non-blocking?

software/spmd/vcache_atomic_histogram/main.c
```diff
@@ -0,0 +1,142 @@
...
+23,30,10,14,23,21,8,21,9,11,17,23,4,0,20,0,1,7,20,2,1,31,13,28,6,13,11,5,10,31,23,12,
+13,8,17,23,21,14,1,29,13,12,0,4,27,16,23,4,10,26,26,7,0,25,5,25,21,5,2,29,19,19,14,4,
+19,8,25,23,0,29,1,2,9,4,7,11,23,12,8,15,12,9,28,17,31,24,18,8,21,14,23,12,15,0,23,1,
+21,18,24,9,21,6,16,10,31,31,6,18,10,30,29,18,23,21,11,29,28,10,23,0,26,18,30,11,20,10,25,19,
+31,0,4,1,4,29,30,15,30,26,24,23,6,17,13,30,5,5,21,5,3,16,25,19,24,8,9,13,31,22,30,17
+};
+
+INIT_TILE_GROUP_BARRIER(r_barrier, c_barrier, 0, bsg_tiles_X-1, 0, bsg_tiles_Y-1);
+
+void do_histogram_work()
```
mrutt92: When running this test as a CUDA program from `bsg_f1`, if I remove the locks from `do_histogram_work()` this test passes. Which is surprising. 

I think the issue is that the code in this test is "too good." Algorithmically, it's much more efficient for each core to lock once and batch the updates. But for the purposes of exercising the AMO feature, I suggest that we replace this function with the following:
```c
void do_histogram_work()
{
  int work = N/(bsg_tiles_X * bsg_tiles_Y);
  int start_idx = __bsg_id * work;

  // get the lock
  int lock_val = 1;

  for (int i = 0; i < work; i++) {
    int local_data = data[start_idx+i];
    // update global histogram
    // acquire lock
    do {
      lock_val = bsg_amoswap_aq(&lock, 1); }
    while (lock_val != 0);

    histogram[local_data] += 1;
    
    // release the lock
    bsg_amoswap_rl(&lock, 0);
  } 
}
```
This will increase the lock contention and decrease the likelihood of a "false-positive" when testing.

machines/4x4/Makefile.machine.include
mrutt92: nm; I see from email that blocking cache was specified.

software/spmd/vcache_atomics/main.c
```diff
@@ -0,0 +1,57 @@
+#include "bsg_manycore.h"
```
mrutt92: is this test meant to be run in single threaded execution?

mrutt92: It's not a problem if so, I'm just asking to know.

software/spmd/vcache_atomic_histogram/main.c
tommydcjung: This test helped find some bug in vanilla_core, so I think it has some merit. vcache_atomic_inc will exercise the lock contention you are talking about. But feel free to add more tests that you think are necessary.

software/spmd/vcache_atomics/main.c
tommydcjung: Yes

software/spmd/vcache_atomic_histogram/main.c
mrutt92: Just because it helped find a bug doesn't mean it can't be improved. What bug does the unmodified version uncover that would not be caught by above?

tommydcjung: I don't know whether the modified test will uncover the same bug, so I am hesitant to change it.

mrutt92: This test passed for me when I was (accidentally) using the non-blocking cache. Which, by my understanding, would not have happened if this were truly exercising the AMO functionality. That's why I'm skeptical of this test's merits.
 
But I understand your hesitation so... does vcache_atomic_inc have sufficient lock contention? It looks like each core just acquires the lock once and adds a value? 
Does that test send a fail packet when AMOs are not working with a 4x4?

tommydcjung: it will timeout most likely.

tommydcjung:
@bornaehsani please review the changes, and update the profiler codes. Let me know if you have questions.

software/spmd/vcache_atomic_histogram/main.c
mrutt92: OK; fyi that doesn't play well with our current f1 testbench (though we should probably fix that). But since you added the posted test, I don't think it's a big deal.

PR #211, opened 2020-01-12
remove bsg_fpu_pkg;

tommydcjung:


drichmond:
Just curious, why was this necessary?

tommydcjung:
https://github.com/bespoke-silicon-group/basejump_stl/issues/211

Since this package only contained macro defines, so I think some tool made complaints. However, if were to add some other constructs (struct, enum), we would want to package it up with "bsg_fpu_defines.vh".

taylor-bsg:
This popped up out of sv-tests.  It may have been a misconfiguration of the
tool they were running (in terms of which files to preload), but I found
the code disturbing because it seemed to imply that defines can be made
part of a package, while they are actually global.

M

taylor-bsg:
It is true that we abuse this a lot when we do parameterized structures
though.

dpetrisko:
>It is true that we abuse this a lot when we do parameterized structures
though.

Indeed.  It's a tough one

PR #218, opened 2020-01-16
Vcache replacement stat

tommydcjung:
Counting the number of replacing block
- invalid
- valid and not dirty
- valid and dirty


drichmond:
Does this still need review?

drichmond:
Rebase? Otherwise, looks good to me?

drichmond:
Actually, rebase, then let's figure out how we can get this data in #428. Maybe merge before #428

drichmond (approved):

PR #225, opened 2020-01-27
Check git version for parallel cloning of toolchain submodules

vb000:


vb000:
No, I tested it with (or ci tested it with) >2.8.

drichmond:
I think it'd be great to test with a lower version, but if you can't, it's not a big deal.

The original issue came from running on an AWS Instance with Centos 7

PR #226, opened 2020-02-04
Added extern c to bsg_manycore.h

drichmond:
Currently we can't call bsg_printf from c++ code. Adding this declaration solves the problem, BUT I'm not sure if it's too heavy handed. I included the whole file which is probably unnecessary.


drichmond:
@eafurst This may solve the problem you were seeing yesterday, too.

drichmond:
I switched it, just to avoid any unintentional issues.

PR #229, opened 2020-02-07
adds a striped array class

mrutt92:


software/bsg_manycore_lib/bsg_striped_array.hpp
```diff
@@ -0,0 +1,69 @@
+#pragma once
+extern "C" {
+#include "bsg_manycore.h"
+}
+
+namespace bsg_manycore {
+    template <typename TYPE, int SIZE, int TG_SIZE_X, int TG_SIZE_Y, int STRIPE_SIZE=1>
```
save-buffer: The template parameters should probably be `size_t` (except for TYPE). Also what is STRIPE_SIZE?

software/bsg_manycore_lib/bsg_striped_array.hpp
```diff
@@ -0,0 +1,69 @@
...
+        }
+
+        TYPE * _address(int i) const {
+            int stripe = _stripe(i);
+            int word_in_stripe  = _word_in_stripe(i);
+            int tile = _tile(stripe);
+            int stripe_in_tile = _stripe_in_tile(stripe);
+
+            TYPE *ptr =
+                (TYPE*) bsg_remote_ptr(bsg_id_to_x(tile), bsg_id_to_y(tile),
```
save-buffer: Use C++ casts!!!! This cast is casting away the volatile-ness of this pointer and will cause incorrect results! If you use `reinterpret_cast<TYPE *>()`, compiler will complain because you have to use `const_cast` to cast away volatile. You want `reinterpret_cast<volatile TYPE *>()`. Dereferencing later down the line will remove volatile on the value.

software/spmd/bsg_cuda_lite_runtime/striped/kernel_striped.cpp
```diff
@@ -0,0 +1,36 @@
...
+extern "C" int  __attribute__ ((noinline)) kernel_striped() {
+    Array data;
+    if (bsg_id == 0) {
+        for (int i = 0; i < data.size(); i++) {
+            data[i].bar = 2*i;
+            data[i].baz = 2*i+1;
+        }
+    }
+
+    bsg_tile_group_barrier(&r_barrier, &c_barrier);
```
save-buffer: This test correctly tests when tile with ID 0 writes to other tiles' memories. Due to the lack of `volatile` above, I believe this will fail down the line if you synchronize, have some other tiles write to their local addresses, and then you have tile 0 read again. It's likely that the value will be cached in a register and you won't see the updated value (I ran into this problem a while ago so that's how I know 😝).

software/bsg_manycore_lib/bsg_striped_array.hpp
```diff
@@ -0,0 +1,69 @@
...
+        TYPE operator[](int i) const {
+            return *_address(i);
+
+        }
+
+        TYPE & operator[](int i) {
+            return *_address(i);
+        }
+
+        TYPE at_local(int i) const {
```
save-buffer: What is `at_local` for? Seems like it takes away from the abstraction of having this be a simple array.

software/bsg_manycore_lib/bsg_striped_array.hpp
```diff
@@ -0,0 +1,69 @@
...
+
+        int size() const { return SIZE; }
+        int stripe_size() const { return STRIPE_SIZE; }
+
+        TYPE operator[](int i) const {
+            return *_address(i);
+
+        }
+
+        TYPE & operator[](int i) {
```
save-buffer: This reference should probably be marked `volatile` too.

software/bsg_manycore_lib/bsg_striped_array.hpp
mrutt92: It's to access the ith entry on the current tile. It seemed like something I would want so I put it there.

software/bsg_manycore_lib/bsg_striped_array.hpp
mrutt92: It's if you want more than one consecutive element on the current tile. And sure.

save-buffer: Oh that makes sense.

software/bsg_manycore_lib/bsg_striped_array.hpp
save-buffer: Fair enough

mrutt92:
>  I'd suggest just making this a wrapper around the existing macros (bsg_tile_group_shared_mem, bsg_tile_group_shared_load, and bsg_tile_group_shared_store).

I was going to do that but then I realized the syntax didn't really work if I wanted to overload the `[]` operator. `bsg_remote_ptr` does most of the heavy lifting anyway, so I leveraged that.

As for marking everything as volatile... I actually put some thought into that. The thing about the volatile keyword is it keeps the compiler from making nice optimizations (i.e. not reading something out of memory twice) and using volatile where it isn't necessary can degrade performance. 

If you think about it, just because data lives on a remote core does not inherently make it volatile. It's the possibility that other cores might change its value that makes it volatile, and that really depends on the application.

Ultimately, I decided that this class should be responsible for *where* data lives but not *how* that data is used. That decision should be left to the client programmer. If they want to mark this data as volatile, they can do that like so:

```c++
using Array = TileGroupSharedArray<volatile SomeType_t, ...>
```
Better yet, they can have a function like this:
```c++
SomeType_t load_volatile(volatile SomeType_t *ptr) { return *ptr; }
```
and be explicit in their program where they really want that load to happen.

Anyway, that's my two cents on the volatile keyword. Maybe providing the volatile load/store functions as part of the class is the way to go.

software/spmd/bsg_cuda_lite_runtime/striped/kernel_striped.cpp
drichmond: That's the point of a memory fence, as we were discussing on Thursday, right? After a fence, it'll go read the remote cores again. 

I think I'm with Max's comment - it's better to have the programmer specify the volatility, unless there's some fundamental reason.

software/bsg_manycore_lib/bsg_striped_array.hpp
drichmond: Is stripe size the right name for this? Size implies bytes, but nelems (or something) would imply that more than one element is used

Ditto for TG_SIZE_X/Y, and SIZE. Why not TG_DIM_X and TG_DIM_Y?

drichmond: Is stripe size the right name for this? Size implies bytes, but nelems (or something) would imply that more than one element is used

Ditto for TG_SIZE_X/Y, and SIZE. Why not TG_DIM_X and TG_DIM_Y?

drichmond:
@bornaehsani Do you think you can try using this in your shared mem matrix multiply?

software/spmd/bsg_cuda_lite_runtime/striped/kernel_striped.cpp
save-buffer: The compiler memory fence that we discussed on Thursday is to prevent the compiler from reordering reads/writes between sync points. I don't think it will necessarily make compiler decide to reread the value from remote tiles. 
I'd say the fundamental reason you want volatile by default is for correctness. It's a massive gotcha where you have to force a remote load. Your results will differ a lot between debug builds and release builds. 
I think that we need to make this shared array be easy to program with correctly. Someone who wants to optimize the use of the array should manually copy the values out into a local variable and then perform whatever they need to do. Otherwise this StripedArray class doesn't really conceptually represent a striped array - it represents some weird mess of having several independent copies of the same element (except it's undefined how many copies), which sometimes interact (except its' undefined when they interact), and which depend entirely on your compiler's register allocator.

mrutt92: > Otherwise this StripedArray class doesn't really conceptually represent a striped array - it represents some weird mess of having several independent copies of the same element (except it's undefined how many copies), which sometimes interact (except its' undefined when they interact), and which depend entirely on your compiler's register allocator.

I'm not sure I agree with this characterization. It doesn't make several independent copies of an array; it builds an array whose elements are interleaved across a group of tiles (a group of tiles is a "tile group," which is in the name of the class, and is a well defined thing in the context of our software stack). The template parameters allow the programmer to control how the elements are interleaved. 

Other than that, it has exactly the same semantics as a normal array and to be honest I'm pretty sure that's what we want by default. @save-buffer you're correct that it will often be necessary to mark this data as volatile, but suppose you just want more memory for an array than the 4KB on a single tile? Maybe there's not enough parallel work to justify actually using all the compute. In such a situation, you probably would not want the array to be treated as volatile all the time. 

> I think that we need to make this shared array be easy to program with correctly. Someone who wants to optimize the use of the array should manually copy the values out into a local variable and then perform whatever they need to do.

The thing is, it's way easier for a programmer to discover that they're getting the wrong answer than it is to find out that their performance is being degraded by some helper class that's marking everything as volatile by default. 

Further, if some poor programmer who knows what they're doing really does want to remove volatile, they'll have to copy and paste the original class and remove all the volatile keywords (or paste a bunch of calls to remove_volatile in their kernel). 

All that being said, catering to the common use case is a reasonable thing to do. I think the best way to achieve that is to make a wrapper that reuses this class but also adds volatile to the returned references. That way we get the best of both worlds.

software/bsg_manycore_lib/bsg_striped_array.hpp
mrutt92: I'm not really sure I agree that SIZE implies bytes in this case because that wouldn't really make any sense at all. It should be noted that many standard C++ containers use the size() function to mean the number of elements, not the number of bytes, in the data structure.

As for TG_SIZE_X/Y vs TG_DIM_X/Y - sure let's change the name.

mrutt92:
> Alright this seems good. Will we want to revisit Mark's idea down the line of using inline assembly to force loads for the volatile one instead of using the volatile keyword?

Yup - but that's a problem for a different PR.

taylor-bsg:
I think the main focus should be tracking down the right syntax for the
address exposed (?) barrier for BSP regions...

save-buffer:
What Mark was talking about for the barriers was this:
```
asm volatile("" ::: "memory");
```

There's actually a more standard way of doing this now in C and C++:
```
std::atomic_signal_fence(std::memory_order_acq_rel);
```

taylor-bsg:
Great! Is there a way to verify that this is doing what we want it to do?

save-buffer:
I believe this is a basic example that shows that it does force a write back to memory:
https://godbolt.org/z/wAXrS5

taylor-bsg:
Wow -- super cool tool. I tweaked the example a little to test a few other
cases, see below. Looks like it considers local variables to be cacheable
across the fence, which is what we want; hence why c survives across the
barrier.

void fenced()
{
int *c = A;
int B = *c;
std::atomic_signal_fence(std::memory_order_acq_rel);
*c = B+*c;
}

void unfenced()
{
int *c = A;
int B = *c;
*c = B+*c;
}

fenced():
lui a5,%hi(A)
ld a5,%lo(A)(a5)
lw a3,0(a5)
lw a4,0(a5)
addw a4,a4,a3
sw a4,0(a5)
ret

unfenced():
lui a5,%hi(A)
ld a4,%lo(A)(a5)
lw a5,0(a4)
slliw a5,a5,1
sw a5,0(a4)
ret
A:
.zero 8

mrutt92:
@drichmond this needs your approval to be merged.

drichmond:
Do you need me to squash it too?

mrutt92:
nope.

PR #231, opened 2020-02-13
Fix linker script to force sections to be word-aligned

save-buffer:
One of my kernels had a `.data.dram` section size of 97 bytes, which caused it to fail to load program onto manycore.


drichmond:
I'm sort of surprised this didn't happen earlier. 

@vb000 Interested in what you're working on

PR #232, opened 2020-02-14
adds cache afl, ainv, aflinv, and tagfl ops to the manycore network

mrutt92:


v/bsg_manycore_pkg.v
```diff
@@ -69,7 +70,14 @@ package bsg_manycore_pkg;
     ,e_amo_maxu
   } bsg_manycore_amo_type_e;
 
+  typedef enum logic [3:0] {
+    e_afl
+    ,e_ainv
+    ,e_aflinv
+  } bsg_manycore_cache_op_type_e;
```
tommydcjung: isn't TAGFL what you need?

mrutt92: I use tagla to scan the cache, so I know what address I need to flush/invalidate anyway.

tommydcjung: But that doesn't tell you whether the cache line is valid or not.

mrutt92: ...so? why does TAGFL solve that problem?

tommydcjung: with tagfl, you can just flush out every line, without having to query the tags. It saves time if there is no need to invalidate.

mrutt92: oh yeah - good point

mrutt92:
can this be approved?

tommydcjung:
I think this can be merged without breaking anything?

mrutt92:
ci seemed to fail over building llvm... which doesn't have anything to do with this PR - I'm mystified.

So I ran the SPMD regression manually:
I ran `make -j 16 recurse-all` from `bsg_manycore/software/spmd` - no universal log was generated.
So then I ran `find -name run.log | xargs grep FAIL` and I found this:
```
./beebs/nbody_run/run.log:[INFO][MONITOR] RECEIVED BSG_FAIL PACKET from tile y,x= 1, 0, data=00000000, time=50786732000
./beebs/st_run/run.log:[INFO][MONITOR] RECEIVED BSG_FAIL PACKET from tile y,x= 1, 0, data=00000000, time=5794705000
```
@vb000 can you tell me something about these tests?

tommydcjung:
nbody and st are expected to fail.

tommydcjung:
make sure to run coremark separately.

mrutt92:
sure - but why does Caremark need to be run separately?

mrutt92:
```
2K performance run parameters for coremark.
CoreMark Size    : 666
Total ticks      : 100
Total time (secs): 100
Iterations/Sec   : 0
Iterations       : 1
Compiler version : GCC9.2.0
Compiler flags   :   
Memory location  : STATIC
seedcrc          : 0xe9f5
[0]crclist       : 0xe714
[0]crcmatrix     : 0x1fd7
[0]crcstate      : 0x8e3a
[0]crcfinal      : 0xe714
Correct operation validated. See README.md for run and reporting rules.
```

mrutt92:
I think that means this is OK to merge?

tommydcjung:
Yup, I think it's good to go.

PR #237, opened 2020-02-27
Removed [INFO][VCORE] print statements

drichmond:
These print statements are constant in CUDA programs. They're annoying to work with because they can interfere with debug messages from the manycore and they're no longer useful (at least, until we need to debug kernel launching again)


drichmond:
Feedback, please. Approve if necessary

PR #238, opened 2020-03-03
branch spmd fix

tommydcjung:


drichmond:
Just curious, but why was this necessary?

tommydcjung:
after sending bsg_finish, pc goes into uninitialized address space, propagating x and crashing the simulation.

PR #239, opened 2020-03-03
launch only 1 tile for coremark

tommydcjung:


drichmond:
I wish we didn't have to specify tile group size in the makefiles, but that's not an issue for this PR

drichmond:
Otherwise, LGTM

PR #240, opened 2020-03-03
beebs exclude

tommydcjung:


tommydcjung:
Routines like sqrt checks fcsr for what the current rounding mode is. Currently that instruction is just treated as nop. So the routine uses whatever value is stored in the register and uses that as rounding mode.

drichmond:
Was st excluded for the same reason?

drichmond:
PR looks good to me - but as a general question, I recall this fcsr issue being discussed before. What's our long-term solution?

tommydcjung:
We confirmed long ago that 'st' came with a bug.
When we switch to hardfloat, we will fully support fcsr.

drichmond:
When we switch to hardfloat, with the FPU still be pipelined? (+@save-buffer so that he sees the answer to this)

drichmond:
Can you add a comment about st having a bug, similar to what you did for nbody?

tommydcjung:
> When we switch to hardfloat, with the FPU still be pipelined? (+@save-buffer so that he sees the answer to this)

yes

drichmond:
How many cycles in the hardfloat FPU pipeline? Do we know yet?

tommydcjung:
It can be 2 or 3, depending on timing closure.

drichmond:
So it's parameterized? 

It would be useful to set a value and stick to it so that we can put an accurate upper-bound in the gcc cost maps that @save-buffer is developing. Or maybe we can have two different maps. 

Anyway, not relevant at the moment. Thanks for the info!

save-buffer:
Having two different cost maps would not be an issue at all to implement - just copy-paste one thing and change the numbers around

drichmond:
That's good to know, but I'm worried about long term proliferation of compile options as well. I suppose this won't be answered until much later, so I'd say we should assume worst case (3 cycles) and get a magical performance improvement later on

You'll also need to know what hazard forwarding the new FPU provides, if any.

taylor-bsg:
Is kind of interesting idea to use ML to change the latencies that the
compiler assumes for each piece of code and then see if you get better
results.

M

PR #241, opened 2020-03-03
remove defunct spmd test

tommydcjung:
these spmds are remove because
- deprecated, no longer works.
- there is a better test for this.


taylor-bsg:
Maybe you can be more specific about what tests are replacing these tests?

tommydcjung:
These mostly relate to dram access, vcache, and float. We already have an extensive list of tests to cover those cases.

drichmond:
I agree with @taylor-bsg - more specificity would be helpful. Otherwise, assuming we have replacements for these tests that test the same functionality, this PR looks good to me!

tommydcjung:
float: float_asm_test, float_div, float_math, float_vector
cache, dram: memtest1, memtest9, vcache_bang

PR #248, opened 2020-03-04
move some files to sim_filelist

tommydcjung:


machines/arch_filelist.mk
```diff
@@ -114,11 +109,6 @@ VSOURCES += $(BASEJUMP_STL_DIR)/bsg_cache/bsg_cache_dma.v
 VSOURCES += $(BASEJUMP_STL_DIR)/bsg_cache/bsg_cache_miss.v
 VSOURCES += $(BASEJUMP_STL_DIR)/bsg_cache/bsg_cache_sbuf.v
 VSOURCES += $(BASEJUMP_STL_DIR)/bsg_cache/bsg_cache_sbuf_queue.v
-VSOURCES += $(BASEJUMP_STL_DIR)/bsg_cache/bsg_cache_to_axi_rx.v
-VSOURCES += $(BASEJUMP_STL_DIR)/bsg_cache/bsg_cache_to_axi_tx.v
```
drichmond: Aren't these files needed for F1 implementation?

tommydcjung: this is instantiated in the testbench, and it's not the core architectural piece. So I think it's appropriate to move there.

drichmond: For manycore, yes. But my question still stands: Aren't these files needed for F1 implementation?

tommydcjung: yes

drichmond: OK, I agree that arch_filelist.mk is not the place for these files in bsg_manycore. 

This PR will break the F1 build and the cosimulation flow, because we don't use sim_filelist.mk.

drichmond: I'm not sure whose responsibility it is, but I'll file a separate PR in Replicant for adding these files back

tommydcjung: F1 has its own list of files for synthesis for things like axil-to-mcl. maybe you should just add it there?

drichmond: That's the plan!

drichmond: I don't need bsg_cache_to_dram_ctrl, right? That's just for manycore?

drichmond: (With updates, this passes in F1, so just waiting for an answer to the previous question)

tommydcjung: That's only for manycore

drichmond: Thanks!

PR #249, opened 2020-03-04
make vcache_atomic_histogram more flexible

tommydcjung:


drichmond:
Approved, because I think I understand what's going on. @mrutt92 should add his two cents

PR #255, opened 2020-03-06
Added safeguard for vanilla stats parser against empty stats

bornaehsani:
Asserts to make sure there are traces in the vanilla_stats_parser, prints out error if it is empty.
Minor color updates for bloodgraph.


bornaehsani:
feeling lazy :D

PR #260, opened 2020-03-10
support AMO on infinite mem

tommydcjung:


machines/4x4/Makefile.machine.include
```diff
@@ -14,9 +14,8 @@ BSG_MACHINE_DRAM_INCLUDED             = 1
 BSG_MACHINE_MAX_EPA_WIDTH             = 28
 
 # supported memory cfg
-# e_infinite_mem
```
mrutt92: I feel like makefiles in this directory shouldn't switch between memory configurations like this. maybe have a subdirectory for each config?

testbenches/common/v/bsg_nonsynth_mem_infinite.v
```diff
@@ -97,46 +87,178 @@ module bsg_nonsynth_mem_infinite
...
+    mem_mask_li = '0;
+    
+    return_packet_li = '0;
+    return_packet_v_li  = 1'b0;
+
+    state_n = state_r;
+    packet_n = packet_r;
+    return_v_n = return_v_r;
+
+    case (state_r)
```
mrutt92: This all looks ok but adding comments where you have nested ternary statements would improve readability.

PR #266, opened 2020-03-17
Fixed origin location for top and bottom memory systems

drichmond:
The python scripts broke when the origin moved

### DIFF
--- a/software/py/vanilla_pc_histogram.py
+++ b/software/py/vanilla_pc_histogram.py
@@ -33,7 +33,7 @@ class PCHistogram:
 
     # Default coordinates of origin tile
     _BSG_ORIGIN_X = 0
-    _BSG_ORIGIN_Y = 1
+    _BSG_ORIGIN_Y = 2
 
     _BSG_PC_MIN = 0x0000
     _BSG_PC_MAX = 0x2000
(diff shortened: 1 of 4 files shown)


mrutt92:
anyway we can inherit those values from the machine makefile instead of hardcoding them into the scripts?

drichmond:
We could parse them as input arguments

drichmond:
OK, fixed. Automatically infer origin from the active tiles

drichmond:
Added comments, and tested on single-tile and full-device CUDA programs.

bornaehsani:
Looks good to me, thanks Dustin! I think this is the best way we can deal with the issue of origin right now, but it banks on the fact that the API starts allocating tiles from the upper left most corner, so we should beware if that ever changes.

drichmond:
Yeah, I'm not a fan of that assumption either. We could put a check / warning in but I haven't thought of what that check/warning would be. Basically if the origin isn't (0,0), (1,0), (0,1) or (1,1)? Or more restrictive - (0, 1) as the way things currently are now. 

But that's effectively hard-coding the origin, which is what this PR is addressing.

drichmond:
This is really only a problem in the stats parser, right?

bornaehsani:
Yeah this is only a problem for the parsers, stats parser, stall graph (maybe) and the pc histogram. I think this will suffice to solve the problem for now, and in the future it's easy to detect if the error pops up again.

drichmond:
See #270 (for CI)

PR #269, opened 2020-03-18
Infmem dma

mrutt92:
* adds backdoor data store to infinite memory module
* infinite_mem_profiler takes a parameter for its log file name

### DIFF
--- a/testbenches/common/v/spmd_testbench.v
+++ b/testbenches/common/v/spmd_testbench.v
@@ -217,6 +217,7 @@ module spmd_testbench;
           ,.addr_width_p(bsg_max_epa_width_p)
           ,.x_cord_width_p(x_cord_width_lp)
           ,.y_cord_width_p(y_cord_width_lp)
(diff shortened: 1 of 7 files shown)


testbenches/common/v/spmd_testbench.v
```diff
@@ -217,6 +217,7 @@ module spmd_testbench;
           ,.addr_width_p(bsg_max_epa_width_p)
           ,.x_cord_width_p(x_cord_width_lp)
           ,.y_cord_width_p(y_cord_width_lp)
+          ,.id_p(num_tiles_x_p*j + i)
```
tommydcjung: does it matter if j is not 0 and 1? I think N,S is 3 and 4?

mrutt92: it doesn't matter - the id is just a key into an std::map

mrutt92:
as of two seconds ago this looks good to go.

drichmond:
Woohoo! merge?

PR #273, opened 2020-03-27
Py Link Gen

vb000:
- Python script to generate link script with different configs
- Simplified coremark -- do away with perpetual simlinks.


software/py/bsg_manycore_link_gen.py
```diff
@@ -0,0 +1,252 @@
+#!/usr/bin/python3
+
+import argparse, re
```
drichmond: ❤️ 

(I opened this PR and wondered if you were using argparse)

software/py/bsg_manycore_link_gen.py
```diff
@@ -0,0 +1,252 @@
...
+                              '.gnu.linkonce.s.*']],
+      ['.sbss'             , ['.sbss','.sbss.*','.gnu.linkonce.sb.*','.scommon']],
+      ['.bss'              , ['.bss','.bss*']],
+      ['.tdata'            , ['.tdata','.tdata*']],
+      ['.tbss'             , ['.tbss','.tbss*']],
+      ['.eh_frame'         , ['.eh_frame','.eh_frame*']],
+      ['.striped.data.dmem', ['.striped.data']],
+      ['.rodata.dram'      , ['.rodata','.rodata*']],
+      ['.dram'             , ['.dram','.dram.*', '.*']],
+      ]
```
drichmond: Do you want this to be a list of lists, or a dictionary with list values? (My sense is the former, because ordering seems to matter)

software/py/bsg_manycore_link_gen.py
```diff
@@ -0,0 +1,252 @@
...
+
+    mem_regions = [
+      # Format:
+      # [<mem region name>, <access>, <start address>, <size>]
+      #
+      # .text section VMAs and dmem memory VMAs overlap with the
+      # current memory layout (both start at address 0x0). This makes
+      # manycore essentially a Harvard architecture. But linker doesn't
+      # allow overlapping VMAs by not letting location couter not move 
+      # backwards. To overcome this, we use combination of VMA and LMA
```
drichmond: Typo, and double-negative. The double-negative is confusing

software/py/bsg_manycore_link_gen.py
```diff
@@ -0,0 +1,252 @@
...
+    script += "\n}} {0} {1}\n\n".format(vma_redirect, lma_redirect)
+
+    return script
+
+  def script(self):
+    # Address constants:
+    #
+    # LMA (Load Memory Address)    => NPA used by loader
+    # VMA (Virtual Memory Address) => Logical address used by linker for 
+    #                                 symbol resolutions
```
drichmond: I like the """ comments for blocks better than the #

software/mk/Makefile.builddefs
```diff
@@ -123,24 +124,29 @@ CRT_OBJ ?= crt.o
...
   else
     $(error Invalid BSG_ELF_DEFAULT_DATA_LOC = $(BSG_ELF_DEFAULT_DATA_LOC); Only LOCAL and SHARED are valid)
   endif
 else ifeq ($(BSG_ELF_OFF_CHIP_MEM), 0)
   ifeq ($(BSG_ELF_DEFAULT_DATA_LOC), LOCAL)
-    LINK_SCRIPT ?= $(BSG_MANYCORE_DIR)/software/spmd/common/link_dmem2.ld
+    LINK_GEN_OPTS ?= --default_data_loc=dmem --dram_size=$(BSG_ELF_VCACHE_SIZE) --sp=$(BSG_ELF_STACK_PTR)
   else ifeq ($(BSG_ELF_DEFAULT_DATA_LOC), SHARED)
-    LINK_SCRIPT ?= $(BSG_MANYCORE_DIR)/software/spmd/common/link_dram2.ld
+    LINK_GEN_OPTS ?= --default_data_loc=dram --dram_size=$(BSG_ELF_VCACHE_SIZE) --sp=$(BSG_ELF_STACK_PTR)
```
drichmond: A general question for discussion: 

Effectively, these options are determined by the machine file. Should we put them there, so that users can more easily change them?

software/mk/Makefile.builddefs
```diff
@@ -123,24 +124,29 @@ CRT_OBJ ?= crt.o
...
   else
     $(error Invalid BSG_ELF_DEFAULT_DATA_LOC = $(BSG_ELF_DEFAULT_DATA_LOC); Only LOCAL and SHARED are valid)
   endif
 else
   $(error Invalid BSG_ELF_OFF_CHIP_MEM = $(BSG_ELF_OFF_CHIP_MEM); Only 0 and 1 are valid)
 endif
 
+LINK_SCRIPT ?= bsg_link.ld
+
+bsg_link.ld: $(LINK_GEN)
```
drichmond: I'd re-write this as `$(LINK_SCRIPT): $(LINK_GEN)` (and maybe depend on BSG_MACHINE_PATH/Makefile.machine.include?)

software/py/bsg_manycore_link_gen.py
```diff
@@ -0,0 +1,252 @@
...
+  include default location of data (dmem or dram), stack pointer and
+  dram memory size. These options won't affect the following:
+
+  1. Data attributed to .dmem* and .dram* sections. Following decalaration
+
+     int foo __attribute__((section (".dram")));
+
+     would place foo in dram region irrespective of default data location.
+
+  2. Data in bsg_manycore_lib would always be placed in dmem region.
```
drichmond: "Will always be placed"

software/py/bsg_manycore_link_gen.py
```diff
@@ -0,0 +1,252 @@
...
+
+  
+  Note on program memory:
+  
+  Manycore's PC width is 24-bit, addressing a total of 16MB program space.
+  Since the 31:25 MBSs are assumed to be 0s, linking should also assume
```
drichmond: Do you mean MSBs?

software/py/bsg_manycore_link_gen.py
```diff
@@ -0,0 +1,252 @@
...
+  0x0-0x01000000 as the valid program space. This means that .text section VMAs
+  lie in 0x0-0x01000000. On the other hand, .text section is loaded to DRAM and
+  as result, it's LMAs shouls be >0x80000000. So, .text section is loaded in the
```
drichmond: "shouls" -> "should"

software/py/bsg_manycore_link_gen.py
```diff
@@ -0,0 +1,252 @@
...
+    help = 'Stack pointer',
+    default = 0x1000,
+    type = lambda x: int(x, 0))
+  args = parser.parse_args()
+
+
+  # Generate linker script
+  link_gen = bsg_manycore_link_gen(args.default_data_loc, args.dram_size,
+      args.sp)
+  print(link_gen.script())
```
drichmond: I think it would be useful to be able to specify the output file as an argument. If it isn't provided, you can print it to stdout.

software/py/nbf.py
```diff
@@ -99,7 +99,7 @@ def read_objcopy(self, section, output_file):
       print("install riscv-tools first...")
       sys.exit()
 
-    cmd = [objcopy_path, "-O", "verilog", "-j", section, self.riscv_file, output_file]
+    cmd = [objcopy_path, "-O", "verilog", "-j", section, "--set-section-flags", "*bss*=alloc,load,contents", self.riscv_file, output_file]
```
drichmond: What's the purpose of this? Is it supposed to be part of this PR?

software/py/bsg_manycore_link_gen.py
vb000: Yep, natural choice are dicts, but realized order matters here. For eg., `.crtbegin` section should always be at the top...

software/py/bsg_manycore_link_gen.py
```diff
@@ -0,0 +1,252 @@
...
+class bsg_manycore_link_gen:
+  """
+  A configurable linker command file generator for linking BSG Manycore
+  kernels. Kernel memory has private and shared regions backed by 4KB
+  on-tile sram (DMEM) and off-chip DRAM, respectively. Configuration options
+  include default location of data (dmem or dram), stack pointer and
+  dram memory size. These options won't affect the following:
+
+  1. Data attributed to .dmem* and .dram* sections. Following decalaration
+
```
drichmond: declaration -> declaration

software/py/bsg_manycore_link_gen.py
```diff
@@ -0,0 +1,252 @@
...
+    "/**********************************\n" \
+    + " BSG Manycore Linker Script \n\n"
+
+  def __init__(self, default_data_loc, dram_size, sp):
+    self._default_data_loc = default_data_loc
+    self._dram_size = dram_size
+    self._sp = sp
+    self._opening_comment += \
+        " data default: {0}\n".format(default_data_loc) \
+        + " dram memory size: 0x{0:08x}\n".format(dram_size) \
```
drichmond: Is there more information that we can put in this opening comment?

drichmond:
I really like this PR. Just reviewing it thoroughly to make it even better and make sure others understand why it's such a good PR.

drichmond:
Should the CoreMark modifications be in this PR? I try to promote 1 feature-per-PR because it's good style, makes it easier to review, and makes it easier to debug/revert (if that ever becomes necessary). 

If they're inseparable, then that should be explained

software/py/bsg_manycore_link_gen.py
mrutt92: I think docstrings are generally better because they can be accessed with help() and I think there are tools to make documentation pages from them.

software/py/nbf.py
vb000: Yep, very much! This is one of the hardest debugs I've done. Apparently, objcopy doesn't output contents of bss sections and as result, tests were failing because of some failure in the C library. Ideally, elf loader should be able to leave space for bss by looking at load addresses, but the spmd_loader doesn't do that. So, this flag is required.

software/py/bsg_manycore_link_gen.py
drichmond: Yes, that's why I like them. 

But you generally have to be pretty pedantic with formats.

software/py/nbf.py
drichmond: Well, glad you debugged it :)

But is it caused by this PR, or is it a separate bug?

software/mk/Makefile.builddefs
vb000: Nope... we shouldn't generate if user provides a path for custom link script. May be I should put that in a comment.

software/py/bsg_manycore_link_gen.py
drichmond: For example, are all the options captured here? Like we do when running simulation, it might be useful to put as much information here as possible so that we can reproduce bugs/errors in a tool

software/mk/Makefile.builddefs
vb000: Some of then differ program to program and some of then are taken from machine file anyway...

software/mk/Makefile.builddefs
drichmond: Hmm. Maybe there's a better way to specify it. What exactly are you trying to accomplish? Provide the ability to generate a link script in the directory that this makefile is included from?

vb000:
> Should the CoreMark modifications be in this PR? I try to promote 1 feature-per-PR because it's good style, makes it easier to review, and makes it easier to debug/revert (if that ever becomes necessary).
> 
> If they're inseparable, then that should be explained

Sometimes the overhead is a lot, 2 PRs = 2 CIs. Coremark refactor is a refactor that's fully backward compatible, not a feature... so I think CI provides a decent preliminary review for you ;).

drichmond:
> > Should the CoreMark modifications be in this PR? I try to promote 1 feature-per-PR because it's good style, makes it easier to review, and makes it easier to debug/revert (if that ever becomes necessary).
> > If they're inseparable, then that should be explained
> 
> Sometimes the overhead is a lot, 2 PRs = 2 CIs. Coremark refactor is a refactor that's fully backward compatible, not a feature... so I think CI provides a decent preliminary review for you ;).

Yes, but it's good form and it helps Max and I as reviewers. Are you telling me that these PRs are inseparable?

Remember, we went through this in another PR recently. It may be a lot of overhead, but it is worth it.

vb000:
Just a thought... I strongly support voluntary efforts to refactor code which may not be directly related to pr. Requiring those efforts to make a separate branch, ci and spend a ton of time just discourages people from touching old code. Open to your comments...

software/py/bsg_manycore_link_gen.py
```diff
@@ -0,0 +1,252 @@
...
+    for m, access, origin, size in mem_regions:
+      script += "{0} ({1}) : ORIGIN = 0x{2:08x}, LENGTH = 0x{3:08x}\n".format(
+          m, access, origin, size)
+
+    script += '}\n'
+
+    return script
+
+  def _section(self, name, address, vma_region, lma_region,
+      in_sections, in_objects):
```
drichmond: You can take this feedback as a general FYI ("Hey this is another coding style") or you can address it if you feel like it.

From my reading it seems like the _section and _memory_region methods are prime candidates to be split into objects. For example, each section has attributes, specific arguments, and a string representation. Furthermore, there's specific checking that can be done for each section. Therefore, I conceptually think of each section as it's own Object (even though

For a simple python script (like this) it's not necessary, but in large programs it can break it up, modularize, and create abstractions. Take it as you will

vb000:
> Yes, but it's good form and it helps Max and I as reviewers. Are you telling me that these PRs are inseparable?
> 
> Remember, we went through this in another PR recently. It may be a lot of overhead, but it is worth it.

I totally understand separating into 2 prs makes it easy for everyone to follow... but all I'm proposing is to consider that case by case rather than a hard rule. If Coremark refactor is hard to understand in this pr, I'm very much open to making a new pr...

vb000:
Dustin, also this pr for some reason (that I couldn't debug), breaks coremark makefile. Instead of doing a quick fix, I chose to do this refactor. Moving this to a separate pr means making the quick fix and a refactor on top it which makes the quick fix obsolete.

drichmond:
> Dustin, also this pr for some reason (that I couldn't debug), breaks coremark makefile. Instead of doing a quick fix, I chose to do this refactor. Moving this to a separate pr means making the quick fix and a refactor on top it which makes the quick fix obsolete.

This is exactly what I was trying to understand. If that's the case, then yes, these are inseparable.

drichmond:
> > Yes, but it's good form and it helps Max and I as reviewers. Are you telling me that these PRs are inseparable?
> > Remember, we went through this in another PR recently. It may be a lot of overhead, but it is worth it.
> 
> I totally understand separating into 2 prs makes it easy for everyone to follow... but all I'm proposing is to consider that case by case rather than a hard rule. If Coremark refactor is hard to understand in this pr, I'm very much open to making a new pr...

> Just a thought... I strongly support voluntary efforts to refactor code which may not be directly related to pr. Requiring those efforts to make a separate branch, ci and spend a ton of time just discourages people from touching old code. Open to your comments...

Going to address both questions here because I believe they're related (if they're not, correct me). I'm going to try to outline my general feelings about the matter in hopes that my position becomes clearer. Please don't take this as a hard and fast rule, instead, take it as when I ask someone to split a PR there's usually a reason.

Also happy to hop on zoom and discuss this.

I understand your point, and it is a good one. I try to strike a balance because I don't want to slow down productivity. Yes, it does discourage people from touching old code and that is an issue that has come up in the past in this repository (Not pointing fingers, this always happens in large projects).

In major features (which I consider this) I don't like to see multiple **unrelated** features addressed because I want to focus on the major feature that is being addressed. This is a great PR, but it requires 120 (ish) file changes because every makefile in software/spmd needs to change. If you added another unrelated change it would be hard for me, as a reviewer, to separate and give feedback on that many files.

I'm perfectly happy with multiple-unrelated-feature PRs if they have a bunch of simple changes. For example, if you changed the clean rule in makefile.builddefs and something else that is simple (I'm sure that you can come up with examples). Ideally those changes would modify a small number of files so that it's easy to see how the changes do or don't interact. Tell the reviewer about them, too.

But it seems like all of this is a moot point. Something broke in CoreMark with this PR. You fixed it and it seems better than the original. Regardless, they're inseparable. Let's get this reviewed and merged.

(One great aspect of this PR is that the CoreMark changes are simple, AND they're confined to one commit. This makes it easier to review in situations where there is a major feature and a few smaller bug-fixes or sub-features, like this.)

You did right here (though it took some back and forth to get to that point). 

There are no hard and fast rules here, so I can't give you a straight answer. But it's the responsibility of the requester to help the reviewer, whether it be comments, splitting up PRs, squashing, or literally sitting down and walking them through the review. On the flip side, I take my responsibility as a reviewer seriously, and I will try to provide an explanation for my comments, concerns, and requests. 

Does that help clarify my thought process?

software/py/bsg_manycore_link_gen.py
```diff
@@ -0,0 +1,252 @@
...
+    sections += "_sp = 0x{0:08x};\n".format(self._sp)
+    sections += "_bsg_data_start_addr = 0x{0:08x};\n".format(_DMEM_VMA_START)
+    sections += "_bsg_data_end_addr = ADDR(.striped.data.dmem) + " \
+                "SIZEOF(.striped.data.dmem);\n"
+    sections += "_bsg_striped_data_start = ADDR(.striped.data.dmem)\n;"
+    sections += "_bsg_dram_t_start_addr = LOADADDR(.text.dram)\n;"
+    sections += "_bsg_dram_d_start_addr = LOADADDR(.text.dram) + SIZEOF(.text.dram)\n;"
+    sections += "_bsg_dram_d_end_addr = LOADADDR(.dram) + SIZEOF(.dram)\n;"
+    sections += "_bsg_dram_end_addr = _bsg_dram_d_end_addr\n;"
+    sections += "_end = _bsg_dram_end_addr\n;"
```
drichmond: Is the function of these variables defined in a google doc somewhere?

software/mk/Makefile.builddefs
vb000: This target says, only generate `bsg_link.ld` and don't generate any other link script...

drichmond:
Can we get rid of legacy linker definitions? Or would that break a part of the codebase?

drichmond:
I also don't see any striping variables defined in the new linker generator. This is fine, but I want to confirm, are we OK with breaking the LLVM Pass? (@proftaylor)

drichmond:
When I run test_symbol_to_eva in replicant I get the following error:

```
BSG INFO: test_symbol_to_eva Regression Test (COSIMULATION)
BSG INFO: test_symbol_to_eva: test        lookup-data: ...
BSG INFO: test_symbol_to_eva: test        lookup-data: PASSED
BSG INFO: test_symbol_to_eva: test        lookup-dram: ...
BSG INFO: test_symbol_to_eva: test        lookup-dram: symbol '_bsg_dram_d_start_addr': FAILED: expected 0x81000000, got 0x800002d0
BSG INFO: test_symbol_to_eva: test         find-start: ...
BSG INFO: test_symbol_to_eva: test         find-start: PASSED
BSG INFO: test_symbol_to_eva: test         bad-buffer: ...
BSG INFO: test_symbol_to_eva: test         bad-buffer: PASSED
BSG INFO: test_symbol_to_eva: test   symbol-not-found: ...
BSG INFO: test_symbol_to_eva: test   symbol-not-found: PASSED
BSG INFO: test_symbol_to_eva: test           bad-size: ...
BSG INFO: test_symbol_to_eva: test           bad-size: PASSED
BSG INFO: test_symbol_to_eva: test        badder-size: ...
BSG INFO: test_symbol_to_eva: test        badder-size: PASSED
BSG INFO: test_symbol_to_eva: test         bad-symbol: ...
BSG INFO: test_symbol_to_eva: test         bad-symbol: PASSED
BSG INFO: test_symbol_to_eva: test            bad-eva: ...
BSG INFO: test_symbol_to_eva: test            bad-eva: PASSED
BSG REGRESSION TEST FAILED
BSG COSIM FAIL: Test failed with exit code:  
```

vb000:
> BSG INFO: test_symbol_to_eva Regression Test (COSIMULATION)
> BSG INFO: test_symbol_to_eva: test        lookup-data: ...
> BSG INFO: test_symbol_to_eva: test        lookup-data: PASSED
> BSG INFO: test_symbol_to_eva: test        lookup-dram: ...
> BSG INFO: test_symbol_to_eva: test        lookup-dram: symbol '_bsg_dram_d_start_addr': FAILED: expected 0x81000000, got 0x800002d0

DRAM data used to start at `0x81000000` and `0x80000000-0x80ffffff` was reserved for I-EVA backing storage. That need not be case and data can start right after instructions end (this pr does this as well).

drichmond:
Ahhhhh

OK, this brings in a previously unrelated question. Let's say that we want to have a heterogeneous Manycore. Program execution initialization and loading may occur after data is allocated for the accelerator, in which case, the instruction segment may be allocated over. 

Does it make sense to reserve the first 18 MB for instructions?

vb000:
> Ahhhhh
> 
> OK, this brings in a previously unrelated question. Let's say that we want to have a heterogeneous Manycore. Program execution initialization and loading may occur after data is allocated for the accelerator, in which case, the instruction segment may be allocated over.

Thought we'd follow a programming model in which we load program segment once at the beginning of the execution.

> 
> Does it make sense to reserve the first 18 MB for instructions?

You mean 16MB?

drichmond:
Yes, my mistake, 16 MB.

drichmond:
Yes, we defaulted to telling people to allocate the CUDA program first if they want to run CUDA and a custom accelerator, which basically is that situation. 

However, for the sake of discussion, let's assume someone has a heterogeneous machine, they might allocate data for an accelerator first and then initialize the CUDA program. We can't change the starting PC value, and we'd like to prevent the allocator from allocating that space. If we want this route, we'd need to reserve a portion of the memory space for instructions.

Hence my question

vb000:
> Yes, we defaulted to telling people to allocate the CUDA program first if they want to run CUDA and a custom accelerator, which basically is that situation.
> 
> However, for the sake of discussion, let's assume someone has a heterogeneous machine, they might allocate data for an accelerator first and then initialize the CUDA program. We can't change the starting PC value, and we'd like to prevent the allocator from allocating that space. If we want this route, we'd need to reserve a portion of the memory space for instructions.
> 
> Hence my question

Sounds good to have the flexibility. All that costs is about 10MB of DRAM. I'd revert back to previous setting.

vb000:
Or this aspect could be a configuration option as well! Power of Python enables this flexibility!

drichmond:
Ooh, there's an interesting idea.

software/py/bsg_manycore_link_gen.py
vb000: Fixed.

software/py/bsg_manycore_link_gen.py
vb000: This is not """ because it's not the description of the method, it's just the description of statements below.

Anyway, I realized a description to this method makes sense explaining the VMA and LMA regions. So added it.

software/mk/Makefile.builddefs
vb000: The script doesn't have any global parameters. Everything configurable, is provided by command line arguments.

software/mk/Makefile.builddefs
vb000: Any suggestion to do this in a better way?

software/py/bsg_manycore_link_gen.py
vb000: Done.

software/py/bsg_manycore_link_gen.py
vb000: Done.

software/py/bsg_manycore_link_gen.py
vb000: Done.

software/py/nbf.py
vb000: We can say, this bug got surfaced in this pr.

software/py/bsg_manycore_link_gen.py
vb000: Done.

software/py/bsg_manycore_link_gen.py
vb000: Done.

software/py/bsg_manycore_link_gen.py
vb000: Missed stack pointer initialization... added it. That should cover all three configuration options. Also added time stamp.

software/py/bsg_manycore_link_gen.py
vb000: I had thought about this quite a bit when writing this and felt that _section and _memory_region doesn't need any state. So making them plain function that gives one output for each input seemed right.

Also, my primary consideration with this script was if someone want change the script, they most probably should only have to touch `.script()` method. So, `.script(..)` method should feel more or less like writing actual `.ld` script with the additional power of Python. So didn't want to deviate from `.ld` semantics by too much "class"ification ;).

software/py/bsg_manycore_link_gen.py
vb000: They are just addresses of boundaries of the memory layout. Start of dmem, end of dmem, start of dram etc. I think verilog loader needs these symbols. May be @tommydcjung can confirm if we still need them...?

vb000:
> Can we get rid of legacy linker definitions? Or would that break a part of the codebase?

I'll create a separate pr removing those after this passes ci on post merging to master.

vb000:
- Added `--imem-size` option to configure allocated imem size!
- Added asserts to check dmem and imem overflows.

software/py/bsg_manycore_link_gen.py
drichmond: Are they part of the LLVM Pass?

software/mk/Makefile.tail_rules
```diff
@@ -34,7 +34,7 @@ spike_call_log: pp_spike_trace
 CLEAN_ITEMS ?=
 
 clean:
-	-rm -rf *.o *.a *.jou *.log *.pelog *.pb bsg_manycore_io_complex_rom.v *.riscv *.wdb *.bin *.hex
+	-rm -rf *.o *.a *.jou *.log *.pelog *.pb bsg_manycore_io_complex_rom.v *.riscv *.wdb *.bin *.hex *.ld
```
drichmond, line 37: If .ld files are now removed by clean, I think you should add .ld files to .gitignore

drichmond: Or at least software/spmd/.gitignore

software/py/bsg_manycore_link_gen.py
drichmond: Cool, thanks!

software/mk/Makefile.tail_rules
vb000: Sounds good!

software/py/bsg_manycore_link_gen.py
```diff
@@ -0,0 +1,295 @@
...
+      ['.striped.data.dmem', ['.striped.data']],
+      ['.rodata.dram'      , ['.rodata','.rodata*']],
+      ['.dram'             , ['.dram','.dram.*', '.*']],
+      ]
+
+    sections = "SECTIONS {\n\n"
+
+    # DMEM sections
+    for i, m in enumerate(section_map):
+      sec = m[0]
```
drichmond, line 176: I'm trying to understand what is happening the two for loops here. It appears that the code is iterating through the section map and finding strings that only contain .dmem (hence the $). After that, I'm a little confused.

Can you help me out? I'm happy to hop on a zoom and have you walk me through it. 


Likewise, for the lower for loop with DRAM.

vb000: Cool, I'll ping you!

software/spmd/bsg_barrier/Makefile
```diff
@@ -8,7 +8,7 @@ include ../Makefile.include
 
 all: main.run
 
-main.riscv:  $(OBJECT_FILES) $(SPMD_COMMON_OBJECTS) $(BSG_MANYCORE_LIB) ../common/crt.o
+main.riscv: $(LINK_SCRIPT)  $(OBJECT_FILES) $(SPMD_COMMON_OBJECTS) $(BSG_MANYCORE_LIB) ../common/crt.o
 	$(RISCV_LINK) $(OBJECT_FILES) $(SPMD_COMMON_OBJECTS) -L. "-l:$(BSG_MANYCORE_LIB)" -o $@ $(RISCV_LINK_OPTS)
```
drichmond, line 12: Not for this PR, but a thought:

Do MOST of the SPMD programs have the same compilation rule? The answer here is obviously no, because software/spmd/branch/Makefile has a different rule. However, many look similar. Is there an important reason why they differ, when they differ? Could we move the main.riscv compilation rule so we don't have to edit 100-ish files every time we do a change like this? Perhaps to tail_rules?

Just a thought

drichmond: For example, we  have discussed the benefit of compiling compiling crt.o in each directory. instead of using a common path. This is another PR that would modify 100 ish files in very similar ways

vb000:
Added descriptions for `if-else`s

software/py/bsg_manycore_link_gen.py
vb000: Not sure, but added descriptions for each symbol does.

software/spmd/bsg_barrier/Makefile
vb000: Oh missed this! May be we could factor out common prerequisites and common argument `$(RISCV_LINK)` command?

In this particular case though, it was a minor bug to not have `$(LINK_SCRIPT)` one of the prerequisites, I think.

vb000:
Any more required changes?

PR #274, opened 2020-03-27
Add a rule to build link script on a wildcard path

drichmond:
This adds a rule that allows replicant to build link scripts on any path. This allows replicant to do two things: 

* build machine-specific link scripts
* Auto-re-build the RISC-V binary when the machine changes. 

I haven't decided if this is the right way to do this. But it is a way to do things.


drichmond:
Closing, no longer relevant

PR #275, opened 2020-03-31
Adds Blocking Vcache stall graph

bornaehsani:
- Generates blocking vcache trace file as vcache_operation_trace.csv
- Generates vcache stall graph as vcache_stall_abstract/detailed.png
- At the moment abstract and detailed versions are identical, will change in the future
- TODO: need @drichmond 's makefile magic to append test_name and call vcache_stall_graph.py on the generated trace.

node: at the moment, the vcache_operation_trace.csv file is generated inside `.<test_name>/vcache_operation_trace.csv` , and vcache_stall_graph.py should be called manually. 

**merge with PR # 543 on bsg_replicant.**

### DIFF
--- a/testbenches/common/v/vcache_profiler.v
+++ b/testbenches/common/v/vcache_profiler.v
@@ -30,6 +30,8 @@ module vcache_profiler
     , input [31:0] global_ctr_i
     , input print_stat_v_i
     , input [data_width_p-1:0] print_stat_tag_i
+
+    , input trace_en_i // from toplevel testbench
   );
 
   // task to print a line of operation trace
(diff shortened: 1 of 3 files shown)


testbenches/common/v/vcache_profiler.v
```diff
@@ -40,10 +48,27 @@ module vcache_profiler
...
   //
 
-  wire inc_ld = v_o & yumi_i & decode_v_r.ld_op;
-  wire inc_st = v_o & yumi_i & decode_v_r.st_op;
-  wire inc_ld_miss = v_o & yumi_i & decode_v_r.ld_op & miss_v;
-  wire inc_st_miss = v_o & yumi_i & decode_v_r.st_op & miss_v;
+  wire inc_ld       = v_o & yumi_i & decode_v_r.ld_op;
+  wire inc_st       = v_o & yumi_i & decode_v_r.st_op;
+  wire inc_mask     = v_o & yumi_i & decode_v_r.mask_op;
+  wire inc_sigext   = v_o & yumi_i & decode_v_r.sigext_op;
```
tommydcjung, line 54: instead of counting it this way, we should separate it this way.
for load: LB, LH, LW, LBU, LHU.
for store: SW, SH, SB
Notice that store is done using SM, so you need to count the number of 1s in the mask bits to figure out which store it is.

bornaehsani: Oh I see, will do, thanks Tommy!

bornaehsani:
Deprecated. Replaced by PR #280 .

PR #276, opened 2020-04-09
stall_force_wb WAW hazard

tommydcjung:


v/vanilla_bean/vanilla_core.v
```diff
@@ -1346,4 +1346,25 @@ module vanilla_core
...
+        assert(int_remote_load_resp_rd_i != wb_r.rd_addr)
+          else $error("[ERROR][VCORE] STALL_FORCE_WB WAW HAZARD with WB_R!!! t=%t, exe_pc=%x", $time, exe_pc);
+      end
+
+    end
+  end
+
+  // synopsys translate_on
+
+
```
vb000, line 1369: If the remote load returned while the corresponding instruction is still in, wb stage for instance, doesn't this assert fail?

tommydcjung: remote load sends out the request in EXE stage, and disappear. so the corresponding instruction wouldn't be in the pipeline.

vb000: Disappear meaning a nop is inserted to the next stage?

tommydcjung: That's right.

vb000: That's cool! 

Just trying to understand all scenarios, what if back-to-back remote loads returns in correct order and the input fifo is full. That's not an error scenario, we should stall a cycle and insert the load into the pipeline, correct?

tommydcjung: You cannot insert into the pipeline when it's stalled. If the fifo is full, and it cannot be inserted into the pipeline, it writes directly into regfile.

vb000: Seems like a deadlock... let's say we have two back-to-back remote loads. The first one returned and we opportunistically inserted into the pipeline with a store for eg., and the second one returns and if the pipeline is stalled for mul for eg. And if at the same time, input fifo is full, we have to write into the reg file, but it'd mean a WAW violation! What do you think?

tommydcjung: that's the bug this is trying to reproduce...

vb000: Oh, thought we don't consider two back to back loads to same reg a dependency at all, to save a cycle. Do we consider that a dependency in the current implementation?

My concern is that this throws an error for many valid write-after-writes... do you intend this to merge into master right away? Or still working on it?

tommydcjung: I think I understand what you are trying to ask. If the second load returns first and it's inserted into the pipeline, and it gets stuck in the MEM or WB stage for a long time, and then the first load is forcibly written into the regfile. After the pipeline unfreezes the second load is written to the regfile. In that case, the assertion will be triggered, but the result is correct. Is that what you are trying to get at?

vb000: Yes.

tommydcjung: Yeah unfortunately I don't think there is any way to differentiate between two cases, which makes this bug particularly tricky and lethal. I think we are trying to understand how often this condition occurs in the wild, and understand what kind of code creates this kind of binary, and figure out what might be the software fix to prevent. So far we haven't encountered any occurrences other than the small assembly test I created.

vb000: It's not a weird case, it's just rare.

I just figured, the bug can be seen any real program with a load followed by a loop/conditional with load to same register. That would involve back to back loads to the same register without a dependency! I'll write up a C program doing this after a while.

taylor-bsg:
So what I am suggesting is that we only generate a warning in a true error.
It’s okay to have other warnings that you can enable but that should not
fire for the Cornell people. For the bug isn’t it sufficient to detect if
there is a emergency write to the RF at the same time a write to the same
register exists in the pipeline?

v/vanilla_bean/vanilla_core.v
tommydcjung: Actually the case you are describing will not happen? The second load cannot be launched unless the first load returns and either it's inserted into the pipeline or force-written into the regfile. So I think the current assertion is sufficient for all cases.

vb000: Consider following case:
```
# A bunch of slow remote loads that return immediately after
# mem[remote_addr2] returns
lw x7, <remote addr3>
lw x7, <remote addr3>
lw x7, <remote addr3>

# Problem scenario
lw x5, <remote addr1>
mul x6, x0, x1
sw x6, <>
lw x5, <remote_addr2>
mul x2, x0, x1
```

If `mem[remote_addr1]` returned coinciding with `sw x6, <>`, and `mem[remote_addr2]` arrived during md stall next, `mem[remote_addr1]` would still be in wb stage, causing the assertion to fail.

tommydcjung: yes, that's the case the assertion is trying to catch, and what the test reproduces...

vb000: Okay, I want to clarify, are there two bugs? One, above scenario. Two, WAW dependency is ignored to save a cycle (?).

tommydcjung:
sounds good. Looking forward to see the C program that reproduces this bug :)

tommydcjung:
Switched to the bind module model, and it prints the PCs of the load that overwrites and is overwritten.

testbenches/common/v/nb_waw_detector.v
```diff
@@ -0,0 +1,92 @@
...
+    end
+  end
+
+
+  always_ff @ (negedge clk_i) begin
+    if (int_remote_load_resp_v_i & int_remote_load_resp_force_i) begin
+
+      if (mem_r.op_writes_rf) begin
+        assert(int_remote_load_resp_rd_i != mem_r.rd_addr)
+          else $error("[ERROR][VCORE] STALL_FORCE_WB WAW HAZARD WITH MEM_R!!! time=%0t, x=%0d, y=%0d, rd=x%0d, aggressor_pc=%x, victim_pc=%x",
```
vb000, line 78: @tommydcjung Since we expect software developers to add a software patch whenever they encounter this scenario, would it help to have the message explicitly say:

```
[ERROR][VCORE] STALL_FORCE_WB WAW HAZARD WITH MEM_R!!! time=%0t, x=%0d, y=%0d, rd=x%0d, aggressor_pc=%x, victim_pc=%x"
This condition will trigger a hardware bug, please include a WAW software patch to avoid this scenario at pc = %x.
```

PR #277, opened 2020-04-10
Vanilla bean hash function fix

mrutt92:
Fixes a bug in EVA => NPA translation for DRAM addresses on designs with power of 2 columns.


tommydcjung:
I think you should create a ci_ branch?

mrutt92:
will do

PR #278, opened 2020-04-10
size of infinite memories scale down inversely to columns

mrutt92:
fixes a bug where infinite memories were constant size regardless of the number of columns

### DIFF
--- a/testbenches/common/v/bsg_nonsynth_mem_infinite.v
+++ b/testbenches/common/v/bsg_nonsynth_mem_infinite.v
@@ -13,6 +13,7 @@ module bsg_nonsynth_mem_infinite
     , parameter x_cord_width_p="inv"
     , parameter y_cord_width_p="inv"
     , parameter id_p="inv"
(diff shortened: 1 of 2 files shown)


testbenches/common/v/bsg_nonsynth_mem_infinite.v
```diff
@@ -13,6 +13,7 @@ module bsg_nonsynth_mem_infinite
     , parameter x_cord_width_p="inv"
     , parameter y_cord_width_p="inv"
     , parameter id_p="inv"
+    , parameter infmem_addr_width_p=(addr_width_p-x_cord_width_p-1)
```
tommydcjung, line 16: I would have a parameter called mem_els_p instead without the default value. This default value is more specific to way EVA map to vanilla core, but we can imagine having an accelerator that can access the full address range of EPA. mem_els_p should be set to some value that is appropriate for the testbench where this module is instantiated.
And of course, we need to set mem_addr_li so that it only takes lg_mem_els_lp bits from the packet.

PR #279, opened 2020-04-11
helping tip for identifying colors in MacOS

taylor-bsg:


tommydcjung:
Maybe there should be a mode to assist with color-blindness.

taylor-bsg:
Borna has set it up so you can do your own custom colors; “build the door
frame, not the extra room.”

dpetrisko:
I believe matplotlib/seaborn have a bunch of color sets designed for high contrast as well as being disability-friendly. Perhaps a future enhancement is to support these packs, in addition to custom schemes

taylor-bsg:
I did look a little bit into this. If there is an off-the shelf option
that’s seems adequate that’s great, but apparently from what I read, color
blindness is not standard in terms of what people see, so the
configurability that Borna provides may actually be the best solution.

drichmond:
Everyone has valid points. I see no reason we can't merge this. What's the standard in manycore? Does the submitter merge?

tommydcjung:
It was just a random thought. just merge it?

PR #280, opened 2020-04-12
Blocking VCache stall graph and stats parser

bornaehsani:
Adds stats parser and stall graph for blocking victim cache 
Major changes to vnialla_stats_parser to also include vcache stats

Merge with PR # 543 in bsg_replicant.

### DIFF
--- a/testbenches/common/v/vcache_profiler.v
+++ b/testbenches/common/v/vcache_profiler.v
@@ -46,10 +46,27 @@ module vcache_profiler
   // event signals
   //
 
-  wire inc_ld = v_o & yumi_i & decode_v_r.ld_op;
-  wire inc_st = v_o & yumi_i & decode_v_r.st_op;
-  wire inc_ld_miss = v_o & yumi_i & decode_v_r.ld_op & miss_v;
-  wire inc_st_miss = v_o & yumi_i & decode_v_r.st_op & miss_v;
+  wire inc_ld       = v_o & yumi_i & decode_v_r.ld_op;
+  wire inc_st       = v_o & yumi_i & decode_v_r.st_op;
+  wire inc_mask     = v_o & yumi_i & decode_v_r.mask_op;
+  wire inc_sigext   = v_o & yumi_i & decode_v_r.sigext_op;
+  wire inc_tagst    = v_o & yumi_i & decode_v_r.tagst_op;
+  wire inc_tagfl    = v_o & yumi_i & decode_v_r.tagfl_op;
+  wire inc_taglv    = v_o & yumi_i & decode_v_r.taglv_op;
+  wire inc_tagla    = v_o & yumi_i & decode_v_r.tagla_op;
+  wire inc_afl      = v_o & yumi_i & decode_v_r.afl_op;
+  wire inc_aflinv   = v_o & yumi_i & decode_v_r.aflinv_op;
+  wire inc_ainv     = v_o & yumi_i & decode_v_r.ainv_op;
+  wire inc_alock    = v_o & yumi_i & decode_v_r.alock_op;
+  wire inc_aunlock  = v_o & yumi_i & decode_v_r.aunlock_op;
+  wire inc_tag_read = v_o & yumi_i & decode_v_r.tag_read_op;
+  wire inc_atomic   = v_o & yumi_i & decode_v_r.atomic_op;
+  wire inc_amoswap  = v_o & yumi_i & decode_v_r.amoswap_op;
+  wire inc_amoor    = v_o & yumi_i & decode_v_r.amoor_op;
+
+  wire inc_ld_miss  = v_o & yumi_i & decode_v_r.ld_op & miss_v;
+  wire inc_st_miss  = v_o & yumi_i & decode_v_r.st_op & miss_v;
+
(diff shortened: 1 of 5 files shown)


bornaehsani:
@tommydcjung Does this count different types of loads/stores correctly? I'm using the cache packet's opcode to determine different load/store types.

testbenches/common/v/vcache_profiler.v
```diff
@@ -23,38 +25,128 @@ module vcache_profiler
...
   //
 
-  wire inc_ld = v_o & yumi_i & decode_v_r.ld_op;
-  wire inc_st = v_o & yumi_i & decode_v_r.st_op;
-  wire inc_ld_miss = v_o & yumi_i & decode_v_r.ld_op & miss_v;
-  wire inc_st_miss = v_o & yumi_i & decode_v_r.st_op & miss_v;
+  wire inc_miss     = miss_v;
+
+  wire inc_ld       = v_o & yumi_i & decode_v_r.ld_op;
+  wire inc_ld_ld    = v_o & yumi_i & decode_v_r.ld_op & (cache_pkt.opcode == LD);
```
tommydcjung, line 66: We are recording stats, when the instruction leaves tag-verify stage, which is the last stage. But cache_pkt.opcode is the signal from the input, so the timing does not align. cache_pkt.opcode is decoded by this module, and the decoded signal is passed down to the later stage. https://github.com/bespoke-silicon-group/basejump_stl/blob/master/bsg_cache/bsg_cache_decode.v
You want to use decode_v_r.data_size_op, decode_v_r.sigext_op, etc to figure out which load it is. For store, since we are using SM, we have to count the number of 1s in mask_v_r to figure out which store.

bornaehsani: Ahhh I get it now, thanks! I've made the change, how does it look now?

bornaehsani:
@tommydcjung thanks for explaining the issues with this PR to me! I think I've finally fixed them, do they look correct to you? Thanks!

testbenches/common/v/vcache_profiler.v
```diff
@@ -30,31 +33,107 @@ module vcache_profiler
...
+  wire inc_ld_lhu   = v_o & yumi_i & decode_v_r.ld_op & ~decode_v_r.mask_op & decode_v_r.data_size_op == 2'b01 & ~decode_v_r.sigext_op; // load half unsigned
+  wire inc_ld_lb    = v_o & yumi_i & decode_v_r.ld_op & ~decode_v_r.mask_op & decode_v_r.data_size_op == 2'b00 & decode_v_r.sigext_op;  // load byte
+  wire inc_ld_lbu   = v_o & yumi_i & decode_v_r.ld_op & ~decode_v_r.mask_op & decode_v_r.data_size_op == 2'b00 & ~decode_v_r.sigext_op; // load byte unsigned
+  wire inc_ld_lm    = v_o & yumi_i & decode_v_r.ld_op & decode_v_r.mask_op; // load mask
+
+  wire inc_st       = v_o & yumi_i & decode_v_r.st_op;
+  wire inc_st_sd    = v_o & yumi_i & decode_v_r.st_op & ~decode_v_r.mask_op & ($countones(mask_v_r) == 8); // store double (reserved for 64-bit)
+  wire inc_st_sw    = v_o & yumi_i & decode_v_r.st_op & ~decode_v_r.mask_op & ($countones(mask_v_r) == 4); // store word
+  wire inc_st_sh    = v_o & yumi_i & decode_v_r.st_op & ~decode_v_r.mask_op & ($countones(mask_v_r) == 2); // store half
+  wire inc_st_sb    = v_o & yumi_i & decode_v_r.st_op & ~decode_v_r.mask_op & ($countones(mask_v_r) == 1); // store byte
```
tommydcjung, line 73: ~decode_v_r.mask_op -> decode_v_r.mask_op

Have you checked to see if the numbers make sense?

bornaehsani: I've checked the numbers to make sure they all add up to the correct values, i.e. sum of all the different types of stores / loads add up to the total load/store count given by the profiler, which it does now. 
As to the mask_op, I thought this signal only goes high when the operation is SM/LM, and not for other types of load/store, becuase of this line of code in bsg_cache_decode: 

https://github.com/bespoke-silicon-group/basejump_stl/blob/36b35d6d087defefbb52a839ee00166fe480e342/bsg_cache/bsg_cache_decode.v#L29

You're saying the mask goes high for all store types?

tommydcjung: It should be. bsg_manycore_link_to_cache uses SM to all the stores.

drichmond:
Can we try to get this merged today? I think @mrutt92 and @eafurst would like to use it

bornaehsani:
> Can we try to get this merged today? I think @mrutt92 and @eafurst would like to use it

Yes Tommy and I have a meeting today to have it settled and fixed.

testbenches/common/v/vcache_profiler.v
bornaehsani: @tommydcjung I believe this fixes the issues we talked about this morning. Shall we have it merged? Thank you Tommy!

PR #283, opened 2020-04-13
NB WAW Testcase

vb000:
Recreated the pytorch buggy scenario in a SPMD testcase. This also sweeps across different delays to catch similar bug with various manycore configurations.


drichmond:
Nice. This seems useful to have as a default test in replicant as well, since we test slightly different machines.

tommydcjung:
I would say hold onto this, and we can add it to the regression once the fix is implemented.

PR #287, opened 2020-04-14
Adds Templatized Barrier

bornaehsani:
- Adds a new library bsg_tile_group_barrier.hpp that supports templatized tile group barries
- Updated all cuda lite runtime tests to use the new templatized barrier
- Switched the previous barrier test to barrier_deprecated test
- Adds the bsg_manycore.hpp library to eventually replace the macros in bsg_manycore.h

Cosim regression passed.
**Merge with PR # 552 in bsg_replicant**


bornaehsani:
@mrutt92 @eafurst @save-buffer I'm tagging you as reviewers just so you're aware that you can use the new template barrier from now on. Thanks!

mrutt92:
🎊 🔥 looks super cool!

software/bsg_manycore_lib/bsg_tile_group_barrier.hpp
```diff
@@ -0,0 +1,294 @@
...
+    void reset() {
+        for (int i = 0; i < BARRIER_X_DIM; i ++) {
+            unsigned int* done_list_ptr = const_cast<unsigned int*> (&_done_list[i]);
+            *done_list_ptr = 0;
+        }
+        return;
+    };
+
+
+    // send the sync singal to the center tile of the row
```
dpetrisko, line 72: signal

bornaehsani: woopsie! Thanks!

software/bsg_manycore_lib/bsg_tile_group_barrier.hpp
```diff
@@ -0,0 +1,285 @@
...
+        do{
+                for( i= 0; i <= range; i++) {
+                        if ( p[ i ] == 0) break;
+                }
+        }while ( i <= range);
+}
+
+// Wait until the specified memory address is written by other tiles
+// with specific value
+inline int bsg_wait_local_int(int * ptr,  int cond ) {
```
tommydcjung, line 271: Isn't this function already in bsg_mutex.h? I'm not sure why it needs to be duplicated here.

tommydcjung: I'm not necessary suggesting that this file should include "bsg_mutex.h"; there might be a better way to organize shared codes.

bornaehsani: Oh you're right my bad. Fixed, thanks!

software/bsg_manycore_lib/bsg_manycore.hpp
```diff
@@ -0,0 +1,21 @@
+#ifndef _BSG_MANYCORE_HPP_
+#define _BSG_MANYCORE_HPP_
+
+#include <cstdint>
+#include <bsg_manycore_arch.h>
+
+
+
+template<typename T>
+T *bsg_remote_pointer(unsigned char x, unsigned char y, T* local_addr) {
```
tommydcjung, line 10: I think this is using tile-group EVA space? "bsg_remote_pointer" seems like a really vague name. maybe others can chime in. Also we could probably use this macro.
https://github.com/bespoke-silicon-group/bsg_manycore/blob/master/software/bsg_manycore_lib/bsg_manycore_arch.h#L68

software/bsg_manycore_lib/bsg_tile_group_barrier.hpp
```diff
@@ -0,0 +1,266 @@
...
+        }
+        return;
+    };
+
+
+    // send the sync signal to the center tile of the row
+    // executed by all tiles in the group.
+    void sync (unsigned char center_x_cord) {
+        //write to the corresponding done
+        volatile unsigned int *done_list_ptr = const_cast<unsigned int*> (reinterpret_cast<volatile unsigned int*> (bsg_remote_pointer(center_x_cord, bsg_y, &_done_list[bsg_x - _x_cord_start])));
```
tommydcjung, line 71: Isn't _x_cord_start always zero? If it's always zero, is there really a need to have this constant? (I'm assuming this is safe since we are using tile-group EVA).
and we can replace whereever _x_cord_end is used with BARRIER_X_DIM?

software/bsg_manycore_lib/bsg_tile_group_barrier.hpp
```diff
@@ -0,0 +1,266 @@
...
+                }
+        #endif
+        return;
+    };
+};
+
+
+// Helper funcitons.
+// check if the char array are all non-zeros
+inline void poll_range( int range, volatile const unsigned int *p){
```
tommydcjung, line 255: is there a reason why we are using int array, when the comment says it's char array? I think we are storing either zero or one, so the char array is okay?

bornaehsani: Ahh no actually, that's a good point! I changed them to char*, thanks!

software/bsg_manycore_lib/bsg_manycore.hpp
bornaehsani: Yes but the whole point of adding templatized c++ barriers is to move away from macros, and get more visibility/type safetyness from c++ code. the bsg_remote_pointer is basically the templatized version of the bsg_remote_ptr macro. And I'm putting it in a separate file, so it gets populated over time with replacements for macros in bsg_mancyore.h

software/bsg_manycore_lib/bsg_tile_group_barrier.hpp
bornaehsani: Yes that makes sense, I was modeling after the legacy code but you're right it's overkill. How does it look now?

tommydcjung:
please create ci_* branch to run the regression. no need to close this branch.

software/bsg_manycore_lib/bsg_manycore.hpp
drichmond: I agree with Tommy on the naming. "Remote Pointer" is too generic, if this is only used for TG pointers bsg_tile_group_remote_pointer would be better, though verbose. It's also more accurate since bsg_tile_group_remote_ptr (the macro in the file above) takes a type like this does.

If you want backward compatibility, have bsg_remote_pointer use bsg_tile_group_remote pointer and mark bsg_remote_pointer as deprecated.

software/bsg_manycore_lib/bsg_tile_group_barrier.hpp
```diff
@@ -0,0 +1,260 @@
...
+
+
+template <int BARRIER_X_DIM>
+class bsg_row_barrier {
+public:
+    volatile unsigned int     _local_alert = 0;
+    volatile unsigned char     _done_list[ BARRIER_X_DIM ] = {0};
+
+
+    bsg_row_barrier (){};
```
drichmond, line 52: Why not call reset when the object is constructed, just to be safe.

software/bsg_manycore_lib/bsg_tile_group_barrier.hpp
```diff
@@ -0,0 +1,260 @@
...
+    void wait_on_sync() {
+        poll_range(this->_done_list, BARRIER_Y_DIM);
+        return;
+    };
+
+
+    // wait on local alert to be set to the given value
+    void wait_on_alert (){
+        // wait until _local_alert flag is set to 1
+        bsg_wait_local_int( (int *) &(this->_local_alert), 1);
```
drichmond, line 162: If we're going "whole hog" here, I would replace `bsg_wait_local_int` (which is in a header file that isn't even included) with an overloaded function called bsg_wait_local that has definitions for int and unsigned int in this header file.

software/bsg_manycore_lib/bsg_tile_group_barrier.hpp
```diff
@@ -0,0 +1,260 @@
...
+        #endif
+        return;
+    };
+
+
+    // send alert to all of the tiles in the column 
+    void alert (){
+        for( int y = 0; y < BARRIER_Y_DIM; y ++) {
+               volatile unsigned int *alert_ptr = const_cast<unsigned int*> (reinterpret_cast<volatile unsigned int*> (bsg_remote_pointer(bsg_x, y, &_local_alert)));
+               *alert_ptr = 1;
```
drichmond, line 144: I am amazed and scared by this line. What does it do?

software/bsg_manycore_lib/bsg_tile_group_barrier.hpp
```diff
@@ -0,0 +1,260 @@
...
+    // executed by all tiles in the center row
+    void sync(unsigned char center_x_cord, unsigned char center_y_cord ){
+        //write to the corresponding done
+        volatile unsigned char *done_list_ptr = const_cast<unsigned char*> (reinterpret_cast<volatile unsigned char*> (bsg_remote_pointer(center_x_cord, center_y_cord, &_done_list[bsg_y])));
+        *done_list_ptr = 1;
+
+        #ifdef BSG_BARRIER_DEBUG
+               //addr 0x0: row sync'ed
+               bsg_remote_ptr_io_store( IO_X_INDEX, 0x0, bsg_y);
+        #endif
```
drichmond, line 135: I would say the same about moving everything to C++ here, but I'm willing to let it slide because it's a debug statement.

software/bsg_manycore_lib/bsg_manycore.hpp
```diff
@@ -0,0 +1,21 @@
...
+        uintptr_t remote_prefix = (REMOTE_EPA_PREFIX << REMOTE_EPA_MASK_SHIFTS);
+        uintptr_t y_bits = ((y) << Y_CORD_SHIFTS);
+        uintptr_t x_bits = ((x) << X_CORD_SHIFTS);
+        uintptr_t local_bits = reinterpret_cast<uintptr_t>(local_addr);
+        return reinterpret_cast<T *>(remote_prefix | y_bits | x_bits | local_bits);
+}
+
+
```
drichmond, line 18: Can you copy over the templated version in baseline as well?

drichmond: Actually, if you "copy over" the templated version, just have it call the non-templated version above

software/bsg_manycore_lib/bsg_tile_group_barrier.hpp
```diff
@@ -0,0 +1,260 @@
...
+
+
+template <int BARRIER_Y_DIM>
+class bsg_col_barrier {
+public:
+    volatile unsigned int      _local_alert = 0;
+    volatile unsigned char      _done_list[ BARRIER_Y_DIM ] = {0};
+
+
+    bsg_col_barrier (){};
```
drichmond, line 113: Same with reset here

software/bsg_manycore_lib/bsg_tile_group_barrier.hpp
```diff
@@ -0,0 +1,260 @@
...
+                if( bsg_x == _center_x_cord && bsg_y == _center_y_cord ){
+                        bsg_print_time();
+                }
+        #endif
+        return;
+    };
+};
+
+
+// Helper funcitons.
```
drichmond, line 247: FWIW, functions

software/bsg_manycore_lib/bsg_tile_group_barrier.hpp
```diff
@@ -0,0 +1,260 @@
...
+// check if the char array are all non-zeros
+inline void poll_range(volatile const unsigned char *p, int range);
+
+
+template <int BARRIER_X_DIM>
+class bsg_row_barrier {
+public:
+    volatile unsigned int     _local_alert = 0;
```
save-buffer, line 48: Why do these have to be public?

software/bsg_manycore_lib/bsg_tile_group_barrier.hpp
```diff
@@ -0,0 +1,260 @@
...
+    };
+};
+
+
+template <int BARRIER_Y_DIM>
+class bsg_col_barrier {
+public:
+    volatile unsigned int      _local_alert = 0;
+    volatile unsigned char      _done_list[ BARRIER_Y_DIM ] = {0};
+
```
save-buffer, line 111: Same - why are they public?

software/bsg_manycore_lib/bsg_tile_group_barrier.hpp
```diff
@@ -0,0 +1,260 @@
...
+               bsg_remote_ptr_io_store( IO_X_INDEX, 0x0, bsg_y);
+        #endif
+        return;
+    };
+
+
+    // send alert to all of the tiles in the column 
+    void alert (){
+        for( int y = 0; y < BARRIER_Y_DIM; y ++) {
+               volatile unsigned int *alert_ptr = const_cast<unsigned int*> (reinterpret_cast<volatile unsigned int*> (bsg_remote_pointer(bsg_x, y, &_local_alert)));
```
save-buffer, line 143: Why do you cast away `volatile` and then reapply it? Seems like you could just get rid of the `const_cast`.

software/bsg_manycore_lib/bsg_tile_group_barrier.hpp
save-buffer, line 162: I'd use a C++ cast here to be consistent with the rest of the code

software/bsg_manycore_lib/bsg_tile_group_barrier.hpp
```diff
@@ -0,0 +1,260 @@
...
+            *done_list_ptr = 0;
+        }
+        return;
+    };
+
+    // send the sync signal to the center tile of the column
+    // executed by all tiles in the center row
+    void sync(unsigned char center_x_cord, unsigned char center_y_cord ){
+        //write to the corresponding done
+        volatile unsigned char *done_list_ptr = const_cast<unsigned char*> (reinterpret_cast<volatile unsigned char*> (bsg_remote_pointer(center_x_cord, center_y_cord, &_done_list[bsg_y])));
```
save-buffer, line 129: Same as below here with the `const_cast`

software/bsg_manycore_lib/bsg_tile_group_barrier.hpp
```diff
@@ -0,0 +1,260 @@
...
+        }
+        return;
+    };
+
+
+    // send the sync signal to the center tile of the row
+    // executed by all tiles in the group.
+    void sync (unsigned char center_x_cord) {
+        //write to the corresponding done
+        volatile unsigned char *done_list_ptr = const_cast<unsigned char*> (reinterpret_cast<volatile unsigned char*> (bsg_remote_pointer(center_x_cord, bsg_y, &_done_list[bsg_x])));
```
save-buffer, line 69: Same as below here with the `const_cast`

software/bsg_manycore_lib/bsg_tile_group_barrier.hpp
```diff
@@ -0,0 +1,260 @@
...
+#include "bsg_set_tile_x_y.h"
+#include "bsg_manycore.h"
+#include "bsg_mutex.h"
+#include "bsg_manycore.hpp"
+
+
+
+
+
+// Helper funcitons.
```
save-buffer, line 40: nit: typo

software/bsg_manycore_lib/bsg_tile_group_barrier.hpp
```diff
@@ -0,0 +1,260 @@
...
+//Worst Case Performance:
+//      1. row sync     :   X_DIM     cycles //all tiles writes to center tile of the row
+//      2. row polling  : 3*X_DIM     cycles // lbu <xx>; beqz <xxx;
+//      3. col sync     :   Y_DIM     cycles //all tiles writes to the center tiles of the col
+//      4. col polling  : 3*Y_DIM     cycles // lbu <xx>; beqz <xx>;
+//      5. col alert        Y_DIM     cycles // store
+//      6. row alert        X_DIM     cycles // store
+//      -----------------------------------------------
+//                        5*( X_DIM + Y_DIM)
+//      For 3x3 group,  cycles = 181, heavy looping/checking overhead for small groups.
```
save-buffer, line 25: This is a great comment!!

tommydcjung:
> As is, this is good enough to be merged. I left a few comments to address.
> 
> As a general question though, why not split the done_list and alert variable into separate classes? Then you can move poll_range, and reset into the done_list class and wait_on_alert into the local_alert class and you won't have duplicated code.
> 
> I think you'll still need separate barriers for column and row, but you'd have less code to maintain

I feel like that would be too granular of classes. maybe it can be refactored as  barrier<1,y> or barrier<x,1>. Or maybe there is even simpler way.

drichmond:
I don't think having too many classes is a problem, but it should be balanced against engineering time. 

barrier<1,y>/<x,1> is one solution. You can also do a boolean (for col/row) and then the dimension. Specialization would let you have separate methods for the column/row sync. 

But at the end of the day, Borna has finished what he set out to do. Barring a few fixes, it's ready to go.

software/bsg_manycore_lib/bsg_manycore.hpp
bornaehsani: I agree with both of you on the confusing name, it makes sense to rename it to bsg_tile_group_remote_pointer. But as far as backward compatibility goes, the compiler barks at me for deprecating a template function, plus bsg_remote_pointer is not used anywhere as just added it for the purposes of template barrier, so I just renamed it to bsg_tile_group_remote_pointer and removed the bsg_remote_pointer. Does that make sense?

software/bsg_manycore_lib/bsg_tile_group_barrier.hpp
bornaehsani: I added a `bsg_mutex`.hpp file that includes a template version of `bsg_wait_local_int` called `bsg_wait_local`. How is that?

software/bsg_manycore_lib/bsg_tile_group_barrier.hpp
bornaehsani: Good idea!

software/bsg_manycore_lib/bsg_tile_group_barrier.hpp
bornaehsani: Good point! fixed, thanks!

software/bsg_manycore_lib/bsg_tile_group_barrier.hpp
bornaehsani: Yeah you're right, it was leftover from a previous commit where I declared _done_list and _local_alert as constexpr. Fixed, thanks!

software/bsg_manycore_lib/bsg_tile_group_barrier.hpp
bornaehsani: Great catch! Fixed, thanks!

software/bsg_manycore_lib/bsg_tile_group_barrier.hpp
bornaehsani: Yes it is! But it's not mine I copied it from Shaolin's code :D

software/bsg_manycore_lib/bsg_tile_group_barrier.hpp
bornaehsani: Yeah that was overly complicated for no reason, how does it look now?

software/bsg_manycore_lib/bsg_manycore.hpp
bornaehsani: I don't understand what you mean, copy what over?

software/bsg_manycore_lib/bsg_manycore.hpp
drichmond: Yeah, makes sense.

software/bsg_manycore_lib/bsg_tile_group_barrier.hpp
drichmond: Good. My suggestion (instead of having templates) was two have two definitions though. I think having a generic version is technically incorrect (and could cause a compiler error) when someone waits on a float/char/etc.

I think what you want is `bsg_wait_local(int, ...)` and `bsg_wait_local(unsigned int, ...)`. C++ is fine with overloading (I think this case is fine)

software/bsg_manycore_lib/bsg_tile_group_barrier.hpp
drichmond: Good

software/bsg_manycore_lib/bsg_tile_group_barrier.hpp
bornaehsani: Yeah you're right! I made the change, how does it look? thanks!

drichmond: Great!

drichmond:
Great! I think we need to run this through ci though

drichmond:
CI is running here: https://gitlab.com/dpetrisko/bsg_manycore/pipelines/137337601

bornaehsani:
Started a fresh CI, will merge if it passes. Thank you Dustin, Tommy, Sasha, Max, Dan and Emily for your comments!!

Issue #281, opened 2020-04-13
Bring back the --cycle option for blood graph


bornaehsani:
Currently, bloodgraph takes in vanilla_stats for timing stats to know what the start and end cycle of the graph should be, but generating bloodgraph should not depend on the stats, but only the traces. So you should bring back the --cycle option as an overriding optional argument.

taylor-bsg:
I think specifying a range is an important option, but maybe also getting
the start and end from the trace file itself would be a useful feature, e.g.

https://github.com/black-parrot/black-parrot/commit/98df9ae4d4990b3a18afeb2c69875642d02aec0a

bornaehsani:
@taylor-bsg  Oh I see, that makes perfects sense! So to my understanding the priority of these options should be in the following order:

1. Default: use the first and last trace from the trace file as start and end cycle
2. If the stats file is given as input, use the first kernel_start and last kernel_end stat --> this narrows the bloodgraph to the window in which the tiles are actually running the kernel
3. If the --cycle option is given, override and use the custom start/end cycle given by the user

Does this make sense? Thanks!

taylor-bsg:
Yep!

bornaehsani:
Addressed in PR #288 . Thanks!

PR #288, opened 2020-04-15
Extending blood graph timing window options

bornaehsani:
Addresses issue #281 .
Adds three options for choosing the timing window of blood graphs:
1. --trace: default, use the first and last trace from the trace file as start and end cycle
2. --stats: if the stats file is given as input, use the first kernel_start and last kernel_end stat --> if given as input, it overrides option 1 and narrows the bloodgraph to the window in which the tiles are actually running the kernel
3. --cycle: custom start/end cycles given by the user which overrides the other two 

Minor restructuring of functions and variables as well.

**Merge with PR # 29 in baseline**

### DIFF
--- a/software/py/blood_graph.py
+++ b/software/py/blood_graph.py
@@ -29,6 +29,8 @@
 import sys
 import csv
 import argparse
+import warnings
+import os.path
 from PIL import Image, ImageDraw, ImageFont
 from itertools import chain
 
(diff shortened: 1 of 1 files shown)


software/py/blood_graph.py
```diff
@@ -37,282 +39,322 @@ class BloodGraph:
...
-                              "icache_miss",
-                              "stall_ifetch_wait",
-                              "bubble_icache",
-                              "bubble_branch_mispredict",
-                              "bubble_jalr_mispredict",
-                              "bubble_fp_op",
-                              "bubble",
-                              "stall_md" ]
+    # List of types of stalls incurred by the core 
+    _STALLS_LIST   = ["stall_depend_remote_load_dram",
```
drichmond, line 45: Good. This is a better design pattern. What motivated you to do it?

software/py/blood_graph.py
```diff
@@ -37,282 +39,322 @@ class BloodGraph:
...
-                                             "bubble"                                 : (0x00, 0x00, 0x00), ## black
 
-                                             "stall_md"                               : (0xff, 0xff, 0xff), ## white
-                                           }
-        self.abstract_unified_instr_color    =                                          (0xff, 0xff, 0xff)  ## white
-        self.abstract_unified_fp_instr_color =                                          (0xff, 0xff, 0xff)  ## white
+    # look through the input file to get the tile group dimension (x,y)
+    def __get_tile_group_dim(self, traces):
+        xs = list(map(lambda t: t["x"], traces))
+        ys = list(map(lambda t: t["y"], traces))
```
drichmond, line 311: Educational moment: 

Instead of iterating over the list twice, you can do a list comprehension: 

`[t["x"] for t in traces]` 

accomplishes the same thing and is usually faster

software/py/blood_graph.py
bornaehsani: I suppose my coding style improved a bit over time thanks to your comments on my PR's over time :D when I came back to iterate over this code, there seemed to be many things wrong with it that I didn't see before.

software/py/blood_graph.py
bornaehsani: Oh yeah that's a good point! I fixed it, thanks!

bornaehsani:
Thanks Dustin!

PR #289, opened 2020-04-15
add retry for other steps as well

tommydcjung:


drichmond:
Seems like a great idea!

PR #293, opened 2020-04-17
Fixed an VCS-specific format string

drichmond:
In verilator %- causes compilation to fail. %0 does not. I think this fixes it, with the same functionality?


drichmond:
Whoops, some files snuck in. One sec

PR #294, opened 2020-04-17
Removed stale file

drichmond:

PR #295, opened 2020-04-17
Fixed old file for new manycore parameters

drichmond:


drichmond:
Instead of removing it in #294, I fixed it.

drichmond:
Can we merge this?

tommydcjung:
I think it makes for sense to remove it from the hetero socket, since it's not going to work functionally or even synthesize correctly. I would say just keep the file for sentimental value, and create an issue to fix it in future.

drichmond:
@taylor-bsg @tommydcjung 

This is blocking verilator progress now. Can we come to a conclusion on the way forward? I've fixed this file but I don't know of any test that checks the functionality.

drichmond:
Should someone be writing a testbench here?

taylor-bsg:
I think it’s okay to just indicate that it has been update to build but has not been tested recently.

drichmond:
OK, I've added a fatal statement in an initial block. If anyone tries to instantiate this module it will fail during simulation -- thereby forcing them to give explicit acknowledgement that this module is untested.

drichmond:
How are we feeling about this? Is my solution acceptable?

taylor-bsg:
LGTM

PR #296, opened 2020-04-17
Removed old files from filelist

drichmond:


drichmond:
Tommy says these are no longer needed, because max wrote a C++ wrapper. CI passes, can we merge this?

PR #297, opened 2020-04-17
Fix multi-driven (?) net issue

drichmond:
Fixed a (Verilator-percieved) issue where a net was driven by both blocking and non-blocking assignments

### DIFF
--- a/v/bsg_manycore.v
+++ b/v/bsg_manycore.v
@@ -100,17 +100,20 @@ module bsg_manycore
   // Pipeline the reset. The bsg_manycore_tile has a single pipeline register
   // on reset already, so we only want to pipeline reset_depth_p-1 times.
   logic [reset_depth_p-1:0][num_tiles_y_p-1:0][num_tiles_x_p-1:0] reset_i_r;
+  logic [num_tiles_y_p-1:0][num_tiles_x_p-1:0]                    reset_i_lo;
+   
+  assign reset_i_lo = {(num_tiles_y_p*num_tiles_x_p){reset_i}};
 
-  assign reset_i_r[0] = {(num_tiles_y_p*num_tiles_x_p){reset_i}};
-
+   always_ff @(posedge clk_i) begin
+      reset_i_r[0] <= reset_i_lo;
+   end
   genvar k;
-  for (k = 1; k < reset_depth_p; k++)
+  for (k = 1; k < reset_depth_p - 1; k++)
     begin
-      always_ff @(posedge clk_i)
-        begin
+       always_ff @(posedge clk_i) begin
           reset_i_r[k] <= reset_i_r[k-1];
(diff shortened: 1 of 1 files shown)


drichmond:
If there's nothing controversial here, can we merge this?

v/bsg_manycore.v
```diff
@@ -100,17 +100,20 @@ module bsg_manycore
...
   // on reset already, so we only want to pipeline reset_depth_p-1 times.
   logic [reset_depth_p-1:0][num_tiles_y_p-1:0][num_tiles_x_p-1:0] reset_i_r;
+  logic [num_tiles_y_p-1:0][num_tiles_x_p-1:0]                    reset_i_lo;
+   
+  assign reset_i_lo = {(num_tiles_y_p*num_tiles_x_p){reset_i}};
 
-  assign reset_i_r[0] = {(num_tiles_y_p*num_tiles_x_p){reset_i}};
-
+   always_ff @(posedge clk_i) begin
+      reset_i_r[0] <= reset_i_lo;
```
tommydcjung, line 108: Doesn't this add an extra layer of flops? This can cause different parts of chip coming out of reset in a different order, which can be a hard bug to track...

dpetrisko:
How about 
```reset_i_r[k] <= (k == 0) ? {(num_tiles_y_p*num_tiles_x_p){reset_i}} : reset_i_r[k-1];```

File: v/bsg_manycore.v (hunk 1 above)
Comment (drichmond) [reply v/bsg_manycore.v:108]: Yes, my mistake.

**Comment by drichmond** [comment]
@dpetrisko used your solution. Waiting for CI to re-run

File: v/bsg_manycore.v (hunk 1 above)
Comment (drichmond) [reply v/bsg_manycore.v]: Fixed (I believe)

File: v/bsg_manycore.v [hunk 2]
```diff
@@ -100,17 +100,18 @@ module bsg_manycore
   // Pipeline the reset. The bsg_manycore_tile has a single pipeline register
   // on reset already, so we only want to pipeline reset_depth_p-1 times.
   logic [reset_depth_p-1:0][num_tiles_y_p-1:0][num_tiles_x_p-1:0] reset_i_r;
-
-  assign reset_i_r[0] = {(num_tiles_y_p*num_tiles_x_p){reset_i}};
+  logic [num_tiles_y_p-1:0][num_tiles_x_p-1:0]                    reset_i_lo;
+   
+  assign reset_i_lo = {(num_tiles_y_p*num_tiles_x_p){reset_i}};
```
tommydcjung, line 105: These lines are unused, and should be removed.

drichmond: Yes, dan pointed that out too. Can I remove them after you approve this so I do have to run CI again?

PR #298, opened 2020-04-21
Change hetero_type_vec_p to 1-Dimensional Vector

drichmond:
Verilator can't infer constant-ness through 2D vectors passed as parameters ([Issue](https://github.com/verilator/verilator/issues/2272)). 

Therefore, I've switched hetero_type_vec_p to a 1-dimensional vector. 

I updated the Makefile.machine.include files. I also removed tabs, and updated the comment describing the vector.

Is there a way to run regression on all the machines in Manycore? Is there one that tests the scatter gather accelerator?

This is the last major blocking issue to compiling Manycore in Verilator.

### DIFF
--- a/testbenches/common/v/spmd_testbench.v
+++ b/testbenches/common/v/spmd_testbench.v
@@ -22,7 +22,7 @@ module spmd_testbench;
   parameter bsg_manycore_mem_cfg_e bsg_manycore_mem_cfg_p = `BSG_MACHINE_MEM_CFG;
   parameter bsg_branch_trace_en_p = `BSG_MACHINE_BRANCH_TRACE_EN;
   parameter vcache_miss_fifo_els_p = `BSG_MACHINE_VCACHE_MISS_FIFO_ELS;
-  parameter int hetero_type_vec_p [0:num_tiles_y_p-2][0:num_tiles_x_p-1] = '{`BSG_MACHINE_HETERO_TYPE_VEC};
+  parameter integer hetero_type_vec_p [0:((num_tiles_y_p-1)*num_tiles_x_p) - 1]  = '{`BSG_MACHINE_HETERO_TYPE_VEC};
 
   // constant params
   parameter data_width_p = 32;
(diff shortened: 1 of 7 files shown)


v/bsg_manycore.v
```diff
@@ -68,6 +69,7 @@ module bsg_manycore
     , input [num_tiles_x_p-1:0][link_sif_width_lp-1:0] io_link_sif_i
     , output [num_tiles_x_p-1:0][link_sif_width_lp-1:0] io_link_sif_o
   );
+//'{default:0}
```
tommydcjung, line 72: not needed?

drichmond: Huh. Wonder where that came from. Fixed

tommydcjung:
the regression should be running gather_scatter and gather_scatter_multi.

drichmond:
Yeah, I noticed that the latest CI passed those. The older regressions/commits did not, but the CI did not report fail...

drichmond:
Yeah I accidentally rebased an old commit. Fixed now.

testbenches/common/v/spmd_testbench.v
```diff
@@ -22,7 +22,7 @@ module spmd_testbench;
   parameter bsg_manycore_mem_cfg_e bsg_manycore_mem_cfg_p = `BSG_MACHINE_MEM_CFG;
   parameter bsg_branch_trace_en_p = `BSG_MACHINE_BRANCH_TRACE_EN;
   parameter vcache_miss_fifo_els_p = `BSG_MACHINE_VCACHE_MISS_FIFO_ELS;
-  parameter int hetero_type_vec_p [0:num_tiles_y_p-2][0:num_tiles_x_p-1] = '{`BSG_MACHINE_HETERO_TYPE_VEC};
+  parameter integer hetero_type_vec_p [0:((num_tiles_y_p-1)*num_tiles_x_p) - 1]  = '{`BSG_MACHINE_HETERO_TYPE_VEC};
```
tommydcjung, line 25: shouldn't this be int?

drichmond: Yeah, I was debugging something else and this snuck in. Will fix.

drichmond: Fixed.

taylor-bsg:
Have we confirmed that this synthesizes?

M

taylor-bsg:
I see -- probably no need to worry that it would not; we are not changing
the type, just the dimensionality.

Nice verilator bug report.  He's trying to turn you into a maintainer. =)

M

PR #301, opened 2020-04-21
General refactoring of Vanilla PC Histogram

bornaehsani:
- Previously the PC histogram would traverse the entire range of possible PC's, possibly from 0x00000000 to 0xffffffff to generate histogram --> very inefficient and painfully slow
- Modified it to only traverse the PC's seen in the vanilla operation trace file
- Updated commentary
- Refactored and added some functions to make the code more modular 
- Got rid of redundant print stats functions 

**Merge with PR # in baseline.**

### DIFF
--- a/software/py/vanilla_pc_histogram.py
+++ b/software/py/vanilla_pc_histogram.py
@@ -31,12 +31,8 @@ class PCHistogram:
     _DEFAULT_START_CYCLE = 0 
     _DEFAULT_END_CYCLE   = 500000
 
-    _BSG_PC_MIN = 0x0000
-    _BSG_PC_MAX = 0x2000
-
     _BSG_PC_ADDR_SHIFT = 2
     _BSG_PC_ADDR_STEP = 1 << _BSG_PC_ADDR_SHIFT
-
 
 
     # formatting parameters for aligned printing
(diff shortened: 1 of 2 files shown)


software/py/vanilla_pc_histogram.py
```diff
@@ -57,20 +47,27 @@ class PCHistogram:
...
-        # parse vanilla_operation_trace.log
-        with open(input_file) as f:
+        # Generate PC count dictionary for the entire network
+        self.manycore_pc_cnt = self.__generate_manycore_pc_cnt(self.tile_pc_cnt)
+
+        return
+
+
+    # parse trace file and extract traces
+    def __parse_traces(self, trace_file):
```
drichmond, line 67: If we get to the point where this is running out of memory, it would be good to have this method return an iterator or parse the statistics here, so that the whole file doesn't need to be read into memory.

bornaehsani:
> I think you should separate parsing the histogram and printing the histogram. That way, if I wanted to, I could come in and insert matplotlib and generate a picture instead of text.
> 
> If that's easy to do, please do it, so we can generate cool graphs.

That's a good idea! Done! Thanks!

software/py/vanilla_pc_histogram.py
bornaehsani: Not entirely sure what you mean. We parse traces this way in all of our profiler tools, blood graph, stats parser and the PC histogram. I think this is subject for another PR right?

drichmond: Yes, this isn't something to fix. Just a note

bornaehsani:
> Looks good. Have you tried hooking it into matplotlib :p ?

Not yet, maybe some day in the distant future :P Thanks!

PR #306, opened 2020-04-27
put a loop to prevent icache miss on uninitialized addr

tommydcjung:


tommydcjung:
Yes, for c programs we usually have bsg_wait_while(1) at the end.

drichmond:
K. So we don't need to file an issue to fix this in all the other tests?

drichmond:
@tommydcjung

tommydcjung:
I think other tests are fine, but if we find something that needs to be fixed, we can fix it then.

taylor-bsg:
Maybe we should use a standard macro for this so that we can change it
uniformly across the tests. For example, it may be preferable to implement
with load on broken reservation.

M

vb000:
Yes, that makes me realize we should rather fix this macro:
https://github.com/bespoke-silicon-group/bsg_manycore/blob/105369dce3e30747b2e7b01c07558b289bab702a/software/bsg_manycore_lib/bsg_manycore_asm.h#L73-L76

PR #307, opened 2020-04-28
Added DPI Interface for Manycore

drichmond:
Will document in time - and more examples to follow. 

BUT - a question for discussion. Should the top-level file testbench file live in replicant, or manycore? 

We can actually use the same top-level file for both SPMD loader and CUDA, it's just the C libraries that differ. We'd need a new SPMD loader that uses a C/C++ API but I don't see any major issues there.

### DIFF
--- a/testbenches/dpi/bsg_nonsynth_dpi_manycore.hpp
+++ b/testbenches/dpi/bsg_nonsynth_dpi_manycore.hpp
@@ -1,0 +1,193 @@
+// This header file defines a C++ API that wraps the System Verilog
+// DPI provided by bsg_nonsynth_dpi_manycore.v
+#ifndef __BSG_NONSYNTH_DPI_MANYCORE
+#define __BSG_NONSYNTH_DPI_MANYCORE
+#include <bsg_nonsynth_dpi.hpp>
+#include <bsg_nonsynth_dpi_fifo.hpp>
+#include <bsg_nonsynth_dpi_errno.hpp>
+#include <bsg_nonsynth_dpi_rom.hpp>
+#include <svdpi.h>
+#include <cstring>
+#include <cstdint>
+// We use __m128i so that we can pass a 128-bit type between Verilog
+// and C.
+#include <xmmintrin.h>
+
+// These are DPI functions provided by SystemVerilog compiler. If they
+// are not found at link time, compilation will fail. See the
+// corresponding function declarations in bsg_nonsynth_dpi_manycore.v
+// for additional information.
+extern "C" {
+        extern unsigned char bsg_dpi_credits_is_window();
+        extern unsigned char bsg_dpi_tx_is_vacant();
+        extern int bsg_dpi_credits_get_cur();
+        extern int bsg_dpi_credits_get_max();
+}
+
+namespace bsg_nonsynth_dpi{
+        /*
+         * dpi_manycore wraps an instantiation of
+         * bsg_nonsynth_dpi_manycore in a verilog design and provides
+         * C/C++-like functionality to the DPI interface.
+         *
+         * Template Parameters:
+         *   N: Number of ROM elements in the Manycore design
+         *
+         * Functions:
+         *   dpi_manycore: Constructor
+         *   get_credits: Get number of available transmit credits
+         *   tx_req: Transmit a request packet
+         *   rx_rsp: Receive a response packet
+         *   rx_req: Receive a request packet
+         */
+        template <unsigned int N>
+        class dpi_manycore : public dpi_base{
+                // DPI To Fifo (Request) Interface Object
+                dpi_to_fifo<__m128i> d2f_req;
+                // Fifo to DPI (Response) Interface Object
+                dpi_from_fifo<__m128i> f2d_rsp;
+                // Fifo to DPI (Request) Interface Object
+                dpi_from_fifo<__m128i> f2d_req;
+                // Maximum available credits (used for fences)
+                int max_credits = -1;
+                // Current available credits (used for flow control, and fences)
+                int cur_credits = 0;
+        public:
+                // Stores configuration data for the manycore DUT.
+                // Each entry is an unsigned 32-bit value
+                dpi_rom<unsigned int, N> config;
+
+                /**
+                 * Return an instance of dpi_manycore
+                 *
+                 * @param[in] hierarchy The path to the
+                 *   bsg_nonsynth_dpi_manycore instance in the design hierarchy
+                 *
+                 * @return a valid instance of dpi_manycore
+                 */
+                dpi_manycore(const std::string &hierarchy)
+                        : dpi_base(hierarchy),
+                          d2f_req(hierarchy + ".d2f_req_i"),
+                          f2d_rsp(hierarchy + ".f2d_rsp_i"),
+                          f2d_req(hierarchy + ".f2d_req_i"),
+                          config(hierarchy + ".rom")
+                {
+                        prev = svSetScope(scope);
+                        max_credits = bsg_dpi_credits_get_max();
+                        svSetScope(prev);
+                }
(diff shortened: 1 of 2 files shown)


drichmond:
I've added comments and documentation

testbenches/dpi/bsg_nonsynth_dpi_manycore.v
```diff
@@ -0,0 +1,293 @@
...
+    ,output [link_sif_width_lp-1:0] link_sif_o
+      
+    ,input [x_cord_width_p-1:0] my_x_i
+    ,input [y_cord_width_p-1:0] my_y_i
+      
+    ,output bit debug_o);
+
+   // Width of the FIFO interfaces, as determined by the padding in
+   // bsg_manycore_endpoint_to_fifos.v
+   localparam fifo_width_lp = 128;
```
mrutt92, line 37: is this hardcoded in `bsg_manycore_endpoint_to_fifos.v` as well?

mrutt92: if so you should put this variable in a package and import it.

drichmond: I think so. Good point

testbenches/dpi/bsg_nonsynth_dpi_manycore.v
```diff
@@ -0,0 +1,293 @@
...
+   
+   logic [`BSG_WIDTH(max_out_credits_p)-1:0] ep_out_credits_lo;
+   
+   // Host -> Manycore Requests
+   logic [fifo_width_lp-1:0] host_req_data_lo;
+   logic                     host_req_v_lo;
+   logic                     host_req_ready_li;
+
+   // Manycore -> Host Responses
```
mrutt92, line 46: bit of a style nit - I would declare these just above the modules that use them.

mrutt92: for improved readability.

drichmond: Is this in the BSG Style Guide?

testbenches/dpi/bsg_nonsynth_dpi_manycore.v
```diff
@@ -0,0 +1,293 @@
...
+
+      // manycore link
+      .link_sif_i      (link_sif_i),
+      .link_sif_o      (link_sif_o),
+      .my_x_i          (my_x_i),
+      .my_y_i          (my_y_i),
+      .out_credits_o   (ep_out_credits_lo)
+      );
+
+   bsg_nonsynth_dpi_rom
```
mrutt92, line 187: unused?

drichmond: It is

drichmond: It's purely DPI, so it has no IO

testbenches/dpi/bsg_nonsynth_dpi_manycore.v
mrutt92: I have no idea - as someone reviewing this code I fine myself scrolling back up and down quite a bit while reading it though.

testbenches/dpi/bsg_nonsynth_dpi_manycore.v
mrutt92: comment?

drichmond: Will do

testbenches/dpi/bsg_nonsynth_dpi_manycore.hpp
```diff
@@ -0,0 +1,175 @@
...
+// We use __m128i so that we can pass a 128-bit type between Verilog
+// and C.
+#include <xmmintrin.h>
+
+// These are DPI functions provided by SystemVerilog compiler. If they
+// are not found at link time, compilation will fail. See the
+// corresponding function declarations in bsg_nonsynth_dpi_manycore.v
+// for additional information.
+extern "C" {
+        extern unsigned char bsg_dpi_credits_is_window();
```
mrutt92, line 21: please add documenting comments here too about what these functions do.

mrutt92: actually - nm. your choice

testbenches/dpi/bsg_nonsynth_dpi_manycore.hpp
```diff
@@ -0,0 +1,175 @@
...
+                                res = get_credits(cur_credits);
+
+                        if(res != BSG_NONSYNTH_DPI_SUCCESS)
+                                return res;
+
+                        if(cur_credits == 0)
+                                return BSG_NONSYNTH_DPI_NO_CREDITS;
+
+                        if(cur_credits < 0)
+                                return BSG_NONSYNTH_DPI_INVALID;
```
mrutt92, line 131: is `BSG_NONSYNTH_DPI_INVALID` supposed so signify that the input is bad?

drichmond: No, it signals that cur_credits is an invalid value

mrutt92: In that case this is the wrong error code. Maybe you just want to return FAIL.

mrutt92: https://github.com/bespoke-silicon-group/basejump_stl/blob/eb8140c080f068f397142aea8e697e7f856037e6/bsg_test/bsg_nonsynth_dpi_errno.hpp#L55

mrutt92: Or maybe you want not valid. The string for that error code should be more specific e.g. "Output not valid"

mrutt92: Actually - what you *really* should do is print an error message saying "hey - the DPI went off the rails at this spot" so that if it ever does happen we know where in the code to look. Then return FAIL, because realistically there's not much to be done about it at runtime.

drichmond: Looking at the code, negative credits can't happen. If a credit error happens, it triggers a separate set of warnings in bsg_manycore_endpoint_standard.

testbenches/dpi/bsg_nonsynth_dpi_manycore.v
drichmond: Done

testbenches/dpi/bsg_nonsynth_dpi_manycore.v
drichmond: I switched this to a parameter. It is a parameter in bsg_manycore_endpoint_fifos.v, pulled from another package

testbenches/dpi/bsg_nonsynth_dpi_manycore.v
drichmond: Hesitant to fix. I understand it's your preference, but my preference is to declare all parameters and wires at the top of a file.

testbenches/dpi/bsg_nonsynth_dpi_manycore.hpp
drichmond: I've removed this check

testbenches/dpi/bsg_nonsynth_dpi_manycore.v
mrutt92: OK

PR #310, opened 2020-05-19
Moved files from BSG Replicant to BSG Manycore

drichmond:
I think these files would be better in BSG Manycore. 

bsg_print_stat_snoop.v: Only used in simulation, so I've moved it OUT of the hardware directory (which contains synthesizable hardware) and into bsg_manycore/testbenches/common/v

bsg_manycore_endpoint_to_fifos.v: This isn't F1 specific, and it's actually quite useful. I think it would be a good idea to have this in BSG Manycore in the long run.

The long-term benefit of this is that the hardware directory in bsg_replicant only contains files that are relevant to FPGA compilation

### DIFF
--- a/testbenches/common/v/bsg_print_stat_snoop.v
+++ b/testbenches/common/v/bsg_print_stat_snoop.v
@@ -1,0 +1,46 @@
+module bsg_print_stat_snoop
+  import bsg_manycore_pkg::*;
+  import bsg_manycore_addr_pkg::*;
+  #(parameter data_width_p="inv"
+    , parameter addr_width_p="inv"
+    , parameter x_cord_width_p="inv"
+    , parameter y_cord_width_p="inv"
+
+    , parameter link_sif_width_lp=
+      `bsg_manycore_link_sif_width(addr_width_p,data_width_p,x_cord_width_p,y_cord_width_p)
+    
+  )
+  (
+    // manycore side
+    input [link_sif_width_lp-1:0] loader_link_sif_in_i
+    , input [link_sif_width_lp-1:0] loader_link_sif_out_i
+
+    // snoop signals
+    , output logic print_stat_v_o
+    , output logic [data_width_p-1:0] print_stat_tag_o
+  );
+
+  `declare_bsg_manycore_link_sif_s(addr_width_p,data_width_p,x_cord_width_p,y_cord_width_p);
+
+  bsg_manycore_link_sif_s loader_link_sif_in;
+  bsg_manycore_link_sif_s loader_link_sif_out;
+
+  assign loader_link_sif_in = loader_link_sif_in_i;
+  assign loader_link_sif_out = loader_link_sif_out_i;
+
+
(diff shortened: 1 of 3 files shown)


tommydcjung:
Why are these file need to be moved if it's not used in bsg_manycore?

drichmond:
That's a great question, and I'm open to suggestions

1. I believe they'll be used in BSG Manycore soon (with Verilator)
2. I believe they're more general than BSG Replicant

I think that bsg_print_stat_snoop can be used here: https://github.com/bespoke-silicon-group/bsg_manycore/blob/98e6f6a1d4be56b50ce7e06dfc954f2576709b19/testbenches/common/v/bsg_nonsynth_manycore_monitor.v#L212

drichmond:
@tommydcjung, We already have a similar module in this repository (https://github.com/bespoke-silicon-group/bsg_manycore/blob/master/v/bsg_manycore_link_to_fifo.v)

This module is more up-to-date and tested

drichmond:
@taylor-bsg Do you have any thoughts on this?

taylor-bsg:
Sorry -- can you explain a little more the situation?

drichmond:
Yep!

I'd like to move these files from Replicant/F1 to Manycore, because they're not FPGA specific, not AXI-specific, and they're generally useful for trying to integrate the Manycore into other projects. My thought is that if these files are generally useful, then they should be in the more general repository (bsg_manycore) so that they can be long-term maintained.

For example, I use the FIFO-interface here for Verilator simulation, which is not FPGA-specific. I submitted this PR partly so that some day we can have a Verilator simulation that depends only on code within BSG Manycore (I did a poor job of explaining this)

It seems like a similar (though out of date?) file already exists (https://github.com/bespoke-silicon-group/bsg_manycore/blob/master/v/bsg_manycore_link_to_fifo.v).

I made this PR so that we can discuss the idea. If it doesn't get merged, it doesn't break anything. If it does get merged, it might help some later project down the road.

I would like to make a decision though. But it's not urgent

taylor-bsg:
Files that are only useful to the manycore should be in the manycore repo.
In some cases the module may be generally applicable to BaseJump STL (I.e.
generic NOC interface component) in which case they should be there.

It would be helpful to differentiate the module you want to commit with the
one already there.

drichmond:
Happy to move this to BaseJump too. It uses the Manycore endpoint standard though, which is in the Manycore repository.

Differentiation: 

[This module ](https://github.com/bespoke-silicon-group/bsg_manycore/blob/master/v/bsg_manycore_link_to_fifo.v) provides a network-to-FIFO interface for write packets that originate in the manycore. The number of network links is parametrizable, and a round robin arbiter selects one network link at a time to send out on the FIFO interface. The module only supports network-originating writes, since there is no read-response interface, and no host request interface. 

In contrast, the module in this PR (that Leonard wrote) provides a network-to-FIFOs interface for a single network link. Like the module above it translates packets from the network into FIFO flits. Unlike the module above, it differentiates between requests and responses. Unlike the module above it also provides FIFO interfaces for sending packets to the network.

The module in this PR serves a slightly different purpose than the original one in this repository, but they have similar functions.

PR #311, opened 2020-05-24
LLVM cross-compilation

vb000:
Neil built a great flow for compiling and running llvm, including a pass. This PR extends that with the following:
- Cross-compiling with generic LLVM build. So Manycore can work with system LLVM! (Provided, system LLVM is >=9.0.0)
- Ability to use C/C++ standard libraries. 
- Options for compiling with accurate ABI and ISA extensions.
- Option to enable/disable the pass.
- With C/C++ stdlib support, any program using `Makefile.builddefs` can be compiled with LLVM now: just run `make CLANG=1`.

Regression with LLVM: 
- Only three tests fail with Clang enabled: `bsg_tile_group_barrier`  `bypass_core` and `bsg_lr_acq`!!!
- May be an extra setting is required for correct atomic support, or could be a problem with codegen for atomics. Will look into it later, as the current priority is updating the cost model.

### DIFF
--- a/software/mk/Makefile.builddefs
+++ b/software/mk/Makefile.builddefs
@@ -187,7 +187,7 @@ RISCV_GXX_OPTS += $(RISCV_GXX_EXTRA_OPTS)
 spmd_defs = -DPREALLOCATE=0 -DHOST_DEBUG=0
 
 ifdef CLANG
-LLVM_DIR          ?= $(BSG_MANYCORE_DIR)/../bsg-llvm-project/build
+LLVM_DIR          ?= $(BSG_MANYCORE_DIR)/software/riscv-tools/llvm/llvm-install
 LLVM_CLANG        ?= $(LLVM_DIR)/bin/clang
 LLVM_CLANGPP      ?= $(LLVM_DIR)/bin/clang++
 LLVM_OPT          ?= $(LLVM_DIR)/bin/opt
--- a/software/mk/Makefile.builddefs
+++ b/software/mk/Makefile.builddefs
@@ -192,13 +192,12 @@ LLVM_CLANG        ?= $(LLVM_DIR)/bin/clang
 LLVM_CLANGPP      ?= $(LLVM_DIR)/bin/clang++
 LLVM_OPT          ?= $(LLVM_DIR)/bin/opt
 LLVM_LLC          ?= $(LLVM_DIR)/bin/llc
-RUNTIME_FNS       ?= $(BSG_MANYCORE_DIR)/software/bsg_manycore_lib/bsg_tilegroup.h
 CLANG_TARGET_OPTS ?= --target=riscv32 -march=$(ARCH_OP) -mabi=$(ABI)
-
 ifdef LLVM_PASS
-PASS_DIR         ?= $(BSG_MANYCORE_DIR)/software/manycore-llvm-pass
-PASS_LIB         ?= build/manycore/libManycorePass.so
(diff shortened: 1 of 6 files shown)


eafurst:
awesome, this looks great - thanks Bandhav!!

software/mk/Makefile.builddefs
```diff
@@ -186,13 +187,17 @@ RISCV_GXX_OPTS += $(RISCV_GXX_EXTRA_OPTS)
...
-PASS_LIB         ?= build/manycore/libManycorePass.so
-RUNTIME_FNS      ?= $(BSG_MANYCORE_DIR)/software/bsg_manycore_lib/bsg_tilegroup.h
+LLVM_DIR          ?= $(BSG_MANYCORE_DIR)/software/riscv-tools/llvm/llvm-install
+LLVM_CLANG        ?= $(LLVM_DIR)/bin/clang
+LLVM_CLANGPP      ?= $(LLVM_DIR)/bin/clang++
+LLVM_OPT          ?= $(LLVM_DIR)/bin/opt
+LLVM_LLC          ?= $(LLVM_DIR)/bin/llc
+CLANG_TARGET_OPTS ?= --target=riscv32 -march=$(ARCH_OP) -mabi=$(ABI)
+ifdef LLVM_PASS
+PASS_DIR          ?= $(BSG_MANYCORE_DIR)/software/manycore-llvm-pass
```
drichmond, line 197: Is LLVM_PASS documented anywhere?

Related: Will we support it in the future?

vb000: Not sure about this pass, but I imagine we could be adding passes. 

Yes, without documentation, this is more of easter egg than a feature. What do you think is nice place to describe these?

drichmond: In a README. Perhaps in the software directory. 

If we're not planning to use the old LLVM pass in the future, I think we should remove it. However, it does serve as a useful example of an LLVM Pass, until you and Emily add more. 

My suggestion is rename it to something less generic than LLVM_PASS, LLVM_SHARED_MEM_PASS, and document it.

As a general thought though, if we plan to have multiple passes I think that Makefile variables aren't a good way to specify which should/shouldn't be run. It's difficult to specify what order the passes are run in. Is there another way we can specify the passes we want to run, and what order? Ideally, this is per-program

vb000: Readme in software directory sounds good!

I agree with you regarding passes. The option I added feels more like enable/disable all passes. With our current understanding though, we don't think we could customize scheduler with out-of-source passes. So, let's defer this per-pass enable/disable problem?

drichmond: Yes, file an issue, write it in your list of someday items, or ...

Oh, and change the flag name. Make it reflect what it actually does

vb000: Great! Yes, I'm about to change the flag name to `ENABLE_LLVM_PASSES`!

vb000: Added [readme](https://github.com/bespoke-silicon-group/bsg_manycore/blob/clang_cpp/software/README.md).

drichmond:
Seems good to me.

drichmond:
Can you run the build-tools portion of CI on this?

vb000:
Done... https://github.com/bespoke-silicon-group/bsg_manycore/commits/ci_clang_cpp.

PR #323, opened 2020-06-07
Fixed the bug that the address does not increase in read burst

hwpeng:


taylor-bsg:
Why was this bug not exposed earlier?

tommydcjung:
The number of bursts greater than 1 hasn't been used before.

taylor-bsg:
Okay thanks!

PR #324, opened 2020-06-08
Adds compiler flags to improve loop unrolling performance.

mrutt92:


mrutt92:
I suppose it is - it's added here because I noticed that it was implied by `-funroll-loops`

PR #326, opened 2020-06-15
modify llvm install script to install our version of llvm

eafurst:
- modified the existing llvm install build file to instead clone our fork of llvm and build with support for our hb specific backend
- our llvm backend is still under development, but I retained the last line of the makefile that deletes the source and build directories - thoughts on whether we should keep or remove that line?

### DIFF
--- a/software/mk/Makefile.llvminstall
+++ b/software/mk/Makefile.llvminstall
@@ -9,7 +9,8 @@ llvm-install:
 	mkdir -p $(LLVM_DIR)/llvm-build && mkdir -p $(LLVM_DIR)/llvm-install
 	# Get LLVM sources
 	cd $(LLVM_DIR) && \
-    git clone https://github.com/bespoke-silicon-group/llvm-project.git ./llvm-src
+    git clone https://github.com/bespoke-silicon-group/llvm-project.git ./llvm-src && \
+    cd ./llvm-src && git fetch && git checkout hb-dev
 	# Install only X86 and RISCV targets
 	cd $(LLVM_DIR)/llvm-build \
 	    && cmake3 -DCMAKE_BUILD_TYPE="Debug" \
@@ -23,4 +24,3 @@ llvm-install:
 	    -DLLVM_OPTIMIZED_TABLEGEN=True \
 	    ../llvm-src/llvm
 	cd  $(LLVM_DIR)/llvm-build && cmake3 --build . -- -j12 && make install
-	rm -rf $(LLVM_DIR)/llvm-build $(LLVM_DIR)/llvm-src
--- a/software/mk/Makefile.llvminstall
+++ b/software/mk/Makefile.llvminstall
@@ -2,10 +2,6 @@ ifndef LLVM_DIR
     $(error Need to define LLVM_DIR)
 endif
 
-ifndef RISCV_INSTALL_DIR
-    $(error Need to define RISCV_INSTALL_DIR)
-endif
-
 # devtoolset-8
 HOST_TOOLCHAIN ?= /opt/rh/devtoolset-8/root/usr/bin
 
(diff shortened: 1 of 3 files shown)


vb000:
Cool, does this checkout `hb-dev` branch by default?

Just checked the size of build: it's 16G, so not sure we should keep the build. We could certainly provide an option to keep builds, using this target (this currently doesn't keep llvm builds because the llvm makefile removes them):
https://github.com/bespoke-silicon-group/bsg_manycore/blob/3bea19e18ff9452dd92917a7d930ff76dd28d404/software/riscv-tools/Makefile#L172-L174

Also, you could use CI to check if this builds correctly as a fresh install by creating a branch with the prefix `ci_`. In this case, it'd be `ci_llvm-update`.

eafurst:
> Cool, does this checkout `hb-dev` branch by default?

Yes! `hb-dev` is the default branch on our fork, so it is automatically checked out. However, if we think that might change in the future, I can add a line that checks out the branch to be safe.

> Just checked the size of build: it's 16G, so not sure we should keep the build. We could certainly provide an option to keep builds, using this target (this currently doesn't keep llvm builds because the llvm makefile removes them):
> https://github.com/bespoke-silicon-group/bsg_manycore/blob/3bea19e18ff9452dd92917a7d930ff76dd28d404/software/riscv-tools/Makefile#L172-L174

Oh cool, is `install-clean` the rule that is used by default when building the tools? In that case, I could just remove that line from this makefile?

> Also, you could use CI to check if this builds correctly as a fresh install by creating a branch with the prefix `ci_`. In this case, it'd be `ci_llvm-update`.

I haven't used CI before - I created and pushed the branch, but is there anything else I need to do on my end?

vb000:
> Yes! `hb-dev` is the default branch on our fork, so it is automatically checked out. However, if we think that might change in the future, I can add a line that checks out the branch to be safe.

I think it's safe to be explicit here because at a glance people can assume we are checking out master.

> Oh cool, is `install-clean` the rule that is used by default when building the tools? In that case, I could just remove that line from this makefile?

Yes. Usage goes here: https://github.com/bespoke-silicon-group/bsg_manycore/blob/3bea19e18ff9452dd92917a7d930ff76dd28d404/software/riscv-tools/README.md. So, you could explain any changes you'd like to make to the interface here.

> I haven't used CI before - I created and pushed the branch, but is there anything else I need to do on my end?

Hopefully nothing. You could monitor the run by clicking "Details" button above.

vb000:
Sorry, I missed to mention. `clean-builds` rule also has to updated to remove llvm build when running `install-clean` target.

https://github.com/bespoke-silicon-group/bsg_manycore/blob/3bea19e18ff9452dd92917a7d930ff76dd28d404/software/riscv-tools/Makefile#L180-L182

vb000:
Also,  to start CI immediately, you could manually sync with CI repo here: https://gitlab.com/dpetrisko/bsg_manycore/-/branches. Click `Update Now` button after signing in. Signing in with your github account would allow you access I think.

eafurst:
Hmm, I don't see an `Update Now` button, but I've requested access to the repo which will hopefully fix that.

dpetrisko:
The manycore CI is on my account for...some reason :) gave you access, should be able to trigger jobs now

eafurst:
Yeah, I was trying to look at the beebs test failure and couldn't figure out how these changes would have affected it, but I also am not familiar with those tests!

vb000:
@drichmond let's merge this? Any thoughts?

drichmond:
If the CI is failing, shouldn't we figure out why?

vb000:
Not super familiar with beebs ci script, and it failing seems to be not related to these changes. I've observed rerunning seems to makes it run successfully, sometimes. @tommydcjung what do you think might be the issue?

vb000:
Doesn't seem to be beebs related as well. It's evident, but confirmed by a local run. May be we could increase retries here? https://github.com/bespoke-silicon-group/bsg_manycore/blob/6a5c703f4c081a54032e2bff2944066e95bd01d7/.gitlab-ci.yml#L162

vb000:
Somehow rerunning seems to fix it, like in this case.

PR #329, opened 2020-06-19
Testcase to compare NANs

vb000:
Checks NAN comparison compliance by disabling `-ffast-math` optimizations.


drichmond:
Rerunning with CI fixes

PR #332, opened 2020-06-23
Change BEEBS repeat factor to 1

vb000:
We currently use beebs as a test rather than a benchmark. So, reduced the repeat factor to just 1 so that we don't have to wait disproportionately long for beebs to finish.

### DIFF
--- a/software/spmd/beebs/Makefile
+++ b/software/spmd/beebs/Makefile
@@ -19,7 +19,7 @@ include $(BSG_MANYCORE_DIR)/software/spmd/beebs/Makefile.bmarklist
 bsg_tiles_X = 1
 bsg_tiles_Y = 1
 
-REPEAT_FACTOR = 2 # Number of iterations each benchmark is run
+REPEAT_FACTOR = 1 # Number of iterations each benchmark is run
 
 BEEBS_DIR := $(BSG_MANYCORE_DIR)/imports/beebs
 RUN_DIR := $(BSG_MANYCORE_DIR)/software/spmd/beebs


tommydcjung:
I think there is a benefit to having it repeat twice, so that it might exercise the icache misses.

vb000:
I think, if we want to test icache, we could have more directed tests. Running beebs twice to test icache incurs huge price in time feels like a convoluted way to test icache tbh.

vb000:
It's reasonable to argue that we test with real programs though. I just feel beebs is particularly bad to use it as a testcase. `nbody` basically spins in this loop for 100 times: https://github.com/bespoke-silicon-group/beebs/blob/049ded9f3aeb5591f553879d3a0376b8614e9422/src/nbody/nbody.c#L176

Also, the loop is also not large enough to spill over icache...

tommydcjung:
Compared to having a bug in the hardware and not knowing it, it's a relatively small price to pay.

taylor-bsg:
Bandhav, perhaps you can provide more stats about this runtime? How many
cores for how many hours?

M

vb000:
On kk9 with 12 cores:
- beebs: 23m54s
- rest of the regression: 1m26s

As is evident, beebs runs disproportionately long and if we go through the source, the longest running beebs benchmark spins 100 x 2 (repeat factor) in a single loop. Making the repeat factor 1 still tests with 100 iterations. This addresses Tommy's concern?

vb000:
If this too controversial and people are against it, I can close this. But I think, this change would make the regression efficient and would be a step towards more agile development.

dpetrisko:
A compromise may be to run shorter locally and longer in CI?

vb000:
> A compromise may be to run shorter locally and longer in CI?

Hm, I'm more trying to say, running beebs twice vs once doesn't provide any novel benefits in terms of testing.

taylor-bsg:
Can we reduce the # of iterations in nbody? Where is the pointer to the
top-level code that calls out to the nbody benchmark function?


M

vb000:
> Can we reduce the # of iterations in nbody? 

This involves a patch to beebs repo. I didn't want to go down this route because, this essentially means running a different program. `REPEAT FACTOR` on the other hand is formal interface provided by beebs.

> Where is the pointer to the top-level code that calls out to the nbody benchmark function?

This: https://github.com/bespoke-silicon-group/beebs/blob/049ded9f3aeb5591f553879d3a0376b8614e9422/support/main.c#L44-l48 

calls
 
this: https://github.com/bespoke-silicon-group/beebs/blob/049ded9f3aeb5591f553879d3a0376b8614e9422/src/nbody/nbody.c#L171-L180

taylor-bsg:
Okay I see .. the argument on one side is that executing code twice does
have very different behavior from an icache perspective. However looking at
this benchmark, I agree that except for the REPEAT FACTOR loop itself that
is not happen, because the loop it calls iterates many times. So then the
question is, are the other benchmarks all like this?

vb000:
Some are as obvious as this, some run the same code but with different data (different data only in the inner loop, exact same benchmark in the outer loop).

These are examples of exactly same and different data, respectively:

https://github.com/bespoke-silicon-group/beebs/blob/049ded9f3aeb5591f553879d3a0376b8614e9422/src/fasta/libfasta.c

and 

https://github.com/bespoke-silicon-group/beebs/blob/049ded9f3aeb5591f553879d3a0376b8614e9422/src/crc32/crc_32.c

drichmond:
Can’t we run CI tests selectively, depending on what changed in the PR?

For, if the compilation follow, or the RTL changes, we would rerun longer tests like this. But for unrelated PRs we just run the shorter tests

Just a thought

taylor-bsg:
How many of the beebs have excessive runtime? Maybe we can just adjust a small subset? Beebs is currently our primary source of entropy for testing, so I hate to reduce what little we have blindly.

taylor-bsg:
Also we can run on kk8, our CI/parallel run machine and use more than 12 threads (use nice command when you do it.)

vb000:
I think the icache argument would be valid for excluding the small subset as well. My argument is that majority portion of a beebs benchmark is a loop and I think, running twice only provides us only small value. Beebs is great for testing floating point, and it's proven to be the case, but icache, I'm not really confident beebs would do as good a job as a large program. Entropy argument makes sense though, since we don't have other real programs. When we find a large program though, running it for few iterations would be a better choice than running beebs twice (times however many in the inner loop).

PR #336, opened 2020-06-26
Hardfloat profiler update

bornaehsani:
Updating profiler tools in accordance with the Hardfloat merge.
Updated stall/instr/miss types in stats parser and blood graph
Minor bug fixes and changes

Generates bloodgraph, vcache stall graph, stats, vcache stats, and pc histogram in Cosim on machines 4x4_fast_n_fake, 4x4_blocking_vcache_f1_model, and timing_v0_8_4.


software/py/vanilla_parser/stats_parser.py
```diff
@@ -437,7 +437,7 @@ def __print_manycore_tag_stats_tile_group_timing(self, stat_file, tag):
                                      ,(self.manycore_stat[tag]["global_ctr"])
                                      ,(self.manycore_stat[tag]["cycle_parallel_cnt"])
                                      ,(self.manycore_stat[tag]["instr_total"] / self.manycore_stat[tag]["global_ctr"])
-                                     ,(100 * self.manycore_stat[tag]["instr_total"] / self.manycore_stat[tag]["instr_total"])
+                                     ,(np.float64(100 * self.manycore_stat[tag]["instr_total"]) / self.manycore_stat[tag]["instr_total"])
```
drichmond, line 440: Is there ever a situation where we have zero instructions?  Why cast this to a double?

drichmond:
Works for my code. @vb000 Can you review, too?

software/py/vanilla_parser/stats_parser.py
```diff
@@ -215,10 +215,10 @@ def __init__(self, per_tile_stat, per_tile_group_stat, per_vcache_stat, vanilla_
         self.stats, self.instrs, self.misses, self.stalls, self.bubbles = self.parse_header(vanilla_input_file)
 
 
-        # bubble_fp_op is a bubble in the Integer pipeline "caused" by
+        # bubbles that are  a bubble in the Integer pipeline "caused" by
         # an FP instruction executing. Don't count it in the bubbles
         # because the procesor is still doing "useful work". 
-        self.notbubbles = ['bubble_fp_op'] 
+        self.notbubbles = []
```
vb000, line 221: Just curious, is this changed fp ops won't create bubbles in integer pipeline anymore?

software/py/vanilla_parser/stats_parser.py
bornaehsani: Normally no. But I've been burned by thinking exactly like this many times, if you forget to put print_stats in your code, it's better to generate a stat file with all zeros instead of barking at you with an error, don't you think?

software/py/vanilla_parser/stats_parser.py
bornaehsani: No, originally bubble_fp_op was just a bubble in the Integer pipeline but the FP pipeline was busy, so we didn't use to count it as bubbles. But in the new version bubble_fp_op is eliminated all together, so no need to have it in the parser anymore.

PR #337, opened 2020-06-26
Fixed Profilers for Verilator

drichmond:
Verilator added support for $root, so I took another stab at this. This is the first (and most important?) of the profilers. 

In general: 

- Replaced forever with `always @(negedge clk_i)`
- Replaced `#1` with `always @(my_x_i)` on the assumption that my_x_i and my_y_i only change once
- Fixed an issue where stat_r had both blocking and non-blocking assignments. 

I'm least comfortable with replacing `#1` with `always @(my_x_i)`. It seems like we have to guarantee that `my_y_i` has to change before `my_x_i. @tommydcjung Do you have any ideas here?

Also, apologies. My text editor "fixed" indentation and inserted whitespace. Turn on ignore whitespace changes to see the 30-ish lines I had to change.

### DIFF
--- a/testbenches/common/v/vanilla_core_profiler.v
+++ b/testbenches/common/v/vanilla_core_profiler.v
@@ -616,11 +616,11 @@ module vanilla_core_profiler
 
   always_ff @ (posedge clk_i) begin
     if (reset_i) begin
-      stat_r <= '0;
+         stat_r = '0;
     end
     else begin
       stat_r.cycle++;
-      stat_r.instr <= stat_r.instr + instr_inc + fp_instr_inc;
+         stat_r.instr = stat_r.instr + instr_inc + fp_instr_inc;
 
       if (stall_all) begin
         if (stall_remote_ld_wb) stat_r.stall_remote_ld_wb++;
@@ -770,14 +770,15 @@ module vanilla_core_profiler
 
   integer fd, fd2;
   string header;
+   initial begin
+      fd = $fopen(logfile_lp, "w");
+      $fwrite(fd,"");
+   end
 
-  initial begin
-  
-    #1; // we need to wait for one time unit so that my_x_i and my_y_i becomes a known value.
-
+   always @(negedge reset_i) begin      
     // the origin tile opens the logfile and writes the csv header.
     if ((my_x_i == x_cord_width_p'(origin_x_cord_p)) & (my_y_i == y_cord_width_p'(origin_y_cord_p))) begin
-      fd = $fopen(logfile_lp, "w");
+      fd = $fopen(logfile_lp, "a");
       $fwrite(fd, "time,");
       $fwrite(fd, "x,");
       $fwrite(fd, "y,");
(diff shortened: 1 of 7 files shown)


drichmond:
Would using @(negedge reset_i) be better?

drichmond:
My editor fixed whitespace. You can turn on `ignore whitespace` while I figure it out.

dpetrisko:
image

Took me a bit to find it, but good to know this is a feature!

drichmond:
OK, fixed the whitespace.

drichmond:
OK, I went ahead and fixed all the profilers. They all pass compilation in VCS in replicant, and they (seemingly) generate valid output for the tools.

What do I need to do to confirm that these are working properly, and get this PR merged?

taylor-bsg:
very cool!

drichmond:
@tommydcjung Do you want to review this too?

testbenches/common/v/vanilla_core_profiler.v
```diff
@@ -764,14 +764,15 @@ module vanilla_core_profiler
...
+   initial begin
+      fd = $fopen(logfile_lp, "w");
+      $fwrite(fd,"");
+   end
 
-  initial begin
-  
-    #1; // we need to wait for one time unit so that my_x_i and my_y_i becomes a known value.
-
+   always @(negedge reset_i) begin
```
tommydcjung, line 772: This reset_i is local to each tile, and tiles can come out of reset in different order. Which means that the origin tile might not be the first to write the headers.

drichmond: Thats a great point. Does it matter here though? The header is only written by the origin tile because of the if statement. 

It does raise another issue though - could the header be written after the first line of data? (perhaps this is what you meant?)

drichmond: Alternatively, we could drive the profiler's reset from the top level

drichmond: @tommydcjung Is this the only thing that needs to be addressed?

tommydcjung: Obviously some numbers will change if the reset comes from the top-level. If that's okay, it might be an acceptable solution

drichmond: OK, let's discuss this in the manycore meeting tomorrow.

taylor-bsg:
Just to clarify, all tiles do come out of reset on the same cycle, i
believe?

M

drichmond:
Regardless of whether or not they do, I don't believe it affects correctness. Only the origin tile prints the header after reset is de-asserted

Let's discuss in the meeting today.

drichmond:
I'd like to return to this. When we last discussed this we were debating whether reset was asserted simultaneously for all cores, and if this is an issue, how to handle it.

What needs to be addressed? I don't believe depending on reset is an issue - only the origin tile will write the header files for the profiler, and my_x_i/my_y_i will be set when reset de-asserts. 

I've also addressed the whitespace issues. This has been rebased on master (and tested)

taylor-bsg:
I think we resolved this, and would also like to re-establish the
expectation that everything within the same clock domain on our chip does
come out of reset at the same time. =)

M

PR #339, opened 2020-06-30
DPI interface for profiler

drichmond:
This DPI interface allows a simulator to read floating point instructions executed, integer arithmetic instructions executed, and total number of instructions executed. 

It works in Verilator and VCS. A corresponding PR (and testbench) will be in bsg_replicant.


drichmond:
This PR is independent of #337 (but also works with Verilator, when rebased)

dpetrisko:
Looks good to me in general.  Just to make sure I understand the use case here: this is only for non-synthesizable counters that are implemented via bind in VCS/Verilator? So this won't be expected to be implemented in AWS/BP/physical hardware

drichmond:
Yes, that's correct.

We COULD implement physical counters in the future. The API in replicant is designed to support that, but not intended at the moment.

PR #346, opened 2020-07-06
Added DPI interface for vanilla_operation_trace.csv

drichmond:
New interface (which is backward compatible, `+trace` still works) allows users to turn tracing on/off from C/C++

For small kernels, this results in a 50% smaller trace file, and at LEAST 94% smaller trace file for kernels with large datasets because the API can eliminate tracing during data transfer. (Or, they can include it. But the user gets to decide now).

I've also added a regression test kernel that performs vector add, and demonstrates the C/C++ API.

### DIFF
--- a/software/spmd/bsg_cuda_lite_runtime/tracer/Makefile
+++ b/software/spmd/bsg_cuda_lite_runtime/tracer/Makefile
@@ -1,0 +1,34 @@
+#########################################################
+# Network Configutaion
+# If not configured, Will use default Values
+	bsg_global_X ?= $(bsg_tiles_X)
+	bsg_global_Y ?= $(bsg_tiles_Y)+1
+
+#########################################################
+#Tile group configuration
+# If not configured, Will use default Values
+	bsg_tiles_org_X ?= 0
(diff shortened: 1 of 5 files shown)


tommydcjung:
Can't the dpi modify trace_en_i coming from top-level?

drichmond:
I thought about that. 

To clarify your question though - Are you suggesting that the DPI modify the signal at the top level? Or drive it internally in themodule.

tommydcjung:
at the top-level

drichmond:
Gotcha, 

I can do that, there is no technical reason I can't. 

I specifically chose this way because I wanted to provide an ability to trace a subset of tiles in the future.

Furthremore, the the DPI interface (and associated class) was already written for this module. This was the simplest path to adding this feature.

One drawback of this approach (as already hinted in my description) is that it doesn't work for vcache tracing. I would need another DPI interface for that.

tommydcjung:
Sure you can also have an array of trace_en_i in the top-level.

drichmond:
That's reasonable. I'm coming around to that approach because I could write a general DPI-GPIO module

PR #347, opened 2020-07-07
Small modification to naming and indexing to make hierarchy path more similar to mesh network

drichmond:
This small change vastly simplifies processing for the DPI hierarchy in C/C++

The remaining differences can be handled in C/C++

### DIFF
--- a/testbenches/common/v/bsg_manycore_top_crossbar.v
+++ b/testbenches/common/v/bsg_manycore_top_crossbar.v
@@ -149,10 +149,9 @@ module bsg_manycore_top_crossbar
     assign link_in[1][i] = io_link_sif_i[i];
   end
 
-
   // Instantiate Vanilla Cores
-  for (genvar i = 2; i <= num_tiles_y_p; i++) begin: py
-    for (genvar j = 0; j < num_tiles_x_p; j++) begin: px
+  for (genvar i = 1; i < num_tiles_y_p; i++) begin: y
+    for (genvar j = 0; j < num_tiles_x_p; j++) begin: x
       bsg_manycore_proc_vanilla #(
         .x_cord_width_p(x_cord_width_lp)
         ,.y_cord_width_p(y_cord_width_lp)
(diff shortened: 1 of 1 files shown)


drichmond:
This passes CI, can we merge it?

tommydcjung:
I think this looks okay, but I don't think there should be a requirement that the hierarchy should be  "similar"...

drichmond:
There isn't. I fixed other hierarchy differences in C/C++. This is purely nice-to-have for replicant, since it reduces the need for a bunch of other machine-dependent strings. 

It's a small change, and doesn't affect anyone else's code.

drichmond:
Can you clarify? Where would the static variable live? What would the DPI calls return?

drichmond:
@taylor-bsg Can you be the ultimate arbiter here?

taylor-bsg:
Hi,

I see validity in both sides — here are my thoughts:

This issue also comes up when writing backend placement scripts for code
that uses generate statements. When there are arbitrary name differences
based on what generate path was taken, the code complexity explodes for the
backend scripts. Similarly here, there would be an increase in the size and
complexity of the test bench code.

The broader term for this kind of parallel structure between test bench or
backend code and the verilog is “aspect oriented programming.” The key idea
is separation of concerns. However, aspect oriented programming always has
the issue that there will always be this coupling between the aspect
program and the main code. For this reason, we need to think of these hooks
(ie the name of generate blocks, array names, etc) as an interface which is
worth of standardization. General software engineering interfaces do argue
that we should strive for more uniformity in interfaces when we want to
connect a module to two other modules that are similar in their abstracted
functionality.

So as long as it does not complicate the verilog code significantly, I do
think it’s a good idea to have more uniformity in this interface. If it
made the verilog much more complicated at the expense of making the
testbench less complicated then I would not be in favor, since I would
rather have bugs in my testbench than in my silicon. Additionally, we
should have comments in the Verilog that document what the expectations are
of the interfacing code and have a link to the files that are being
interfaced.


M

drichmond:
I've added comments in the bsg_replicant HDL and library files to explain how the DPI functions find and set scope, and what to do if the naming hierarchy changes.

I've also added a check in the dpi_base class in BaseJump STL that fails early in simulation, and provides a more helpful failure message than what VCS or Verilator provide, when a scope is not found.

Running regression on replicant now

PR #348, opened 2020-07-07
Link to cache crit path

tommydcjung:


taylor-bsg:
What is the gist of the change? You pulled invariant code out of a case statement?

tommydcjung:
That's right.

PR #350, opened 2020-07-10
Added tracer regression test (for Replicant)

drichmond:
This test performs parallel vector add

We use this to demonstrate the tracers in replicant.


drichmond:
The purpose of this kernel is to test the trace and log generation, not the profiler (that's the purpose of test_profiler)

PR #360, opened 2020-07-15
Fixed undriven reset_r index

drichmond:
* Different parts of hierarchy were coming out of reset at different times
* 0'th bit of reset delay was not driven

### DIFF
--- a/v/bsg_manycore.v
+++ b/v/bsg_manycore.v
@@ -103,11 +103,11 @@ module bsg_manycore
   logic [reset_depth_p-1:0][num_tiles_y_p-1:0][num_tiles_x_p-1:0] reset_i_r;
 
   genvar k;
-  for (k = 1; k < reset_depth_p; k++)
+  for (k = 0; k < reset_depth_p; k++)
     begin
       always_ff @(posedge clk_i)
         begin
-          reset_i_r[k] <= (k == 1) ? {(num_tiles_y_p*num_tiles_x_p){reset_i}} : reset_i_r[k-1];
+          reset_i_r[k] <= (k == 0) ? {(num_tiles_y_p*num_tiles_x_p){reset_i}} : reset_i_r[k-1];
         end
    end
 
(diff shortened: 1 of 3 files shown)


drichmond:
After discussion with @tommydcjung I've updated this PR to drive reset_r[0] in both hierarchies

v/bsg_manycore.v
```diff
@@ -103,11 +103,11 @@ module bsg_manycore
...
   genvar k;
-  for (k = 1; k < reset_depth_p; k++)
+  for (k = 0; k < reset_depth_p; k++)
     begin
       always_ff @(posedge clk_i)
         begin
-          reset_i_r[k] <= (k == 1) ? {(num_tiles_y_p*num_tiles_x_p){reset_i}} : reset_i_r[k-1];
+          reset_i_r[k] <= (k == 0) ? {(num_tiles_y_p*num_tiles_x_p){reset_i}} : reset_i_r[k-1];
         end
    end
```
tommydcjung, line 112: Can we move the for-loop inside the always_ff and use integer instead of genvar?

taylor-bsg: I think we need more explanation of why we are changing the manycore reset latency, and why this would not make it inconsistent with the rest of the SoCs.

Generally we prefer using generate statements over behavioral verilog.

drichmond:
I like how you used dff_chain

taylor-bsg:
Be sure to test on verilator; this code is the way it is because of an
issue with verilator in the original:

https://github.com/bespoke-silicon-group/bsg_manycore/commit/b996bf359ab4fdf89ec4da27bfe542cbb31aecfb

drichmond:
Yes, and no. I had `reset_depth_lp` as 3 in the verilator top level, which exposed the issue. It wasn't an issue with Verilator itself, per-se.

FWIW, I did test this with Verilator and VCS today. Both passed.

tommydcjung:
Can I merge?

taylor-bsg:
Merge, If you guys are happy with it...

M

PR #361, opened 2020-07-16
Fix parameter setting indexing for Verilator

drichmond:
Verilator doesn't like when parameters are re-assigned -- for example when initialized to a value, and then a subset are re-written. 

This occurs in the crossbar. I've fixed it.


tommydcjung:
Is this a bug in verilator?

drichmond:
Possibly. I can add it to my list of Todos and file an issue there to find out.

PR #364, opened 2020-07-24
barrier perf test [bug fix]

tommydcjung:
Running on 16x8, the barrier gets stuck and the program hangs (after the first barrier).


tommydcjung:
The bug was that the done_list was being cleared after all the tiles are woken up. The tiles wake up and enters the 2nd barrier, sending 1s to the middle tiles. But the middle tile hasn't finished clearing the done list...

taylor-bsg:
Borna, have you fixed this bug in the C++ version? Perhaps Tommy, you can also review that one?

bornaehsani:
Yes we came against this bug when I was developing the C++ version, you can see in the link below that call to alert() occurs after the call to reset() that clears the done list for the row and column barriers.
https://github.com/bespoke-silicon-group/bsg_manycore/blob/bbfd528de7e36dc64db6cf427f30609a80483c98/software/bsg_manycore_lib/bsg_tile_group_barrier.hpp#L273

PR #369, opened 2020-07-27
Switch to bsg riscv-gnu-toolchain

drichmond:
We've decided to fork our own version of the riscv-gnu-toolchain repository. 

In this PR I did three things: 

First, I switched to our fork: https://github.com/bespoke-silicon-group/riscv-gnu-toolchain

Second, I switched to our branch: https://github.com/bespoke-silicon-group/riscv-gnu-toolchain/tree/bsg_custom_git_modules, which points to our riscv-newlib fork.

Third, I removed newlib.patch by applying the patched changes to the branch above. 

@mrutt92 - Your fork of gcc can be brought in by making a PR to the branch above, and then updating the commit in the Makefile here.


drichmond:
We can bump the submodules however we like, eventually. One change at a time and this PR keeps the toolchain commits the same (though the build process is slightly different because i removed the patches)

taylor-bsg:
Looks good to me! So will this mean that we are building newlib and all the software using our private gcc?

drichmond:
Yes. For newlib, we were before but we had a patch to do it manually. This is a more stable way. For GCC, that is in #375

drichmond:
And Yes, I think this is ready to merge. Nothing about the toolchain changes in this PR.

drichmond:
Just how we get it

PR #373, opened 2020-07-31
Move pipeline reset dff outside

tommydcjung:


tommydcjung:
More flexible to have this outside for physical design, if the adapter were to be placed inside a placement boundary.

drichmond:
Do we still want this?

PR #376, opened 2020-08-03
Removed .* matching from .dram section in link gen

drichmond:
The new linker script generator globs everything left during linking, into .dram. This includes debug symbols. 

This PR fixes that, by removing .* from .dram. This re-enables the ability to print in-line sources during disassembly. 

With the current solution the debug symbols are ALL loaded into the simulation during execution. This signficantly slows down runtime for large kernels when debug symbols are compiled.

If there's a better fix, I'm all ears. This is a high-priority needs-to-be-fixed for me.


bornaehsani:
Can we please have this merged in ASAP? It would really help me! Thanks!

drichmond:
@vb000 Can you think of any issues this would cause?

drichmond:
No worries! I want it to make it in to the next release. It's dead useful.

PR #378, opened 2020-08-05
Added a comment that describes barrier initialization

drichmond:
I've added a comment that I think is worth documenting. 

This is something that scared me. I was worried that we were not initializing the barriers before launching a tile. Thankfully, Max knows his C/C++ and ELF files and handled this last year.

However, the process of confirming this comment took a few days, and a meeting with Max. I've added a comment to avoid anyone performing the same traversal.

### DIFF
--- a/software/bsg_manycore_lib/bsg_tile_group_barrier.hpp
+++ b/software/bsg_manycore_lib/bsg_tile_group_barrier.hpp
@@ -40,6 +40,22 @@ template <int BARRIER_X_DIM>
 class bsg_row_barrier {
 private:
 
+        // The initialization assignments here have no effect if the
+        // barrier is instantiated at global scope (as is recommended
+        // practice)
+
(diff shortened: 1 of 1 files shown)


software/bsg_manycore_lib/bsg_tile_group_barrier.hpp
```diff
@@ -40,6 +40,22 @@ template <int BARRIER_X_DIM>
 class bsg_row_barrier {
 private:
 
+        // The initialization assignments here have no effect if the
+        // barrier is instantiated at global scope (as is recommended
+        // practice)
+
+        // When the barrier is declared at global scope _local_alert
+        // and _done_list are located in the .bss.dmem segment.
```
mrutt92, line 48: Section not segment. Two different things. Segments don’t have names.

drichmond: fx'd

PR #382, opened 2020-08-07
Added bsg prefixes to non-blocking load related macros

vb000:
Update non-blocking loads related macros with bsg prefixes:

- `__remote` --> `bsg_attr_remote`
- `NOALIAS` --> `bsg_attr_noalias`
- `UNROLL` --> `bsg_unroll`

Tested on pytorch regression.

### DIFF
--- a/software/bsg_manycore_lib/bsg_manycore.h
+++ b/software/bsg_manycore_lib/bsg_manycore.h
@@ -147,12 +147,13 @@ inline void bsg_fence()      { __asm__ __volatile__("fence" :::); }
 //------------------------------------------------------
 
 // Pointers to remote locations (non-scratchpad) could be qualified
-// with bsg_attr_remote. Doing so would tell compiler to assign a different
-// address space to the contents of those pointers and latencies of
-// memory accesses would be considered as 20 cycles. For example,
-// `bsg_attr_remote float* foo;` essentially declares foo as `bsg_attr_remote float*`
-// type, and the compiler assumes loads from `foo` to take 20 cycles
-// on average.
+// with bsg_attr_remote to tell the compiler to assign a remote
+// address space to the data pointed by those pointers. Latencies of
+// memory accesses from those pointers would be considered as 20 cycles.
+// `bsg_attr_remote` acts as a type qualifier for pointers and globals,
+// and `bsg_attr_remote float* foo;` essentially declares foo as
+// `bsg_attr_remote float*` type. Compiler assumes that loads from `foo`
+// would have 20 cycle latency on average.
 #ifdef __clang__
 #define bsg_attr_remote __attribute__((address_space(1)))
 #elif defined(__GNUC__) && !defined(__cplusplus)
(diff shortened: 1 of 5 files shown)


drichmond:
Looks good to me, I'm going to verify that it does what we expect in my examples/experiments. 

How about adding a remote equivalent for GCC, but not G++? (@mrutt92 what was your keyword?)

mrutt92:
`remote`

vb000:
> Looks good to me, I'm going to verify that it does what we expect in my examples/experiments.
> 
> How about adding a remote equivalent for GCC, but not G++? (@mrutt92 what was your keyword?)

Would it fail for g++? Or, it'd just be ignored?

drichmond:
You can wrap in ifdef __cplusplus, to avoid it

mrutt92:
what Dustin said.

vb000:
I'd refer else case to be empty rather than if case... so, to be clear, would this be correct?

```
#ifdef __GNUC__
#ifndef __cplusplus
#define bsg_attr_remote remote
#endif
#endif
```

drichmond:
Yeah - you can use ifndef or ifdef, whatever floats your 🐐

vb000:
Well, the main question is, is that the right way to check for just "C"?... could you verify this in your test programs? Thanks!

drichmond:
Yeah, it's the way we check for C++ (or, not C) in all of our code. Its how you make header files C and C++ compatible so that the linker works, etc.

drichmond:
https://gcc.gnu.org/onlinedocs/cpp/Standard-Predefined-Macros.html

vb000:
Cool, thanks! Could you verify if the latest update on this branch that works? I don't have our forked toolchain installed and am currently working on tapeout branches...

drichmond:
Initial pass, this looks good. @vb000 Can you free up some cycles to verify this in C, C++ x LLVM and GCC? 

It would be good to have some simple spmd regression tests that use these in them. Even if they don't verify code generation correctness, we can at least verify that compilation doesn't break.

vb000:
Test sounds great, added one to exercise all three keywords.

`remote` doesn't seem to be present in latest riscv-gcc toolchain pointed by `bsg_manycore`. I defined `bsg_attr_remote` as empty for GCC for this PR. `bsg_attr_noalias` and `bsg_unroll` are working for both Clang and GCC.

mrutt92:
My bad - the keyword is actually `__remote` - https://github.com/bespoke-silicon-group/riscv-gcc/blob/894ea43d262f38c40592bbab93162225adb734d0/gcc/config/riscv/riscv.h#L925

software/bsg_manycore_lib/bsg_manycore.h
```diff
@@ -147,38 +147,41 @@ inline void bsg_fence()      { __asm__ __volatile__("fence" :::); }
...
-// `__remote float* foo;` essentially declares foo as `__remote float*`
+// `bsg_attr_remote float* foo;` essentially declares foo as `bsg_attr_remote float*`
 // type, and the compiler assumes loads from `foo` to take 20 cycles
 // on average.
 #ifdef __clang__
-#define __remote __attribute__((address_space(1)))
+#define bsg_attr_remote __attribute__((address_space(1)))
+#elif defined(__GNUC__) && !defined(__cplusplus)
+// TODO: Define the following with correct keyword for GCC
+#define bsg_attr_remote
```
mrutt92, line 160: __remote is what you want

vb000:
All three keywords working with both gcc and clang.

software/bsg_manycore_lib/bsg_manycore.h
vb000: That worked... even made slight improvement in the codegen with this testcase!

drichmond:
Can we make it build with both toolchains? Have kernel.c, kernel.cpp, and kernel.cc? Make one build with GCC, one with G++, and one with LLVM and link them all together?

vb000:
I kind of expected you'd ask for a perfect test ;). Let's defer it for a different PR, it's unclear how we can do that with current compilation setup.

Btw, I think the current compilation flow doesn't support .cc extension.

PR #394, opened 2020-09-26
Ruche network

tommydcjung:


testbenches/common/v/spmd_testbench.v
```diff
@@ -591,63 +670,65 @@ module spmd_testbench;
...
-    logic [S:N] dram_req_v_lo;
-    logic [S:N] dram_write_not_read_lo;
-    dram_ch_addr_s [S:N] dram_ch_addr_lo;
-    logic [S:N] dram_req_yumi_li;
-
-    logic [S:N] dram_data_v_lo;
-    logic [S:N][hbm2_data_width_p-1:0] dram_data_lo;
-    logic [S:N] dram_data_yumi_li;
+    logic [S:N][hbm2_num_channel_per_side_p-1:0] dram_req_v_lo;
+    logic [S:N][hbm2_num_channel_per_side_p-1:0] dram_write_not_read_lo;
```
drichmond, line 674: I like this change. Did Ruche Networks force this change?

machines/16x8_crossbar/Makefile.machine.include
```diff
@@ -8,17 +8,16 @@ BSG_MACHINE_VCACHE_DMA_DATA_WIDTH     = 32
 BSG_MACHINE_VCACHE_MISS_FIFO_ELS      = 32
 
 BSG_MACHINE_DRAM_SIZE_WORDS           = 536870912  # 2^29 words (2GB)
-BSG_MACHINE_DRAM_BANK_SIZE_WORDS      = 67108864
+BSG_MACHINE_DRAM_BANK_SIZE_WORDS      = 16777216
```
drichmond, line 11: Not relevant to this PR, but a question for discussion

Is there a way we can compute this, so that we don't need to change it for every machine?

machines/Makefile
```diff
@@ -2,7 +2,7 @@
 
 BSG_MANYCORE_DIR := $(shell git rev-parse --show-toplevel)
 # By convention, basejump_stl is in the same directory as $(BSG_MANYCORE_DIR)
-BASEJUMP_STL_DIR := $(BSG_MANYCORE_DIR)/../basejump_stl
+BASEJUMP_STL_DIR := $(abspath $(BSG_MANYCORE_DIR)/../basejump_stl)
```
drichmond, line 5: Out of curiosity, why was this necessary? (I like the change)

software/spmd/vcache_atomic_inc_multi/Makefile
```diff
@@ -1,13 +0,0 @@
-all: main.run
-
-OBJECT_FILES=main.o
```
drichmond, line 3: Why was this removed?

v/bsg_manycore_link_sif_tieoff.v
```diff
@@ -1,92 +1,55 @@
...
+      // handle errant fwd packet
+      if (link_sif_in.fwd.v) begin
+        $error("[BSG_ERROR] Errant fwd packet detected: src_x=%0d, src_y=%0d, dest_x=%0d, dest_y=%0d.",
+          fwd_packet.src_x_cord, fwd_packet.src_y_cord, fwd_packet.x_cord, fwd_packet.y_cord);
+      end
+
+      // handle errant rev packet
+      if (link_sif_in.rev.v) begin
+        $error("[BSG_ERROR] Errant rev packet detected: dest_x=%0d, dest_y=%0d.",
+          rev_packet.x_cord, rev_packet.y_cord);
```
drichmond, line 49: I like these. Would printing the tie-off's location be helpful?

v/bsg_manycore_link_sif_tieoff.v
```diff
@@ -1,92 +1,55 @@
...
-   wire   _unused1 = & (link_sif_i_cast.fwd.data); // mostly used; sometimes we use the return packet
-   wire   _unused2 = & (link_sif_i_cast.rev.data);
-   wire   _unused3 = link_sif_i_cast.rev.v;       // we ignore return packets coming in
-
-   // we don't inject any non-return data into the array
-   assign link_sif_o_cast.fwd.v         = 1'b0;
-   assign link_sif_o_cast.fwd.data         = 0;
-
-   // we will absorb incoming packets, but only if we can turn around and send back a credit
-   // on return channel
```
drichmond, line 45: Do we no longer do this?

drichmond:
Got through most of it in the time i had - so good so far

v/bsg_manycore_link_sif_tieoff.v
tommydcjung: If there is any packet arriving at the tie-off endpoint (whether it's because of HW or SW bug), it should be a hard fail. Having this loop back could be hiding a bigger bug, and it makes the wiring on the side of array dirty.

v/bsg_manycore_link_sif_tieoff.v
tommydcjung: the tie-off module does not have input my_x/y_i to know its location, and when the vcs is printing the error message, it will say from which instance is this error being triggered.

machines/16x8_crossbar/Makefile.machine.include
tommydcjung: Yeah, we could. That would be less error-prone.

v/bsg_manycore_link_sif_tieoff.v
drichmond: I like this answer. I prefer a hard fail, too.

machines/Makefile
tommydcjung: Just for aesthetic change.

v/bsg_manycore_link_sif_tieoff.v
drichmond: Cool. Wasn't sure that was the case.

machines/16x8_crossbar/Makefile.machine.include
drichmond: OK. Let's "file an issue" to do this in some alternate reality where we have time to do such things.

machines/Makefile
drichmond: Cool.

testbenches/common/v/spmd_testbench.v
tommydcjung: It's not necessary now, but it's useful, if we were to add 32x16 machine size.

drichmond: Got it. Still like it

software/spmd/vcache_atomic_inc_multi/Makefile
tommydcjung: I added it back.

v/bsg_manycore_ruche_x_link_sif_tieoff.v
```diff
@@ -0,0 +1,62 @@
...
+  // synopsys translate_off
+  always_ff @ (negedge clk_i) begin
+    if (~reset_i) begin
+      
+      if (invert_lp ^ ruche_link_in.fwd.v)
+        $error("[BSG_ERROR] Errant fwd packet detected.");
+
+      if (invert_lp ^ ruche_link_in.rev.v)
+        $error("[BSG_ERROR] Errant rev packet detected.");
+    end
```
drichmond, line 55: Is it possible to print the dest/source for these error messages, like in the mesh instances?

v/bsg_manycore_tile_ruche.v
```diff
@@ -0,0 +1,213 @@
...
+  assign links_sif_li[4].fwd = {
+    ruche_link_li[0][W].fwd[$bits(bsg_manycore_fwd_ruche_x_link_s)-1:(2*x_cord_width_p)+y_cord_width_p],
+    my_y_i,
+    ruche_link_li[0][W].fwd[(2*x_cord_width_p)+y_cord_width_p-1:0]
+  }; 
+  assign links_sif_li[4].rev = {
+    ruche_link_li[0][W].rev[$bits(bsg_manycore_rev_ruche_x_link_s)-1:x_cord_width_p],
+    my_y_i,
+    ruche_link_li[0][W].rev[x_cord_width_p-1:0]
+  };
```
drichmond, line 184: Just to make sure I understand this correctly

This is where the y-coordinate is being inserted (or removed) from the incoming or outgoing packet for this tile. 

The code knows the layout of the packet and slices/concatenates the signal appropriately.

Is my understanding correct?

drichmond (commented):
Overall: LGTM. 

Two questions left to answer.

v/bsg_manycore_tile_ruche.v
tommydcjung: That's right. The coordinates on the packet have fixed locations, so we are able to do this.

v/bsg_manycore_ruche_x_link_sif_tieoff.v
tommydcjung: Added.

v/bsg_manycore_tile_ruche.v
drichmond: Would it be better/possible to do this with a macro, instead of raw bit-indexing?

My concern is that while this is correct, it's hard to maintain. There's a lot of duplicated indexing

v/bsg_manycore_ruche_x_link_sif_tieoff.v
drichmond: Rad

v/bsg_manycore_tile_ruche.v
tommydcjung: alright, should look cleaner and more reusable with the new macros.

drichmond: I dig it

drichmond (approved):
Looks awesome.

CI?

PR #395, opened 2020-09-29
Added a SPMD program that demonstrates bsg_attr_remote

drichmond:
* Self-checking
* Passes in GCC and LLVM
* Runs the stats parser following `make`

My goal (tomorrow) is to generate the output for each example function (C, C++) and compiler (maybe?) into a text file, and then generate rules to automate this. My next goal is to fail regression (on this test) when code generation changes, so that the output and documentation is updated when the compiler changes (rare)

PR #396, opened 2020-09-29
Added a SPMD program that demonstrates bsg_attr_noalias

drichmond:
Like #395, adds an example SPMD program that demonstrates use-cases and pitfalls for bsg_attr_restrict

Works in LLVM and GCC. In rare (documented) cases, these two do not produce the same result.


mrutt92 (approved):

software/spmd/bsg_attr_noalias_ex/main.cpp
```diff
@@ -0,0 +1,320 @@
...
+#include <bsg_set_tile_x_y.h>
+// saxpy.h is generated by saxpy.py. It defines MACRO_A, MACRO_B, MACRO_GOLD, and N_ELS
+#include "saxpy.h"
+// saxpy-c.h and saxy-cpp.hpp contain the C and C++ function
+// definitions, respectively.
+#include "saxpy-c.h"
+#include "saxpy-cpp.hpp"
+#include <string.h>
+
+#define SAXPY_TAG_BASE 0
```
vb000, line 59: Are these solely intended for profiling? If so, a small comment or macro reflecting that could be useful... these seem to need some time to decode.

software/spmd/bsg_attr_noalias_ex/saxpy-c.c
```diff
@@ -0,0 +1,132 @@
...
+}
+
+
+/*
+ * Use caution when applying bsg_attr_noalias to multi-dimensional
+ * pointers. In this example LLVM produces worse code than GCC.
+
+ * This is a known issue in LLVM with a patch under review (as of 8/20):
+ * Google: restrict support in llvm
+ * LLVM Patch: https://reviews.llvm.org/D69542
```
vb000, line 41: Just fyi, I had tried this patch but found that noalias information was not being passed onto the llvm backend. The patch author's main goal has been to enable vectorization, which needs similar guarantees, on the data pointed by pointer, to that of our use case. https://lists.llvm.org/pipermail/llvm-dev/2020-June/142684.html

software/spmd/bsg_attr_noalias_ex/saxpy-c.c
```diff
@@ -0,0 +1,132 @@
...
+ * saxpy_noalias_A result: The compiler cannot guarantee that stores
+ * to C do not affect the second dimension of A, so loop iterations are
+ * not overlapped
+
+ * saxpy_A_noalias result: The compiler cannot guarantee that stores
+ * to C do not affect the first dimension of A, so loop iterations are
+ * not overlapped
+
+ * saxpy_noalias_noalias result: In LLVM, the compiler does not infer
+ * correct aliasing. GCC does (see below)
```
vb000, line 67: This is probably because LLVM has basic support of restrict compared to GCC. GCC probably supports most of what "C" standard says. Along those lines, a good addition could be a test with restrict on struct members. If GCC supports most of what C standard says, GCC should better hide load latency while accessing pointers that are struct members.

vb000 (approved):

PR #400, opened 2020-10-08
Optimized MM implementation for CUDA

drichmond:
* Added optimized MM code
* Simplified example code with attributes
* Initially written by @vb000 and @yodada


drichmond:
This is stale. A more up-to-date version can be found in https://github.com/bespoke-silicon-group/bsg_replicant/tree/matmul_blocked/examples/cuda/sgemm_group_cooperative

PR #403, opened 2020-10-08
fix amo_rl bug

tommydcjung:


drichmond (approved):
Looks good to me, but get @vb000's opinion too

vb000:
Makes sense, out_credits is increments a cycle later, but we have to stall now, correct? Btw, how did you find this bug, `fence` instruction test?

v/vanilla_bean/vanilla_core.v
```diff
@@ -1051,6 +1051,7 @@ module vanilla_core
   wire int_remote_load_in_exe = remote_req_in_exe & exe_r.decode.is_load_op & exe_r.decode.write_rd;
   wire float_remote_load_in_exe = remote_req_in_exe & exe_r.decode.is_load_op & exe_r.decode.write_frd;
   wire fdiv_fsqrt_in_fp_exe = fp_exe_r.fp_decode.is_fdiv_op | fp_exe_r.fp_decode.is_fsqrt_op;
+  wire remote_credit_pending = (out_credits_i != max_out_credits_p);
```
vb000, line 1054: A nitpick, could this be `remote_credit_pending = (out_credits_i != max_out_credits_p) | remote_req_in_exe`, so that we don't have to "or" when we use it?

vb000 (approved):

v/vanilla_bean/vanilla_core.v
tommydcjung: Nope, remote_credit_pending means what it is.

PR #404, opened 2020-10-08
network tests

tommydcjung:


drichmond (changes requested):
Needs more comments in the source files, particularly describing the intent of each test.

drichmond:
Are certain tests for random access network bandwidth, latency, etc?

tommydcjung:
Not sure what you are asking. It can be a combination of things

taylor-bsg:
Yes -- we always want to convey the intent of the code (ie the testing rationale, what is being tested, how is being tested, what is the expected behavior, what would be considered incorrect behavior) so that the reader can check whether the implementation matches the intent.

software/spmd/network_test_pointer_chasing2/main.c
```diff
@@ -0,0 +1,330 @@
+//  TEST DESCRIPTION
+//  Every tile starts at a different position in cyclic list.
```
drichmond, line 2: Questions I have when I read this: 

1. What is a cyclic list? A cyclic linked list?
2. Where is the cyclic linked list located, in DRAM or DMEM?
3. Where is the congestion? In the network? Or is there contention in the vcache? Is that machine size dependent?
4. Why do all the keys appear to be `-1`
5. Is the layout of the cyclic list dependent on what data is in DRAM? (Does the node array have to be declared first?)

I read the code and created the following description based on my inferences:

Test Description: Pointer Chasing Network Stress Test. 

A cyclic linked list is compiled into the binary, and loaded into DRAM. Each tile starts at a different position in the linked list, and iterates until it returns to its initial position. This generates random network traffic to the vcaches, and congestion in the network or vcache contention (depending on the size of the manycore array).

The linked list is a static linked list of 4096 nodes. The correctness of the code is dependent on the array being aligned with the start of the .dram section (i.e. nothing should be declared in .dram before the data array)

The tiles synchronize using a barrier when complete. 

But I still don't understand the stop condition? How are the keys and the stop conditions linked? 

Explain the assumptions in the code (dram, the stop condition). Will this only work for a certain size of machine?

software/spmd/network_test_pointer_chasing2/main.c
```diff
@@ -1,9 +1,19 @@
...
+//  Test Description: Pointer Chasing Network Stress Test.
+//
+//  A cyclic linked list is compiled into the binary, and loaded into DRAM. Each tile starts at a different position in the linked list, and iterates until it returns to its initial position.
+//  This generates random network traffic to the vcaches, and congestion in the network or vcache contention (depending on the size of the manycore array).
+//
+//  The linked list is a static linked list of 4096 nodes. The correctness of the code is dependent on the array being aligned with the start of the .dram section (i.e. nothing should be declared in .dram before the data array)
+//
+//  The tiles synchronize using a barrier when complete.
+//
+//  The program loads the "key" associated with linked load. If the key matches the tile id.
```
drichmond, line 11: Linked load? or linked list?

What happens if the key matches the tile id

drichmond:
Awesome, now do it for all of them. Try to put yourself in the shoes of someone who wants to reuse your code, but has a lot of other things to do (like me!)

tommydcjung:
Can you point out how other descriptions are not sufficient?

drichmond:
I've given you an example, and explained what questions I asked when I tried to understand it. I've provided an explanation for my reasoning. Try using that as a template for fixing the remaining ones.

drichmond (approved):

Issue #392, opened 2020-09-18
Use environment Python interpreter?


yodada:
in scripts like `bsg_manycore/software/py/bsg_manycore_link_gen.py`, python interpreter is hardcoded to `/usr/bin/python`, which not as flexible as `/usr/bin/env python`

drichmond:
Awesome. Submit a PR?

PR #413, opened 2020-10-26
Switch instances of /bin to env in Makefiles

drichmond:
This is the distro independent way of calling unix utility functions

https://en.wikipedia.org/wiki/Env

Related, but does not solve https://github.com/bespoke-silicon-group/bsg_manycore/issues/392


vb000 (approved):

yodada:
sweet!

PR #415, opened 2020-10-26
fix profiler pc output

tommydcjung:


software/spmd/simple_memory_test/main.c
```diff
@@ -0,0 +1,37 @@
+#include "bsg_manycore.h"
```
drichmond, line 1: Is this file supposed to be part of this PR?

testbenches/common/v/vanilla_core_profiler.v
```diff
@@ -1,3 +1,18 @@
...
+ *    vanilla_core_profiler.v
+ *
+ */
+
+
+//  Operation trace format = global_counter,x,y,pc,op
+//  PC field will print out the PC of the executed instruction in this cycle (i.e. instruction in EXE stage)
+//  When there is a bubble in EXE stage, it will print out the PC of the stalled instruction in ID.
+//  Or in case there was a bubble because of branch mispredict, it will print out the PC of the mispredicted branch instruction.
+//  If the pipeline gets stalled by incoming remote load response or idiv/fdiv, it will print out whatever PC is in the EXE stage.
```
drichmond, line 11: Just to confirm my understanding - the PC in the EXE stage during this time is NOT the PC in the idiv/fdiv stage? it could be the instruction immediately after?

drichmond:
Seems good to me. 

Have you run the bloodgraph tool on an output?

software/spmd/simple_memory_test/main.c
tommydcjung: Yes

testbenches/common/v/vanilla_core_profiler.v
tommydcjung: There is no idiv/fdiv stage. When the idiv/fdiv starts calculating, the instruction would be counted then.

software/spmd/simple_memory_test/main.c
drichmond: Does it test something in this PR?

testbenches/common/v/vanilla_core_profiler.v
drichmond: Right - but I'm asking about the PC that is stalled. The one that gets reported as stalling is the instruction immediately after the fdiv/idiv, right?

software/spmd/simple_memory_test/main.c
tommydcjung: I wrote this test to test that the profiler is working. It's not necessary

testbenches/common/v/vanilla_core_profiler.v
tommydcjung: It might not be the instruction immediately after. There could be many instructions in between.

tommydcjung:
This change should not affect the blood graph tool, since we are only changing what is being printed in the pc field.

testbenches/common/v/vanilla_core_profiler.v
drichmond: Yeah, that's why I said "could" earlier, but removed it in my repeat.

drichmond:
But the Bloodgraph uses the .csv file, right? And the PC histogram uses this file, and the PC counter field?

tommydcjung:
It seems like the blood_graph.py is already broken.

Traceback (most recent call last):
  File "blood_graph.py", line 44, in <module>
    from . import common
ValueError: Attempted relative import in non-package

drichmond:
Great, 

Out of curiosity, what command did you run?

tommydcjung:
```
python ../../py/vanilla_parser/blood_graph.py --trace vanilla_operation_trace.csv --stats vcache_operation_trace.csv
```

tommydcjung:
I think changing ```from . import common``` to ```import common``` fixes the issue.

drichmond:
What about `PYTHONPATH=../../py/vanilla_parser/ python3 -m vanilla_parser --trace vanilla_operation_trace.csv --stats vanilla_stats.csv`

tommydcjung:
People shouldn't need to add PYTHONPATH for a script to work, so I don't think that's an acceptable solution, and not a problem for this PR. But with the change I mentioned the blood graph works.
blood_detailed

drichmond:
Not a problem for this PR, and also, not a problem. The vanilla parser is a package so it needs to be on PYTHONPATH

You can see the documentation here: 

https://github.com/bespoke-silicon-group/bsg_manycore/tree/master/software/py/vanilla_parser

software/spmd/simple_memory_test/main.c
drichmond: I don't have an issue with adding another test - but it's not clear why the test is here. What does it test? How does it test the profiler?

The code can be removed, or stay as part of the PR. But it needs an explanation.

The rest of the PR is good to go

tommydcjung:
Still complains...
```
$ PYTHONPATH=../../py/vanilla_parser/ python3 -m vanilla_parser --trace vanilla_operation_trace.csv --stats vanilla_stats.csv
/usr/bin/python3: No module named vanilla_parser
```

drichmond:
Did I give you the right relative path? `../../py/vanilla_parser` was based on the path in the earlier comment

tommydcjung:
Yeah I think so
```
$ ls ../../py/vanilla_parser/
blood_graph.py  common.py  common.pyc  __init__.py  __main__.py  pc_histogram.py  README.md  stats_parser.py  trace_parser.py  vcache_stall_graph.py

```

drichmond:
Try `PYTHONPATH=../../py/`. I didn't put the final `..` from the documentation

drichmond:
FWIW, you've proven that the blood graph still works. No need to keep iterating on that

software/spmd/simple_memory_test/main.c
tommydcjung: It's just an easy program to analyze in the waveform and debug the profiler.

tommydcjung: If this is not checked in, I will need to spend time writing this program again next time I work on the profiler, .

vb000:
@tommydcjung Since profiler infrastructure is a set of scripts, to hand-off it as a single entity to Cornell for eg., we have to make it a python package. In fact, the very motivation of making this a package was a long list of back and forth questions about profiling on Cornell slack. 

Python packages need pythonpath, which could be made implicit by "installing" our profiler package in a project level build directory and having that directory in pythonpath. Because manycore doesn't do an explicit build and most has to be used from source as of now, we have to add the pythonpath.

tommydcjung:
Not sure if that's a standard solution, but my understanding was that we try to avoid using custom environmental variables. For this reason, the script printed error on the first time I tried.

vb000:
How do we avoid custom env variables?  If you're working with makefiles, it's done by explicitly setting paths in Makefiles and invoking commands using those variables. And, if you're working on command line, you provide explicit path on command line. Same is the case here, you either invoke the profiler package from makefiles by setting the PYTHONPATH in makefile or explicitly provide it on command line.

To make the objective clear, the old command to profile was: 
```
python <path to profiler>/stats_parser.py <flags> 
python <path to profiler>/blood_graph.py <flags>
..
``` 

The new command is just:
 ```
PYTHONPATH=<path to profiler>/.. python -m vanilla_parser
```
So, this is not any worse than explicitly invoking scripts. Moreover, a single generic command (no need to remember flags) generates all the necessary information from multiple parsers.

In addition to that we have this flexibility...

If you want to run specific profilers, you can do it as:
 ```
PYTHONPATH=<path to profiler>/.. python -m vanilla_parser --only trace_parser vcache_stall_graph
```

If you want to add something to the default set of profilers:
 ```
PYTHONPATH=<path to profiler>/.. python -m vanilla_parser --also trace_parser
```

mrutt92:
FWIW I wouldn’t classify PYTHONPATH as a custom environment variable. It’s pretty standard.

mrutt92:
...but I’m wondering if this the right forum to have this discussion? Should we move this to an issue and make it a discussion?

tommydcjung:
I see the problem you are trying to solve (making it easier to use for users), and I'm sure there are various ways to achieve that using more object-oriented solution that doesn't require making the infrastructure rely on modifying the env variable.

tommydcjung:
Agreed with @mrutt92. happy to argue more in an issue.

software/spmd/simple_memory_test/main.c
drichmond: I have no doubt that this program is useful and I'm not asking that it be removed. However, like in #404 it is helpful if the test documents its intent: what it does, and why it matters. A few sentences is sufficient.

I have no issues with any of the rest of this PR. I was particularly happy that there's now a comment in testbenches/common/v/vanilla_core_profiler.v!

drichmond (approved):

PR #417, opened 2020-10-28
WIP: Fixed vcache profiler bugs

drichmond:
* Order of operations: Python binds or higher than >
* Counting of repeated tags. Previously, only the last tag was counted

There's still a bug here. I'm trying to figure out how to address it without re-writing the cache parsing code. Posting this so @yodada and @eafurst are aware there is an issue


drichmond:
Superseded by #428

PR #418, opened 2020-10-28
Check for empty tile pc_list when building pc_histogram

marcusnchow:
Small profiler fix
This happens when a kernel does not fully occupy every available tile and some tiles do not have any pc_lists
If pc_list is empty the profiler crashes. now it can silently continue to another tile

### DIFF
--- a/software/py/vanilla_parser/pc_histogram.py
+++ b/software/py/vanilla_parser/pc_histogram.py
@@ -197,19 +197,17 @@ class PCHistogram:
 
         # Repeat once more for the last basic block 
         # Number of times basic block is executed
-        block_pc_cnt = pc_cnt[pc_list[start]]
+        if pc_list:
+            block_pc_cnt = pc_cnt[pc_list[start]]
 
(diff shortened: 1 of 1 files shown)


software/py/vanilla_parser/pc_histogram.py
```diff
@@ -197,19 +197,17 @@ def __generate_pc_histogram(self, pc_cnt, pc_cycle):
 
         # Repeat once more for the last basic block 
         # Number of times basic block is executed
```
drichmond, line 199: Can you update the comment for the new if statment?

drichmond:
Thanks for fixing this!

software/py/vanilla_parser/pc_histogram.py
marcusnchow: I have it checking for pc_list earlier and the function now and comment that the function returns empty dirctionary if pc_list is empty

drichmond (approved):

PR #419, opened 2020-10-29
remote_load_trace

tommydcjung:
example output
```
start_cycle,end_cycle,src_x,src_y,dest_x,dest_y,type,latency
25059,25069,0,2,1,0,icache,10
25074,25084,0,2,1,0,icache,10
25089,25099,0,2,1,0,icache,10
25104,25114,0,2,1,0,icache,10
25119,25129,0,2,1,0,icache,10
25133,25145,0,2,2,0,int,12
25134,25146,0,2,2,0,icache,12
25151,25163,0,2,2,0,icache,12
25168,25180,0,2,2,0,icache,12
25185,25197,0,2,2,0,icache,12
25202,25214,0,2,2,0,int,12
25219,25231,0,2,2,0,int,12
25236,25248,0,2,2,0,int,12
25255,25267,0,2,2,0,icache,12
25272,25284,0,2,2,0,icache,12

```

### DIFF
--- a/testbenches/common/v/spmd_testbench.v
+++ b/testbenches/common/v/spmd_testbench.v
@@ -1103,7 +1103,6 @@ module spmd_testbench;
   );
 
 
-
   // router profiler
   if (network_cfg_lp[e_network_mesh] | network_cfg_lp[e_network_half_ruche_x]) begin
     bind bsg_mesh_router router_profiler #(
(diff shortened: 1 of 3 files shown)


drichmond:
I like this (so consider it an approval) - but I want to think about the parser and figure out how it might figure into this PR.

drichmond:
Two (ish) questions, while thinking about the parser: 

- Are atomics counted as integer remote loads? Can they be counted in their own category?
- Can we print the PC of the remote load, and the PC of the instruction fetch? This is necessary to backtrack and determine which remote loads are causing issues. 

I'll post more as I think of them

tommydcjung:
I could make atomic its own category.
The PC of remote load can be correlated with vanilla_core trace output since they share the same global_ctr as a key.

testbenches/common/v/spmd_testbench.v
```diff
@@ -36,6 +36,9 @@ module spmd_testbench;
   parameter axi_data_width_p = 256;
   parameter axi_burst_len_p = 1;
 
+  `define ORIGIN_X_CORD_P 0
+  `define ORIGIN_Y_CORD_P 2
+
```
drichmond, line 41: This is fine, for now. But I think in the long run we should define where the origin is in Makefile.machine.include. We kinda do in replicant.

Can you file an issue for that?

tommydcjung: Yeah we will add that in Makefile.machine.include, when we add a feature to shift x-cord.

drichmond (approved):

PR #422, opened 2020-10-29
vcache_atomic_inc_multi

tommydcjung:


drichmond:
What's the difference between this and other atomic tests? This one is multi-core?

tommydcjung:
This one has multiple locks and multiple data.

drichmond:
Ah, so the other one has multiple cores too?

tommydcjung:
Yes...if there is only one core, what's the point of locking?

drichmond:
Well, could be testing the lock implementation

But yes. I just don't see the tile group commands in the other test. But that's not an issue for this test -- this one looks good.

drichmond (approved):

PR #428, opened 2020-11-05
Rewrite Victim Cache Profiler Parser

drichmond:
Depends on #426 

The original parser had a few bugs, at least one of which was fundamental. This PR solves this and adds a few new features

New Features:
1. Atomic Miss Counting
2. Response Network Congestion counting
3. Memory access latency
4. Field documentation (see examples below)

Bug Description: 

1. Order of operations (this was in #417)
2. Multiple start/end calls with the same tag (A temporary fix was in #417)
3. Perceived ordering of Start/End relative to tile program order

The original parser was written so that only the first (ever, within a simulation run) and last (ever, within a simulation run) statistics for a given (tag, group) were counted. This ignores that a (tag, group) may have a loop in it, and cause start/end to be called multiple times. If this happens, then the profiler inadvertently counts all cache operations between the first (ever) and last (ever) call to start/stop respectively. 

What we want instead is the sum of the windows, where each window is defined by a start/end window pair. This is well defined when a single tile is iterating and calling start/end, but complicated when multiple tiles are iterating on the same start/end calls.

The problem is as follows: Tiles execute independently. This means that the fastest tile in a tile group can be on a different loop iteration than the slowest tile in a tile group. 

This is actually quite common. In the matrix multiply code at the end, there are two tags : 0, and kernel. Tag 0 surrounds the computation of an output block, and a tile may compute multiple output blocks. Thus, it is important to match up the iterations of tag 0 with other corresponding iterations that happened around the same time.

For consistent and precise statistics, we need the window defined by the start/end calls within a tag to be the first start (from any tile) and last end (from any tile), from matching iterations.

This approach will fail when tiles do not agree on an iteration order - for example if the tiles independently process work dynamically, the work items are different methods, and the work items are different tags. Then Tag 0 on Tile 0 doesn't necessarily correspond to Tag 0 on Tile 1 because the work items are processed differently. A good example of this would be a work-stealing library. In cases like this, only coarse regions should be profiled.

(here is no good way to handle this case aside from doing per-tile vcache statistics -- which this PR enables, but does not do.)

This is PR is mostly written with pandas. ~10% of the code is dedicating to parsing the CSV file and organizing the data (a significant reduction), and the rest is pretty-printing and documentation to describe the problem above and the solution. 

This is an example of the per-bank statistics:

```
################################################################## Per-Bank Victim Cache Stats ###################################################################
Table Fields: 
	- Cache Coordinate (Y,X): Cache Coordinate within HammerBlade Pod
	- Total Cycles: Total Cache Execution Cycles
	- # Misses: Total Number of Cache Misses
	- Operations: Total Number of Cache Operations (Loads + Stores + Atomics + Management)
	- Miss Rate (%): 100 * (Number of Misses / Number of Ops)
	- Memory Access Latency: Average Memory Access Latency for Misses (Total Miss Cycles / Number of Misses)
	- Miss Percent: 100 * (Total Miss Cycles / Total Cycles)
	- Idle Percent: 100 * (Total Idle Cycles / Total Cycles)
	- Response Stall Percent: 100 * (Total Response Stall Cycles / Total Cycles)
	- Operations Percent Cycles: 100 * (Total Operation Cycles / Total Cycles)

Note: inf (Infinite) occurs when a tag window captures miss stall cycles that bleed into its window, but has no misses
============================================================================= Tag: 0 =============================================================================
Cache Coordinate (Y,X)      Total Cycles    # Misses    # Operations    Miss Rate    Mem. Latency    Miss Percent    Idle Percent    Stall Percent    Ops. Percent
------------------------  --------------  ----------  --------------  -----------  --------------  --------------  --------------  ---------------  --------------
(0, 0)                            216299          64           34670         0.18           72.20            2.14           81.61             0.23           16.03
(0, 1)                            216299          64           34699         0.18           70.16            2.08           81.52             0.37           16.04
(0, 2)                            216299          64           34685         0.18           77.83            2.30           80.25             1.41           16.04
(0, 3)                            216299          63           34742         0.18           75.24            2.19           81.35             0.40           16.06
(0, 4)                            216299          64           34943         0.18           72.78            2.15           81.29             0.40           16.15
(0, 5)                            216299          64           35072         0.18           77.89            2.30           80.18             1.30           16.21
(0, 6)                            216299          63           35146         0.18           74.29            2.16           81.10             0.49           16.25
(0, 7)                            216299          64           35408         0.18           69.52            2.06           81.43             0.14           16.37
(6, 0)                            216299          64           35478         0.18           73.06            2.16           81.13             0.31           16.40
(6, 1)                            216299          64           35655         0.18           76.59            2.27           80.64             0.61           16.48
(6, 2)                            216299          64           35558         0.18           66.98            1.98           79.59             1.99           16.44
(6, 3)                            216299          64           35590         0.18           69.83            2.07           80.66             0.82           16.45
(6, 4)                            216299          64           35591         0.18           72.47            2.14           80.74             0.66           16.45
(6, 5)                            216299          64           35630         0.18           81.72            2.42           78.82             2.29           16.47
(6, 6)                            216299          64           35621         0.18           71.69            2.12           80.59             0.82           16.47
(6, 7)                            216299          64           35103         0.18           73.77            2.18           81.29             0.29           16.23

```

Here's an example of per-tag statistics:

```
#######################  Per-Tag Victim Cache Stats  #######################
Table Rows:
	Load Operations:
		-instr_ld_l[wu,w,hu,h,du,d,bu,b]: Load [w]ord/[h]alf/[b]yte/[d]ouble [u]nsigned/[]signed
	Store Operations:
		-instr_sm_s[w,h,d,b]: Store [w]ord/[h]alf/[b]yte/[d]ouble
	Cache Management Operations:
		-instr_tagst: Tag Store (Not caused by Vanilla Core)
		-instr_tagfl: Tag Flush (Not caused by Vanilla Core)
		-instr_taglv: Tag Load Valid (Not caused by Vanilla Core)
		-instr_tagla: Tag Load Address (Not caused by Vanilla Core)
		-instr_afl: Address Flush (Not caused by Vanilla Core)
		-instr_aflinv: Address Flush Invalidate (Not caused by Vanilla Core)
		-instr_ainv: Address Invalidate (Not caused by Vanilla Core)
		-instr_alock: Address Lock (Not caused by Vanilla Core)
		-instr_aunlock: Address Unlock (Not caused by Vanilla Core)
	 RISC-V Atomic Operations:
		-instr_amoswap: Atomic Swap
		-instr_amoor: Atomic OR
	 Cache Stall Operations:
		-stall_miss: Miss Operation (Stall)
		-stall_idle: Idle Operation (Stall)
		-stall_rsp: Response Network Congestion Stall

 *** All operations take one cycle. *** 
================================== Tag: 0 ==================================
Operation Cycle Counts                                                      
----------------------------------------------------------------------------
                              Count    % of Type Cycles    % of Total Cycles
--------------------------  -------  ------------------  -------------------
--instr_ld_lwu                    0                0.00                 0.00
--instr_ld_lw                531897              100.00                15.37
--instr_ld_lhu                    0                0.00                 0.00
--instr_ld_lh                     0                0.00                 0.00
--instr_ld_ldu                    0                0.00                 0.00
--instr_ld_ld                     0                0.00                 0.00
--instr_ld_lbu                    0                0.00                 0.00
--instr_ld_lb                     0                0.00                 0.00
Load Operation Total         531897              100.00                15.37
--------------------
--instr_sm_sw                 31694              100.00                 0.92
--instr_sm_sh                     0                0.00                 0.00
--instr_sm_sd                     0                0.00                 0.00
--instr_sm_sb                     0                0.00                 0.00
Store Operation Total         31694              100.00                 0.92
---------------------
--instr_amoswap                   0                0.00                 0.00
--instr_amoor                     0                0.00                 0.00
Atomic Operation Total            0                0.00                 0.00
----------------------
--instr_tagst                     0                0.00                 0.00
--instr_taglv                     0                0.00                 0.00
--instr_tagla                     0                0.00                 0.00
--instr_tagfl                     0                0.00                 0.00
--instr_aunlock                   0                0.00                 0.00
--instr_alock                     0                0.00                 0.00
--instr_ainv                      0                0.00                 0.00
--instr_aflinv                    0                0.00                 0.00
--instr_afl                       0                0.00                 0.00
Management Operation Total        0                0.00                 0.00
--------------------------
--stall_rsp                   27125                0.94                 0.78
--stall_miss                  75115                2.59                 2.17
--stall_idle                2794953               96.47                80.76
Stall Operation Total       2897193              100.00                83.71
---------------------
Total Cycles                3460784              100.00               100.00
----------------------------------------------------------------------------

----------------------------------------------------------------------------
Miss Statistics                                                             
----------------------------------------------------------------------------
Type       Misses    Accesses    Miss Rate (%)
-------  --------  ----------  ---------------
Stores       1022       31694             3.22
Loads           0      531897             0.00
Atomics         0           0             0.00
Total        1022      563591             0.18
----------------------------------------------------------------------------
```

Example Code: 

```c++

        // Start profiling
        bsg_cuda_print_stat_kernel_start();

        // Iterate through available output blocks in the X and Y
        // dimensions. Jump by the tile group size between iterations
        // to assign unique work.
        // Yes, this should be TGID
        for (int by_i = __bsg_y; by_i < r1/BY; by_i += BSG_TILE_GROUP_Y_DIM) {
                for (int bx_i = bx_start; bx_i < bx_end ; bx_i ++) {

                        // Multiply the blocks, and accumulate into the result

                        bsg_cuda_print_stat_start(0);
                        for (int bz_i = 0; bz_i < blocks; bz_i++) {
                                load_block<BY, BX, LOAD_M1_TRANSPOSED>(block_row, mat1, mat1_strides, by_i, bz_i);

                                load_block<BY, BX, false>(block_col, mat2, mat2_strides, bz_i, bx_i);

                                accum_block<BY, BY/2, BX, BX/2, LOAD_M1_TRANSPOSED>(psum, block_row, block_col);

                        }

                        // Store the result, AND zero the psum array
                        // to leverage parallel remote and local
                        // stores.
                        store_block_and_reset<BY, BX>(psum, result, result_strides, by_i, bx_i);

                         bsg_cuda_print_stat_end(0);
                }
        }

        // End profiling
        bsg_cuda_print_stat_kernel_end();

```

### DIFF
--- a/software/py/vanilla_parser/stats_parser.py
+++ b/software/py/vanilla_parser/stats_parser.py
@@ -687,7 +687,7 @@ class CacheTagStats(CacheStats):
 
         # Compute the miss rate by dividing the misses by the accesses
         # Nans are expected -- 0/0. Just turn them into 0's
-        df["Miss Rate (%)"] = (100 * df["Misses"] / df["Accesses"]).fillna(0)
+        df["Miss Rate (%)"] = 100 * (df["Misses"] / df["Accesses"]).fillna(0)
 
         # Set index to the type
         df = df.set_index(["Type"])
@@ -823,23 +823,23 @@ class CacheBankStats(CacheStats):
         doc += "\t- Operations: Total Number of Cache Operations (Loads + Stores + Atomics + Management)\n"
         pretty["# Operations"] = ops
 
-        doc += "\t- Miss Rate (%): 100 * (Number of Misses / Number of Ops)\n"
-        pretty["Miss Rate"] = 100 * self.df["total_miss"] / ops
+        doc += "\t- Miss Rate: 100 * (Number of Misses / Number of Ops)\n"
+        pretty["Miss Rate (%)"] = 100 * (self.df["total_miss"] / ops)
 
         doc += "\t- Memory Access Latency: Average Memory Access Latency for Misses (Total Miss Cycles / Number of Misses)\n"
         pretty["Mem. Latency"] = (self.df["stall_miss"] / self.df["total_miss"]).fillna(0)
 
-        doc += "\t- Miss Percent: 100 * (Total Miss Cycles / Total Cycles)\n"
-        pretty["Miss Percent"] = 100 *(self.df["stall_miss"] / self.df["global_ctr"])
+        doc += "\t- Percent Miss Cycles: 100 * (Total Miss Cycles / Total Cycles)\n"
+        pretty["Percent Miss Cycles"] = 100 *(self.df["stall_miss"] / self.df["global_ctr"])
 
-        doc += "\t- Idle Percent: 100 * (Total Idle Cycles / Total Cycles)\n"
-        pretty["Idle Percent"] = 100 *(self.df["stall_idle"] / self.df["global_ctr"])
+        doc += "\t- Percent Idle Cycles: 100 * (Total Idle Cycles / Total Cycles)\n"
+        pretty["Percent Idle Cycles"] = 100 *(self.df["stall_idle"] / self.df["global_ctr"])
 
-        doc += "\t- Response Stall Percent: 100 * (Total Response Stall Cycles / Total Cycles)\n"
-        pretty["Stall Percent"] = 100 *(self.df["stall_rsp"] / self.df["global_ctr"])
+        doc += "\t- Percent Response Stall Cycles: 100 * (Total Response Stall Cycles / Total Cycles)\n"
+        pretty["Percent Stall Cycles"] = 100 *(self.df["stall_rsp"] / self.df["global_ctr"])
 
-        doc += "\t- Operations Percent Cycles: 100 * (Total Operation Cycles / Total Cycles)\n"
-        pretty["Ops. Percent"] = 100 * (ops / self.df["global_ctr"])
+        doc += "\t- Percent Operations Cycles: 100 * (Total Operation Cycles / Total Cycles)\n"
+        pretty["Percent Ops."] = 100 * (ops / self.df["global_ctr"])
 
         doc += "\n"
         doc += "Note: inf (Infinite) occurs when a tag window captures miss stall cycles that bleed into its window, but has no misses"
--- a/software/py/vanilla_parser/stats_parser.py
+++ b/software/py/vanilla_parser/stats_parser.py
@@ -508,7 +508,7 @@ class CacheTagStats(CacheStats):
         mismatches = self.find_mismatches(starts, ends)
         if(list(mismatches)):
             raise RuntimeError("Unpaired calls to Start/End detected."
-                               f" Check the following: {tuple(mismatches.names)}:"
+                               f" Check the following: {tuple(mismatches)}:"
                                f"{list(mismatches)}")
 
         # For all tag iterations, find the minium arrival time
--- a/software/py/vanilla_parser/stats_parser.py
+++ b/software/py/vanilla_parser/stats_parser.py
@@ -254,9 +254,9 @@ class CacheStatsParser:
 
         return p
 
-    # Get device-level characteristics: dim, origin
+    # Get group-level characteristics: dim, origin
     @classmethod
-    def parse_dev_characteristics(cls, df):
+    def parse_group_characteristics(cls, df):
         dim = ManycoreCoordinate(
             df["Tile Coordinate (Y,X)"].map(lambda l: l.y).max() + 1,
             df["Tile Coordinate (Y,X)"].map(lambda l: l.x).max() + 1)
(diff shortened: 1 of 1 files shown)


mrutt92:
This sounds great. I'm wondering if you have a kernel that causes the problem fixed by this PR. Would you mind posting it or posting a link if it exists already? The smaller test case, the easier it will be to understand. If there's a specific platform/machine pair on which the bug manifests please post that too.

In general though, I'm a fan of using pandas for parsing and analyzing our profiler output. There's no reason why parsing a csv should be more than a one-liner afaik.

mrutt92:
If I'm understanding this right, too-down BFS would be an example due to workload imbalance.

drichmond:
Bug 1 is caused by... any code. 

Bug 2 will manifest in any code where a tag is called multiple times. Workload imbalance doesn't matter.

Bug 3 manifests in this matrix multiply code, which isn't simple: https://github.com/bespoke-silicon-group/bsg_manycore/blob/ci_mm_more_opt/software/spmd/bsg_cuda_lite_runtime/mm_opt/kernel.cpp

The imbalance is caused by the locality of data relative to each tile. It happens on any machine.

mrutt92:
Cool. For the sake of posterity, would you mind posting the source code for the example report in your og brief?

drichmond:
Yeah, i'll just add the main kernel, not the leaf functions.

software/py/vanilla_parser/stats_parser.py
```diff
@@ -137,7 +167,767 @@ def isKernelEnd(self):
...
+    def parse_raw_tag(cls, df):
+        # Parse raw_tag data using CudaStatTag
+        cst = df.raw_tag.map(CudaStatTag)
+
+        p = pd.DataFrame()
+        # Update the table with information parsed from CudaStatTag
+        p["Tile Group ID"] = cst.map(lambda e: e.getTileGroupID)
+        p["Tag"] = cst.map(lambda e: e.getTag)
+        p["Action"] = cst.map(lambda e: e.getAction)
+        p["Tile Coordinate (Y,X)"] = cst.map(lambda e: ManycoreCoordinate(e.y, e.x))
```
drichmond, line 253: Is CudaStatTag the Tile Coordinate? or Tile-Group Coordinate?

mrutt92 (approved):

PR #432, opened 2020-11-12
[bsg_manycore_link_gen] fix location counter boundary check

tommydcjung:


drichmond:
Is the memcpy assembly part of this PR?

tommydcjung:
It doesn't have to be. I added as an example to show that we can allocate full 4KB.

drichmond:
Well, i'm not opposed to it, since it does test the bug you found. But it would be better to have a test whose sole purpose is to test the bug you found. It doesn't need to do anything, aside from allocate 4KB.

Alternatively, are you done with the DMA test, or do you anticipate modifying it? If you anticipate modifying it, then it shouldn't be part of this PR. 

If you do keep it, as in previous PRs, the program needs a simple descriptive comment. 

It now seems that there are two bugs in this PR? One for NBF and the other for the location counter.

tommydcjung:
I replaced the test with something else.
They are the same bug.

drichmond (approved):
Looks good to me - but I can't speak for the linker change. @vb000 can you verify that?

vb000:
Great test to have, linker fix looks good to me.

PR #443, opened 2020-12-08
[quicktouch] shorten quicktouch for gate-level sim test

tommydcjung:
Reduces the nbf file from 2000s to 445.


drichmond (approved):
Seems fine. Is this just because there are more dynamic instructions now?

tommydcjung:
The majority comes from bsg_printf()

drichmond:
That'll do it.

We actually see a lot of similar situations in replicant-land. bsg_printf usually ends up in the icache because of the linker, so if the kernel is small you take a double, or triple, hit. Increased loading times, icache misses from every tile to start the program, etc

PR #447, opened 2020-12-14
x_cord_width_p, start_x_cord_p

tommydcjung:


drichmond:
Is it necessary to have a "Start Y Coord" too?

tommydcjung:
No, the top vcache has y=0, and you cannot attach anything above that.

drichmond:
I see. This is about the global x offset. 

So this is separate from the X/Y origin we discussed previously?

tommydcjung:
origin Y is fixed at 2, but origin X is equal the start_x_cord_p. So I don't think we need separate machine parameters for origin x/y.

drichmond:
We will, but I'm fine with leaving it out of this PR

machines/README.md
```diff
@@ -15,12 +15,14 @@ Machine.machine.include parameters
                                             If this flag is set to zero, the manycore can only operate in NO-DRAM mode,
                                             meaning that the vcache is only used as block memory.
 - BSG_MACHINE_MAX_EPA_WIDTH             =   Width of word address on the mesh network.
-- BSG_MACHINE_MEM_CFG                   =   e_vcache_non_blocking_axi4_nonsynth_mem
+- BSG_MACHINE_MEM_CFG                   =   e_vcache_non_blocking_axi4_nonsynth_mem (look at bsg_manycore_mem_cfg_pkg.v)
```
drichmond, line 18: Why not: "Memory Configuration for manycore machine (e.g. infinite memory, see bsg_manycore_mem_cfg_pkg.v for valid values)"

drichmond:
Question for discussion: Why "Start X Coordinate" and not Origin X Coordinate? We've always called upper-most left coordinate the origin -- why change now?

drichmond:
Otherwise, PR looks good

PR #448, opened 2020-12-15
cache_to_axi use bsg_arb_round_robin

tommydcjung:


tommydcjung:
Using bsg_arb_round_robin to support num_cache_p beyond 16.

taylor-bsg:
What's the motivation for switching the implementation instead of just upgrading the existing module to handle larger els_p? What are the scaling implications? Which system uses this module? FPGA?

dpetrisko:
https://github.com/bespoke-silicon-group/basejump_stl/pull/305

tommydcjung:
This is used for simulation, if you were to go beyond 16 in actual hw, you wouldn't want to use this module any way. The problem with the old round robin module is that if you want to go beyond 32, then you have to update the module again.

drichmond:
Is this just for simulation? If it is - why not move this file into the testbenches folder (or similar) so that it doesn't accidentally get used in Silicon?

taylor-bsg:
If it is synthesizeable but not scalable, okay just to note it in the
comments — both where it is used and in the implementation.

PR #451, opened 2020-12-18
Added missing file to arch_filelist.mk

drichmond:
* Also in sim_filelist.mk, which is not used in Repliant.

I think this is independent of #448?


tommydcjung:
This used to be in sim filelist, because this module was used in crossbar network, but we also use this in the mesh router.
can you also remove it from the sim_filelist?

drichmond:
Yep, good catch. Fixed

tommydcjung (approved):

PR #459, opened 2021-01-26
Bigblade dmem addr shift

tommydcjung:


software/spmd/asm_dmem_test/main.S
```diff
@@ -5,10 +5,6 @@
 #include "bsg_manycore_asm.h"
 #define N 1024
 
-// allocate 4KB in DMEM
-.data
-  dmem_arr: .space (N*4)
-
```
drichmond, line 11: isn't the test still allocating 4KB in DMEM?

software/py/nbf.py
```diff
@@ -233,7 +233,7 @@ def init_dmem(self, pod_origin_x, pod_origin_y):
         x_eff = self.tgo_x + x + pod_origin_x
         y_eff = self.tgo_y + y + pod_origin_y
           
-        for k in range(1024, self.bsg_data_end_addr):
+        for k in range(0, self.bsg_data_end_addr):
```
drichmond, line 236: I'm generally not a fan of hard-coded values. Is there not a self.bsg_data_start_addr that can be referenced?

`DMEM_BASE_EPA = 0x400` on line 23 should be updated too.

software/py/bsg_manycore_link_gen.py
```diff
@@ -174,7 +175,7 @@ def script(self):
     # DMEM sections
     for i, m in enumerate(section_map):
       sec = m[0]
-      laddr = '0x1000'
+      laddr = '0x0008'
```
drichmond, line 178: What about VMA start, instead of a hard-coded value?

software/py/bsg_manycore_link_gen.py
```diff
@@ -135,8 +135,9 @@ def script(self):
     # LMA (Load Memory Address)    => NPA used by loader
     # VMA (Virtual Memory Address) => Logical address used by linker for
     #                                 symbol resolutions
-    _DMEM_VMA_START   = 0x1000
-    _DMEM_VMA_SIZE    = 0x1000
+    # bsg-tommy: we reserve the first two words of DMEM for interrupt handlers.
+    _DMEM_VMA_START   = 0x0008
+    _DMEM_VMA_SIZE    = 0x0ff8
```
drichmond, line 140: This seems odd. The Tile can still access the addresses 0x0-0x7 right? Wouldn't that mean that the VMA region actually starts at 0x0000?

It seems like a cleaner way to do this would be to define VMA start to 0x0000, and then create variables at 0x0 and 0x4, either here, or in some assembly file. An added benefit would be that then C/C++ can directly dereference those locations instead of doing pointer manipulations

Am I missing something?

software/spmd/asm_dmem_test/main.S
drichmond: On a related note, dmem_arr was a variable that could be referenced. Was there a reason it was replaced by a 0 offset? 

It seems nice to keep things like this. If we ever move the location of DMEM, or change where DMEM allocates data pushing the dereferencing logic into the compiler would save us a lot of work -- like all the changes in the .S files here.

software/spmd/load_dependency/main.S
```diff
@@ -9,30 +9,30 @@ li ra, 10;
...
     // Single cycle remote loads
     // nops to test zero delay remote loads
-    bsg_asm_remote_load(s0, 0, 0, 0x1000); // 1
-    bsg_asm_remote_load(s1, 0, 0, 0x1004); // 2
-    bsg_asm_remote_load(s2, 0, 0, 0x1008); // 3
-    bsg_asm_remote_load(s3, 0, 0, 0x100c); // 4
+    bsg_asm_remote_load(s0, 0, 0, 0x0000); // 1
+    bsg_asm_remote_load(s1, 0, 0, 0x0004); // 2
+    bsg_asm_remote_load(s2, 0, 0, 0x0008); // 3
+    bsg_asm_remote_load(s3, 0, 0, 0x000c); // 4
```
drichmond, line 35: Is there a symbol we can dereference instead of hardcoding 0x0000?

drichmond (changes requested):
Generally: Good. 

My comments (changes?) reflect sw engineering questions. It seems like there's an opportunity to name a symbol instead of using hardcoded values. This makes the code easier to understand, and automate changes (if we ever change the address map again)

drichmond:
Making more linker symbols would also make porting CUDA lite easier. We could read the values from the binary instead of hard-coding them in the runtime.

software/py/bsg_manycore_link_gen.py
tommydcjung: This doesn't prevent accessing 0~7. The point is that we prevent the linker from placing objects in this space without having to include some special code.

software/spmd/asm_dmem_test/main.S
tommydcjung: The programmer should know that DMEM starts from 0. It's not necessary to create .data

software/spmd/load_dependency/main.S
tommydcjung: The programmer  knows that DMEM starts from 0.

software/py/bsg_manycore_link_gen.py
drichmond: But why not declare symbols for the linker, and force it to reserve those spaces? That would make it easier to access the two reserved locations.

tommydcjung: okay, how do you do that?

tommydcjung: Actually, the location is already defined as nullptr. So the programmer can dereference from that?

drichmond (approved):
Looks good to me. I still feel that we should not replace the dmem_arr declarations with hard-coded 0 to make the code easier to quickly understand, 

But functionally - looks great to me! Glad to see this in!

drichmond:
My last comment is - I would hold off on merging this until Max and I can verify all of the current head commits and create a branch in bladerunner (should happen by noon?)

PR #460, opened 2021-01-26
More Interrupt Testing

sripathi-muralitharan:
This PR adds tests for the following cases - 
1. Trace/Remote interrupt interaction with integer divider, multiplier and floating point divider
2. Behavior under multiple remote interrupts (not an expected case but tried to check if the latest arriving remote interrupt overrides the previous ones)
3. Generating I$ misses in the interrupt handler to check behaviour of mret in these situations (specifically please take a closer look at the dual interrupt case with misses in the handler - either there is a bug in my brain, in my test code or in the implementation)

I have added a short description at the top of the file and comments at all places to describe my intent. 

There are more tests to come so this PR can be kept open and periodically reviewed at the discretion of the maintainer...


sripathi-muralitharan:
Update since last time: There was a small bug which Tommy fixed. Working on coverage for `pc_n` (looking at conditions that change the PC)

Added a regression suite of interrupt tests and a module that gathers coverage data for the next PC logic. The makefile changes aren't super clean and can be revisited later if we want to merge this in. I have managed to generate coverage reports on a per-test basis and users can view that using DVE but I am having some problems in getting a consolidated coverage report across all tests using the URG tool.  Should not be too difficult to fix.

taylor-bsg:
Thanks Sripathi, I chatted with Tommy and there is a subtlety regarding the
pc_n register — on return from interrupt it should be restored from the EPC
so that if an interrupt handle changes EPC in the handler, upon return from
interrupt and an immediate interrupt after that, it saves the correct
(modified) EPC. An example test is a threading package that alternates
between two threads by using async interrupts and in the interrupt handler,
switching the EPC (and the GPRs) between the two threads to switch back and
forth. Perhaps you could code that up as a test, and combine with the trace
interrupt to see if you can trace one thread and not the other and get the
exact right sequence of PC’s?

tommydcjung:
Nice, can we separate this into two PRs? one for all the test suites, and another for coverage stuff? and make them merge to ci_bigblade bramch.

software/spmd/interrupt_tests/trace_remote_load_loop_test.S
```diff
@@ -0,0 +1,139 @@
...
+  fcvt.s.w f22,x0
+  fcvt.s.w f23,x0
+  fcvt.s.w f24,x0
+  fcvt.s.w f25,x0
+  fcvt.s.w f26,x0
+  fcvt.s.w f27,x0
+  fcvt.s.w f28,x0
+  fcvt.s.w f29,x0
+  fcvt.s.w f30,x0
+  fcvt.s.w f31,x0
```
tommydcjung, line 87: maybe we can make this a macro and put it in  "bsg_manycore_asm.h"?

testbenches/common/v/bsg_nonsynth_manycore_testbench.v
```diff
@@ -652,5 +652,13 @@ module bsg_nonsynth_manycore_testbench
...
   );
 
+  // Functional Coverage
+  bind vanilla_core 
+    bsg_nonsynth_manycore_vanilla_core_pc_cov
+    #(.icache_tag_width_p(icache_tag_width_p)
+     ,.icache_entries_p(icache_entries_p)
+     )
+     pc_cov
+     (.*);
```
tommydcjung, line 662: can we add  a parameter "enable_coverage_p", and wrap this around? by default, we don't want to have coverage on.

tommydcjung:
@sripathi-muralitharan I create a fix for npc_r/mret problem in this PR. I think it would be nice to create a test that exposes this bug before we can merge it.

sripathi-muralitharan:
Sure. I can write a test for that as well!

tommydcjung:
https://github.com/bespoke-silicon-group/bsg_manycore/pull/463

software/spmd/interrupt_tests/trace_remote_load_loop_test.S
sripathi-muralitharan: Addressed in #468

testbenches/common/v/bsg_nonsynth_manycore_testbench.v
sripathi-muralitharan: Addressed in #469

sripathi-muralitharan:
Opened two new PRs #468 , #469 as per Tommy's request (GitHub doesn't seem to allow modifying the current PRs base branches). I have addressed some of the review comments in these PRs. I am still working on the tests. 

The maintainer can choose to keep this PR open or close it. I will be using my fork's `bigblade_interrupt` branch for my own changes before pushing to the `ci_bigblade_*` branches , so there are going to be commits here from time to time.

software/spmd/interrupt_tests/dual_threading_test.S
```diff
@@ -0,0 +1,117 @@
+// this program creates a tiny threading package. The threads communicate with each other using remote interruts.
```
sripathi-muralitharan, line 1: @taylor-bsg is this what you were referring to in the threading test? The vanilla core is able to switch between the two threads and exit the handler correctly to the alternate thread. I have enabled trace for both of them and I can see that the PCs are being traced correctly for each thread.

sripathi-muralitharan: What did you exactly mean by the following?

> Perhaps you could code that up as a test, and combine with the trace
> interrupt to see if you can trace one thread and not the other and get the
> exact right sequence of PC’s?

software/spmd/interrupt_tests/dual_threading_test.S
```diff
@@ -0,0 +1,117 @@
...
+  li x1, 0x30000
+  csrrw x0, mie, x1
+
+// t1 is a +10 loop
+t1_entry:
+  beq x0, s2, t1_done
+  addi s3, s3, 10
+  addi s2, s2, -1
+  sw t2, 0(t1)
+  nop
```
taylor-bsg, line 84: comment explaining why these nops are here and also in the other spot of the program. Are they necessary? Are you reducing coverage by putting noops here?

software/spmd/interrupt_tests/dual_threading_test.S
taylor-bsg: This is good. But maybe have some kind of verification on the trace interrupt itself? Perhaps have it increment a counter based on which thread it was tracing in, which you check against the correct answer?

dpetrisko:
@tommydcjung same here: https://github.com/bespoke-silicon-group/bsg_manycore/pull/477#issuecomment-880237412

sripathi-muralitharan:
I think we can close this PR. It was superseded by #468 which is already merged in.

PR #463, opened 2021-02-01
npc mret fix

tommydcjung:


v/vanilla_bean/vanilla_core.v
```diff
@@ -1418,7 +1421,7 @@ module vanilla_core
       npc_write_en = 1'b0;
     end
     else begin
-      npc_write_en = exe_r.valid & mstatus_r.mie;
+      npc_write_en = (exe_r.valid & mstatus_r.mie) | exe_r.decode.is_mret_op;
```
taylor-bsg, line 1424: I assume you want to check that the is_mret_op is valid as well...

tommydcjung: You can't have exe_r.decode.is_mret_op without exe_r.valid so it doesn't make difference in the end.

taylor-bsg:
Okay, I see, you are saying that the decode bits are implicitly already
gated with valid; that you have the invariant that  exe_r.valid |
(~|exe_r.decode)?

PR #468, opened 2021-02-04
CI Bigblade Interrupt Tests

sripathi-muralitharan:
This is a tinier version of #460 consisting of only the different tests for interrupts, a regression suite for testing and the Makefile flow for it.


software/bsg_manycore_lib/bsg_manycore_asm.h
```diff
@@ -99,4 +99,71 @@
     li t0, val;             \
     op reg,t0,dest;
 
+// start code
+#define bsg_asm_start  \
```
tommydcjung, line 103: can we rename it to something more descriptive? e.g. bsg_asm_init_regfile

sripathi-muralitharan: Done

tommydcjung:
Are these ready to merge or are there more tests coming?

sripathi-muralitharan:
We can leave this open. I have a couple more tests coming.

sripathi-muralitharan:
Added a test for the `npc mret` problem (#463). With this and the `interrupt_tests/dual_handler_test` we can expose the bug more clearly. 

The `interrupt_tests/dual_handler_test` would stall without the fix and looking at the output log messages which reports the kind of interrupt and the MEPC, you can see that after exiting the first remote interrupt it returns back to the trace interrupt handler rather than the code. 

The `interrupt_tests/dual_npc_mret_test` would go into a infinite loop without the fix because interrupts are called within interrupts.

Additionally, I think it is sufficient to have `npc_write_en = exe_r.valid & mstatus_r.mie` in #463 and we don't need to OR this with `exe_r.decode.is_mret_op` since the CSR completes re-enabling interrupts only in the next pipeline stage which is when interrupts can be taken. At this time there is a different instruction in the pipeline and the exe_r pipeline register has changed. So I think we will not hit that condition (unless there are traps to M-mode for things other than interrupts).

sripathi-muralitharan:
In hindsight, I'll take back what I said about the `npc_write_en`. We do need the fix in the PR because of the case if someone modifies the MEPC register in the interrupt handler.

sripathi-muralitharan:
I am thinking of adding a couple more tests exercising the JALR and trace/remote interrupts. I can maybe make a new PR for those later?

tommydcjung:
Could you add a short README.md in spmd/interrupt_tests to describe what commands are the users supposed to run the regression and check if this passed or not?

software/spmd/interrupt_tests/Makefile
```diff
@@ -0,0 +1,51 @@
+find-dir-with = $(shell /usr/bin/perl -e 'chomp($$_ = `pwd`); while ($$_ ne "" && ! -e "$$_/$(1)") { m:(.*)/[^/]+/??:; $$_ = $$1; } print;')
```
tommydcjung, line 1: `BSG_MANYCORE_DIR := $(shell git rev-parse --show-toplevel)` might be simpler?

tommydcjung (approved):

software/spmd/interrupt_tests/Makefile
sripathi-muralitharan: Yup. Change is reflected in the recent commit.

sripathi-muralitharan:
Added a README file

PR #470, opened 2021-02-04
Adds an explicit ENTRY command to linker script

mrutt92:
Sets the entry word in the RISCV header to the address of '_start'

As of now, the default entry address is set to 0x0.


tommydcjung:
As we discussed, the entry can be set using `-Wl,-eentry` without requiring any change in linker.

mrutt92:
yup - the default location will remain zero unless specified o/w on the command line.

taylor-bsg:
what's the story on this pull request? Do we merge or delete?

mrutt92:
let's close. we opted to not set ENTRY universally.

PR #476, opened 2021-02-19
Atomic Add support 

sripathi-muralitharan:
The manycore currently supports only AMOSWAP and AMOOR operations. This PR adds AMOADD support as well. 

Correspondly macros for software use in `bsg_manycore_atomic.h` have been added, an update to the `vcache_atomics` test and a multicore test exercising the use of atomic adds.

### DIFF
--- a/software/spmd/amoadd_test/Makefile
+++ b/software/spmd/amoadd_test/Makefile
@@ -1,7 +1,8 @@
 export BSG_MANYCORE_DIR := $(shell git rev-parse --show-toplevel)
 
-bsg_tiles_X = 4
-bsg_tiles_Y = 4
+# Running tests on full manycore array. Uncomment and modify for a smaller array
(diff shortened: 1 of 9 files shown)


software/spmd/amoadd_test/Makefile
```diff
@@ -0,0 +1,16 @@
+export BSG_MANYCORE_DIR := $(shell git rev-parse --show-toplevel)
+
+bsg_tiles_X = 4
+bsg_tiles_Y = 4
```
tommydcjung, line 4: Can we make it so that it uses the full manycore array to test? You would just need to comment out these lines.

sripathi-muralitharan: Sure! Reflected in the latest commit.

tommydcjung (approved):

PR #477, opened 2021-02-21
PC Profiler

sripathi-muralitharan:
This PR adds a software-based PC Profiler that uses the trace interrupt handler to build a PC histogram in DRAM.

There are 2 main files - 
1. profiler.S --> This is a C runtime file that does most of the initialization but in addition has the trace interrupt handler defined that updates the histogram. The profiler (i.e trace interrupt handler) uses the first 2 entries in DMEM to spill any registers before performing any action. It computes the index into the histogram based on the EPC and then atomically increments that location in memory. This depends on #476 to work correctly. Each PC has a 4 byte bin assigned to it at HISTOGRAM_BASE_ADDR + EPC.

2. main.c --> This contains a C program that sets up the tile and computes the first 10 fibonacci numbers (just a sample, could ideally be anything you want here). This program gets profiled by the profiler. It also includes a function that reads out this histogram and prints the data to the console and compares this against a statically generated expectation of what the number of hits for that PC should be. 

Testing: 
I tested this to work with a histogram array of size 4KB (simulation was taking too long at 16MB even for 1 tile) and used 16 tiles. There is a parameter that can be configured to alter the histogram array size as well. 

PS: This shows only a first, preliminary version of what the profiler should look like. I have listed some questions I had (there are probably more things I may not have thought about) about different parts of this exercise in the comments near the relevant piece of code (which will be removed before merge once answered).

### DIFF
--- a/software/spmd/pc_profiler/main.c
+++ b/software/spmd/pc_profiler/main.c
@@ -32,7 +32,6 @@ void fibonacci()
 
 void read_histogram()
 {
-  // Fixme: This array is populated statically by observing disassembly. Need to make it standalone
   int expectation[10] = {1 * NUM_TILES, 1 * NUM_TILES, 10 * NUM_TILES, 10 * NUM_TILES, 10 * NUM_TILES, 10 * NUM_TILES, 10 * NUM_TILES, 1 * NUM_TILES, 1 * NUM_TILES, 1 * NUM_TILES};
 
   int *i;
--- a/software/spmd/pc_profiler/main.c
+++ b/software/spmd/pc_profiler/main.c
@@ -3,11 +3,7 @@
 #include "bsg_manycore.h"
 #include "bsg_set_tile_x_y.h"
 
-// Histogram location?
-// 1. Can histogram base address be accessed directly from symbol table?
-// 2. Should linker be modified so that histogram is statically assigned a region in memory?
-// Fixme: Histogram base address adjusted to the right value after looking at disassembly
-#define HISTOGRAM_BASE_ADDR 0x81000290
+extern int _histogram_arr;
 
 #ifndef HIST_SPACE
 #define HIST_SPACE 16777216
@@ -36,14 +32,13 @@ void fibonacci()
 
 void read_histogram()
 {
-  int val;
-  int count = 0;
   // Fixme: This array is populated statically by observing disassembly. Need to make it standalone
   int expectation[10] = {1 * NUM_TILES, 1 * NUM_TILES, 10 * NUM_TILES, 10 * NUM_TILES, 10 * NUM_TILES, 10 * NUM_TILES, 10 * NUM_TILES, 1 * NUM_TILES, 1 * NUM_TILES, 1 * NUM_TILES};
-  // Adjust start value based on disassembly since function is small and therefore inlined with O2 optimization
-  int *i;
 
-  for (i = (HISTOGRAM_BASE_ADDR); i < (HISTOGRAM_BASE_ADDR + HIST_SPACE); i++)
+  int *i;
+  int val;
+  int count = 0;
+  for (i = &_histogram_arr; i < (&_histogram_arr + (HIST_SPACE >> 2)); i++)
   {
     val = *i;
     if (val != 0)
@@ -52,8 +47,7 @@ void read_histogram()
       if (expectation[count] != val)
         bsg_fail();
       count++;
-    }
-      
+    }    
   }
 
   bsg_finish();
--- a/software/spmd/pc_profiler/main.c
+++ b/software/spmd/pc_profiler/main.c
@@ -1,0 +1,90 @@
+// Program to print 10 fibonacci numbers; A sample program that will get profiled
+
+#include "bsg_manycore.h"
+#include "bsg_set_tile_x_y.h"
+
+// Histogram location?
+// 1. Can histogram base address be accessed directly from symbol table?
+// 2. Should linker be modified so that histogram is statically assigned a region in memory?
+// Fixme: Histogram base address adjusted to the right value after looking at disassembly
+#define HISTOGRAM_BASE_ADDR 0x81000290
+
+#ifndef HIST_SPACE
+#define HIST_SPACE 16777216
+#endif
+
+#define NUM_TILES bsg_group_size
+
+// Sample program to print 10 fibonacci numbers
+void fibonacci()
+{
+  int i;
+  int a = 0;
+  int b = 1;
+  int temp = 0;
+
+  for (i = 0; i < 10; i++)
+  {
+    temp = a + b;
+    a = b;
+    b = temp;
+  }
+
+  if (a != 55)
+    bsg_fail();
+}
+
+void read_histogram()
+{
+  int val;
+  int count = 0;
+  // Fixme: This array is populated statically by observing disassembly. Need to make it standalone
+  int expectation[10] = {1 * NUM_TILES, 1 * NUM_TILES, 10 * NUM_TILES, 10 * NUM_TILES, 10 * NUM_TILES, 10 * NUM_TILES, 10 * NUM_TILES, 1 * NUM_TILES, 1 * NUM_TILES, 1 * NUM_TILES};
+  // Adjust start value based on disassembly since function is small and therefore inlined with O2 optimization
+  int *i;
+
+  for (i = (HISTOGRAM_BASE_ADDR); i < (HISTOGRAM_BASE_ADDR + HIST_SPACE); i++)
+  {
+    val = *i;
+    if (val != 0)
+    {
+      bsg_printf("%p: %x\n", i, val);
+      if (expectation[count] != val)
+        bsg_fail();
+      count++;
+    }
+      
+  }
+
+  bsg_finish();
+}
+
+void main()
+{
+  // tile setup
+  bsg_set_tile_x_y();
+
+
+  // Start profiling code, enable interrupts
+  int mie, mstatus;
+  mie = 0x20000;
+  mstatus = 0x8;
+  __asm__ __volatile__ ("csrw mie, %0": : "r" (mie));
+  __asm__ __volatile__ ("csrw mstatus, %0" : : "r" (mstatus));
+
+  // Code to be profiled
+  fibonacci();
+
+  // Reading histogram, disable interrupts
+  mie = mstatus = 0;
+  __asm__ __volatile__ ("csrw mie, %0": : "r" (mie));
+  __asm__ __volatile__ ("csrw mstatus, %0" : : "r" (mstatus));
+
+  if (__bsg_id == 0)
+  {
+    // read histogram
+    read_histogram();
+  }
+  
+  bsg_wait_while(1);
+}
--- a/software/spmd/pc_profiler/main.c
+++ b/software/spmd/pc_profiler/main.c
@@ -60,19 +60,13 @@ void main()
 
 
   // Start profiling code, enable interrupts
-  int mie, mstatus;
-  mie = 0x20000;
-  mstatus = 0x8;
-  __asm__ __volatile__ ("csrw mie, %0": : "r" (mie));
-  __asm__ __volatile__ ("csrw mstatus, %0" : : "r" (mstatus));
+  bsg_pc_profiler_start();
 
   // Code to be profiled
   fibonacci();
 
   // Reading histogram, disable interrupts
-  mie = mstatus = 0;
-  __asm__ __volatile__ ("csrw mie, %0": : "r" (mie));
-  __asm__ __volatile__ ("csrw mstatus, %0" : : "r" (mstatus));
+  bsg_pc_profiler_end();
 
   if (__bsg_id == 0)
   {
--- a/software/spmd/pc_profiler/profiler.S
+++ b/software/spmd/pc_profiler/profiler.S
@@ -10,6 +10,7 @@
   The profiler uses the following registers - 
   t0 --> Base address of the histogram array (actual value stored in dmem[1])
   t1 --> Scratchpad (actual value stored in dmem[0])
+  t2 --> Scratchpad (actual value stored in dmem[2])
 */
 
 #include "bsg_manycore_arch.h"
@@ -31,9 +32,10 @@
   Defines a 2 element (8 byte) array to store the base pointer of the histogram array and for use by the interrupt handler to spill some registers
   dmem[0] = histogram base pointer, temporary storage
   dmem[1] = temporary storage
+  dmem[2] = temporary storage
 */
 .section .dmem.interrupt, "aw"
-  _interrupt_arr: .space 8, 0x0
+  _interrupt_arr: .space 12, 0x0
 
 /*
   DRAM section:
@@ -85,31 +87,30 @@ _trace_interrupt:
   lw t0, 0(zero)
   // Store another register to dmem[0]
   sw t1, 0(zero)
+  // Store a register to dmem[2]
+  sw t2, 8(zero)
 
   // read the mepc register
   csrr t1, mepc
   // find the correct index in the histogram array
-  add t0, t0, t1
+  add t1, t0, t1
 
   // atomically increment the value at the computed index
-  li t1, 1
+  li t2, 1
   // We don't care about the result hence assigned to zero register
-  amoadd.w.aq zero, t1, (t0)
-
-  // remap t0 to histogram array base address
-  csrr t1, mepc
-  sub t0, t0, t1
+  amoadd.w zero, t2, (t1)
 
   // Restore registers and store base ptr back to dmem[0]
   lw t1, 0(zero)
   sw t0, 0(zero)
   lw t0, 4(zero)
+  lw t2, 8(zero)
 
   mret
 
 _remote_interrupt_handler:
   // Add code here when using in actual programs, for now these are turned off
-  j fail
+  bsg_asm_fail(IO_X_INDEX, 0)
 #endif
 
 /*
@@ -226,7 +227,3 @@ _start:
 2:
   // Should never reach this point
   j 2b
-
-  // Must not come here
-fail:
-  j fail
--- a/software/spmd/pc_profiler/profiler.S
+++ b/software/spmd/pc_profiler/profiler.S
@@ -44,6 +44,7 @@
   2. Base pointer in DRAM (for now it is mapped to start of DRAM) [Consider static assignment in linker]
 */
 .section .dram, "aw"
+.globl _histogram_arr
   _histogram_arr: .space HIST_SPACE, 0x0
 #endif
 
--- a/software/spmd/pc_profiler/profiler.S
+++ b/software/spmd/pc_profiler/profiler.S
@@ -9,8 +9,7 @@
   The program being profiled is free to use any registers within if there are no other imposed constraints.
   The profiler uses the following registers - 
   t0 --> Base address of the histogram array (actual value stored in dmem[1])
-  t1 --> Scratchpad (actual value stored in dmem[0])
-  t2 --> Scratchpad (actual value stored in dmem[2])
+  t1 --> Scratchpad (actual value stored in dmem[2])
 */
 
 #include "bsg_manycore_arch.h"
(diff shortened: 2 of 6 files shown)


software/spmd/pc_profiler/profiler.S
```diff
@@ -0,0 +1,231 @@
...
+  li t0, 0x20000
+  csrrc zero, mip, t0
+
+  // load address of histogram array
+  lw t0, 0(zero)
+  // Store another register to dmem[0]
+  sw t1, 0(zero)
+
+  // read the mepc register
+  csrr t1, mepc
```
dpetrisko, line 89: code golfing, but you could csrrw t1, mepc. Then you wouldn't need to sub t0, t0, t1 down blow

dpetrisko, line 89: Shouldn't we index with mepc >> 2? This will be sparse otherwise

software/spmd/pc_profiler/profiler.S
```diff
@@ -0,0 +1,231 @@
...
+*/
+_remote_interrupt:
+  j _remote_interrupt_handler
+
+/*
+  Trace interrupt handler (should map to address 4)
+  Clears the pending trace interrupt bit in the MIP CSR and atomically increments the histogram bin count based on the MEPC CSR value
+
+  The bin index in the histogram can be found using the following equation index = base ptr + EPC
+  (Although EPC is 4-byte aligned, we are assigning 4 bytes per bin, so we don't need to do EPC >> 2)
```
dpetrisko, line 69: nice!

software/spmd/pc_profiler/profiler.S
```diff
@@ -0,0 +1,232 @@
...
+  csrr t1, mepc
+  // find the correct index in the histogram array
+  add t0, t0, t1
+
+  // atomically increment the value at the computed index
+  li t1, 1
+  // We don't care about the result hence assigned to zero register
+  amoadd.w.aq zero, t1, (t0)
+
+  // remap t0 to histogram array base address
```
taylor-bsg, line 99: rather than rematerializing the base address, so you can store it back, I suggest that you instead just use one more word of data memory for your second temporary. you reduce icache footprint by 3 instructions, and increase tracing performance by 3/(15+stall cycles (6?)).

software/spmd/pc_profiler/profiler.S
```diff
@@ -0,0 +1,232 @@
...
+  Since we are using the MEPC, we don't have the PC of the first instruction in the program. 
+  (Probably does not matter because it's just 1 datapoint)
+*/
+_trace_interrupt:
+  // Save a register to dmem[1]
+  sw t0, 4(zero)
+
+  // clear mip.trace
+  li t0, 0x20000
+  csrrc zero, mip, t0
```
taylor-bsg, line 82: @tommydcjung suggest putting mip.trace at bit 0 of mip register, which I believe is unused, so we can just use a csrrci instruction to clear this; this will speed up the interrupt routine and reduce the cache foot print.

software/spmd/pc_profiler/profiler.S
```diff
@@ -0,0 +1,232 @@
...
+
+  // read the mepc register
+  csrr t1, mepc
+  // find the correct index in the histogram array
+  add t0, t0, t1
+
+  // atomically increment the value at the computed index
+  li t1, 1
+  // We don't care about the result hence assigned to zero register
+  amoadd.w.aq zero, t1, (t0)
```
taylor-bsg, line 97: We should have a justification for which combination of .aq and .rl we use. My opinion here is that we don't need either .aq or .rl because we are not using the address as a lock, but as a histogram, and that this is implicitly doing a memory fence after we do the amoadd operation, which is possibly responsible for the slow down you are seeing.

taylor-bsg (commented):
Super cool! See feedback.

software/spmd/pc_profiler/profiler.S
```diff
@@ -0,0 +1,232 @@
...
+
+/*
+  Trace interrupt handler (should map to address 4)
+  Clears the pending trace interrupt bit in the MIP CSR and atomically increments the histogram bin count based on the MEPC CSR value
+
+  The bin index in the histogram can be found using the following equation index = base ptr + EPC
+  (Although EPC is 4-byte aligned, we are assigning 4 bytes per bin, so we don't need to do EPC >> 2)
+
+  Lingering concerns:
+  Since we are using the MEPC, we don't have the PC of the first instruction in the program.
```
taylor-bsg, line 73: Maybe provide more detail about this? I see the instruction where you enable instructions, and then you jump to the code to trace. Presumably the jump instruction is not profiled, but the instructions after the jump are?

software/spmd/pc_profiler/profiler.S
dpetrisko: *puts on risc-v hat* that’s the user-mode software interrupt bit, would be a non-compliant. Probably fine since you’re not going to run OpenSBI or anything, but should note it somewhere.

software/spmd/pc_profiler/main.c
```diff
@@ -0,0 +1,84 @@
...
+  // tile setup
+  bsg_set_tile_x_y();
+
+
+  // Start profiling code, enable interrupts
+  int mie, mstatus;
+  mie = 0x20000;
+  mstatus = 0x8;
+  __asm__ __volatile__ ("csrw mie, %0": : "r" (mie));
+  __asm__ __volatile__ ("csrw mstatus, %0" : : "r" (mstatus));
```
tommydcjung, line 67: Can we make this into a macro? (e.g. bsg_manycore_pc_profiler_start(), bsg_manycore_pc_profiler_end())

tommydcjung:
Nice. Is the idea that if you want to have pc_profiler enabled, you would link profiler.o instead of crt.o?

software/spmd/pc_profiler/main.c
dpetrisko: if you do pull it out, should do csrs/c instead of csrw

dpetrisko:
>Nice. Is the idea that if you want to have pc_profiler enabled, you would link profiler.o instead of crt.o?

If you took out the start definition (and put it in the test folder) and made the handlers weak symbols, I think you could link profiler.o in addition to crt.o rather than replacing it, which is a little more flexible

software/spmd/pc_profiler/profiler.S
```diff
@@ -0,0 +1,232 @@
...
+  j main
+#endif
+
+2:
+  // Should never reach this point
+  j 2b
+
+  // Must not come here
+fail:
+  j fail
```
tommydcjung, line 232: maybe send bsg_fail packet?

tommydcjung:
> > Nice. Is the idea that if you want to have pc_profiler enabled, you would link profiler.o instead of crt.o?
> 
> If you took out the start definition (and put it in the test folder) and made the handlers weak symbols, I think you could link profiler.o in addition to crt.o rather than replacing it, which is a little more flexible

I'm trying to understand if this is just a proof-of-concept code, or something that can be actually used for existing cuda-lite programs. Presumably, you would also need some ways to link the remote interrupt handler, instead of just jumping to fail.

software/spmd/pc_profiler/main.c
```diff
@@ -0,0 +1,90 @@
+// Program to print 10 fibonacci numbers; A sample program that will get profiled
+
+#include "bsg_manycore.h"
+#include "bsg_set_tile_x_y.h"
+
+// Histogram location?
+// 1. Can histogram base address be accessed directly from symbol table?
```
mrutt92: I would think so.

software/spmd/pc_profiler/profiler.S
```diff
@@ -0,0 +1,231 @@
...
+  Clears the pending trace interrupt bit in the MIP CSR and atomically increments the histogram bin count based on the MEPC CSR value
+
+  The bin index in the histogram can be found using the following equation index = base ptr + EPC
+  (Although EPC is 4-byte aligned, we are assigning 4 bytes per bin, so we don't need to do EPC >> 2)
+
+  Lingering concerns:
+  Since we are using the MEPC, we don't have the PC of the first instruction in the program. 
+  (Probably does not matter because it's just 1 datapoint)
+*/
+_trace_interrupt:
```
mrutt92, line 76: So are both address 0x4 and 0x0 special addresses now? So specific interrupt handlers live at special code addresses?

software/spmd/pc_profiler/profiler.S
sripathi-muralitharan: Isn't that going away from the spec that we could only use the first 2 entries in the dmem for interrupts? If it's not a strict requirement then that's great!

Just out of curiosity, how did you compute the divisor for the expression you mentioned?

software/spmd/pc_profiler/profiler.S
sripathi-muralitharan: Agreeing with Dan from the RISC-V compatibility standpoint.

software/spmd/pc_profiler/profiler.S
sripathi-muralitharan: That makes a lot of sense. Without the .aq one can just send the amoadd and forget about it since the return register is zero (Need to confirm though if there are no return packets when the rd register is the zero register)

software/spmd/pc_profiler/profiler.S
sripathi-muralitharan: Yeah, this comment might probably be stale but might still have relevance for smaller programs. With the O2 optimization enabled, smaller functions are actually inlined which means there aren't going to be any jumps in the code. This would result in the very first instruction in the program to not be profiled. 

Two somewhat related question however are - 
1. How small does the function need to be for it to be inlined and what are the typical program sizes for real workloads in the manycore?
2. Even though the compiler inlined the function itself, there still seems to be a dedicated portion of the disassembly that duplicates the function code which increases the code size. Is this something we should care about or am I missing some compiler option here?

software/spmd/pc_profiler/main.c
sripathi-muralitharan: @dpetrisko why is that necessary? Is it because we don't want to disturb the other interrupt bits?

software/spmd/pc_profiler/main.c
sripathi-muralitharan: Yup, found a way to do that. Reflected in the latest commit.

software/spmd/pc_profiler/profiler.S
sripathi-muralitharan: Yup. They should be as per the interrupt specification that was given. (Not able to share the doc here since I am not a owner, but if you have access to a common drive where all the HB docs are present, look for the file - HammerBlade Machine Interrupts. That should give all the interrupt specific info)

sripathi-muralitharan:
> > > Nice. Is the idea that if you want to have pc_profiler enabled, you would link profiler.o instead of crt.o?
> > 
> > If you took out the start definition (and put it in the test folder) and made the handlers weak symbols, I think you could link profiler.o in addition to crt.o rather than replacing it, which is a little more flexible
> 
> I'm trying to understand if this is just a proof-of-concept code, or something that can be actually used for existing cuda-lite programs. Presumably, you would also need some ways to link the remote interrupt handler, instead of just jumping to fail.

@tommy @dpetrisko I believe this code is still at the proof-of-concept stage and I can imagine there is still quite a bit to do to be worthy of integrating with the cuda-lite library (for example, tommy's example with the remote interrupts). But I was thinking we can maybe create a new CRT altogether and use ifdefs to turn on/off profiling? But Dan's idea sounds much better.. 

Somewhat unrelated: Is there any use case for trace interrupts apart from profiling? In that case maybe we have to think about having multiple versions of this handler (???)

software/spmd/pc_profiler/main.c
dpetrisko: Yes. Imagine you have a remote interrupt pending. Now enable tracer. Oops

software/spmd/pc_profiler/profiler.S
dpetrisko: I know there’s a slippery slope argument, but maybe just go to four right now so that when we add a fast remote interrupt handler we have a little more breathing room?

software/spmd/pc_profiler/profiler.S
dpetrisko: https://docs.google.com/document/d/1HKa3Qts55TLqltLKpoAalcVJokGHOQm27u-1FXmbTQE/edit?ts=60075876

@mrutt92 high-level: when you take a trap, you go to BASE + 4*CAUSE. We hardcode BASE to -4*TRACE_CAUSE so that the handler pointers get placed at the bottom of DMEM

software/spmd/pc_profiler/profiler.S
dpetrisko: To clarify, I'm on the side of changing it and adding to a list of risc-v incompatibilities somewhere.  But it's a hardware change, so separate PR

tommydcjung:
I think you could use trace interrupt, for step by step debugging? I think you would want to be able to swap out different kinds of trace interrupt hander just as you would for the remote interrupt handler.

dpetrisko:
Yeah exactly what Tommy is saying. that’s why I’m suggesting looking at this profile code like a standalone handler to be linked in when you need it, not to try to make one crt to rule them all. 

Let the cuda lite folks figure out how they want to link it in for the general case

software/spmd/pc_profiler/profiler.S
mrutt92: You would want the aq if you require the amoadd to complete before any following memory operation in the program. That's not the case here.

software/spmd/pc_profiler/profiler.S
mrutt92: I am curious why we are hardcoding that address for interrupts? Does the hardware treat those locations differently? Is it just we can avoid using the extra instructions to load relocatables?

software/spmd/pc_profiler/profiler.S
mrutt92: So are there only two causes? Shouldn't this `_trace_interrupt` be one instruction to jump to another location like with `_remote_interrupt`? Or is this only meant for a specific test?

software/spmd/pc_profiler/profiler.S
mrutt92: It's not just how small a function is, but how many times it's used. A medium sized function can be inlined if it's static and only used once.

Which function? If you do not mark the function as `static` the compiler has to generate the actual function even if it inlines it. This is so the function can be called from another source file.

software/spmd/pc_profiler/profiler.S
mrutt92: @dpetrisko am I misunderstanding something? Do you mean DMEM or IMEM in this case? I see that we've also hardcoded DMEM addresses for interrupt handler storage, but I think you're talking about PC addresses.

dpetrisko: Disregard, I misread your question

mrutt92:
Why not just make CRT to rule them all? Having one crt that you link in when you want tracing vs no profiling vs some other interrupt could turn into a headache. It all seems avoidable by not moving much of this code into its own source file.

dpetrisko:
>Why not just make CRT to rule them all?

The cuda lite folks have figured out how they want to link it in for the general case

software/spmd/pc_profiler/profiler.S
sripathi-muralitharan: The function I'm talking about is the fibonacci function in `main.c`. There is only a call to it from `main()`. Thanks for the explanation though, that clears stuff up for me.

software/spmd/pc_profiler/profiler.S
sripathi-muralitharan: Right now, the hardware expects interrupt handlers at these addresses as per https://github.com/bespoke-silicon-group/bsg_manycore/blob/43ba88eadbaebc7a446fcb60077dece3aaea3f66/v/vanilla_bean/bsg_vanilla_pkg.v#L323

Maybe @tommy could throw some light why these need to be hardcoded?

mrutt92: So there seems to be two things that are hardcoded:
1. The PC address that is set when an interrupt is triggered
2. The private data in DMEM that is reserved for the interrupt handler's use

I understand why (1) - I'm a little less clear on (2).

tommydcjung: interrupt jump address can be hardcoded, because it's good enough to do what we want, and area-efficient. For reserved space in DMEM for interrupt handler, there is nothing in hardware that hardcodes that. But we put it there, so that programmer can access it using x0 as base address (e.g. sw x3, 0(x0) or sw x3, 4(x0))

mrutt92: ok - so (2) is for efficient addressing of the interrupt data.

dpetrisko: Yeah normally this is done using a swap with a dedicated scratch CSR and interrupt stack in memory.

 This approach is a scratch register and 1-2 address stack (located in DMEM)

software/spmd/pc_profiler/profiler.S
sripathi-muralitharan: lol, thanks!

software/spmd/pc_profiler/profiler.S
taylor-bsg: Yeah, I visited this a while back from my consulting days. Basically, there are really two sweet spots in compatibility that matter: full toolchain compatibility and full OS-level compatibility. We are the former. Partial OS-level compatibility does leaves everybody complaining about what you are missing.

software/spmd/pc_profiler/profiler.S
```diff
@@ -0,0 +1,229 @@
...
+  sw t0, 4(zero)
+
+  // clear mip.trace
+  li t0, 0x20000
+  csrrc zero, mip, t0
+
+  // load address of histogram array
+  lw t0, 0(zero)
+  // Store another register to dmem[0]
+  sw t1, 0(zero)
```
taylor-bsg, line 89: Unnecessary store. You only need two temporary registers for this code.

software/spmd/pc_profiler/profiler.S
```diff
@@ -0,0 +1,229 @@
...
+  // find the correct index in the histogram array
+  add t1, t0, t1
+
+  // atomically increment the value at the computed index
+  li t2, 1
+  // We don't care about the result hence assigned to zero register
+  amoadd.w zero, t2, (t1)
+
+  // Restore registers and store base ptr back to dmem[0]
+  lw t1, 0(zero)
```
taylor-bsg, line 104: these two instructions also unnecessary

software/spmd/pc_profiler/profiler.S
dpetrisko: Agreed, but I would phrase it a little differently. 

- The manycore is mostly compatible with RISC-V unprivileged mode (LR/SC behaves differently than expected, other than that I think it is). This is fine because the compiler won't insert LR/SC automatically.
- The manycore has a custom privileged ISA. (missing ebreak, ecall, missing most CSRs, repurposed fields in CSRs). This is fine and expected because these aren't Unix capable cores, which is what the vanilla RISC-V privileged spec is designed for.

dpetrisko:
@tommydcjung Sripathi is leaving imminently, so if there's anything you need from him to get this merged in, make sure to ask in the next few days!

sripathi-muralitharan:
@tommydcjung could you please update the target branch to master? Would be easier to view the diffs. I don't have the permissions to do it.

tommydcjung:
Can you fix the pc_profiler/ so that I can run it by type "make" like all the other regressions?

sripathi-muralitharan:
I think you should be able to run with just "make" now. You must additionally say PROFILE=1 and HIST_SPACE=4096. The second variable is set just to make simulation go faster. Otherwise, most of the time will be spent in loading 16MB of the DRAM with zeros.

tommydcjung (approved):

PR #479, opened 2021-02-28
Endpoint standard refactor / Credit limit CSR

tommydcjung:


v/vanilla_bean/mcsr.v
```diff
@@ -15,6 +15,8 @@ module mcsr
   #(parameter reg_addr_width_lp = RV32_reg_addr_width_gp
     , parameter reg_data_width_lp = RV32_reg_data_width_gp
     , parameter pc_width_p="inv"
+    , parameter credit_limit_default_val_p = 32
+    , parameter credit_counter_width_p=`BSG_WIDTH(32)
```
dpetrisko, line 19: `BSG_WIDTH(credit_limit_default_val_p)?

tommydcjung: No, the default val can be set lower than what the credit counter width allows.

dpetrisko: ah gotcha

v/bsg_manycore_endpoint_standard.v
```diff
@@ -94,7 +97,7 @@ module bsg_manycore_endpoint_standard
     , output                                   returned_credit_v_r_o
     , output [bsg_manycore_reg_id_width_gp-1:0]  returned_credit_reg_id_r_o
 
-    , output [$clog2(max_out_credits_p+1)-1:0] out_credits_o
+    , output [credit_counter_width_p-1:0] out_credits_o
```
taylor-bsg, line 100: // This holds the number of credits that we are expecting to (eventually) come back to us.
// Note: Prior to March 2021, this used to hold the number of credits
// that were currently available to send. This change allows the clients to have more control
// over their credits.

v/bsg_manycore_endpoint_standard.v
```diff
@@ -131,108 +134,73 @@ module bsg_manycore_endpoint_standard
...
-    ,.comb_returning_data_o(comb_returning_data_lo)
-    ,.comb_returning_v_o(comb_returning_v_lo)
-  );
 
+  // When a request is dequeued, the return info (src coord, response type, etc) is stored in this module.
+  // accelerators can't dequeue another request without returning the response for the already dequeued request.
+  
 
-  // we hide the request if the returning path is not ready
+  // credit counting on response fifo
```
taylor-bsg, line 168: seems like this is a bsg_counter_up_down?

v/bsg_manycore_endpoint_standard.v
taylor-bsg: Maybe rename out_credits_o to out_credits_used_o so that it is more clear that we have broken the interface relative to prior code.

v/bsg_manycore_endpoint_standard.v
```diff
@@ -241,206 +209,205 @@ module bsg_manycore_endpoint_standard
...
-  wire load_store_ready = holded_returning_v_lo;
+  // synopsys translate_off
+  always_ff @ (negedge clk_i) begin
+    if (~reset_i) begin
+      assert(~lock_v_r | ~returning_v_i)
+        else $error("[BSG_ERROR] lock_v_r and returning_v_i both 1.");
+    end
+  end
+
+  logic [1:0] return_v_r;
```
taylor-bsg, line 325: another bsg_counter_up_down ?

v/bsg_manycore_endpoint_standard.v
```diff
@@ -131,108 +134,73 @@ module bsg_manycore_endpoint_standard
...
 
+  // When a request is dequeued, the return info (src coord, response type, etc) is stored in this module.
+  // accelerators can't dequeue another request without returning the response for the already dequeued request.
+  
 
-  // we hide the request if the returning path is not ready
+  // credit counting on response fifo
+  // By default, 3 credits are needed, because the round trip to get the credit back takes three cycles.
+  // FWD_FIFO->CORE->REV_FIFO->CREDIT.
+  logic [lg_rev_fifo_els_lp-1:0] rev_fifo_credit_r;
```
taylor-bsg, line 171: Maybe an additional comment here about how you are using a credit interface to guarantee that any incoming item in FWD_FIFO is guaranteed to have space in the REV FIFO, guaranteeing that the transaction will always go through once it starts?

v/bsg_manycore_endpoint_standard.v
```diff
@@ -241,206 +209,205 @@ module bsg_manycore_endpoint_standard
...
-    ,.init_val_p(max_out_credits_p)
+    .max_val_p((1<<credit_counter_width_p)-1)
+    ,.init_val_p(0)
     ,.max_step_p(1)
   ) out_credit_ctr (
     .clk_i(clk_i)
     ,.reset_i(reset_i)
-    ,.down_i(launching_out) // launch remote store
-    ,.up_i(returned_credit) // receive credit back
+    ,.down_i(returned_credit) // launch remote packet
```
taylor-bsg, line 364: comments seem wrong here

taylor-bsg: both lines

v/bsg_manycore_endpoint_standard.v
```diff
@@ -241,206 +209,205 @@ module bsg_manycore_endpoint_standard
...
-          & (packet_lo.op_v2 != e_remote_amoand)
-          & (packet_lo.op_v2 != e_remote_amoor)
-          & (packet_lo.op_v2 != e_remote_amomin)
-          & (packet_lo.op_v2 != e_remote_amomax)
-          & (packet_lo.op_v2 != e_remote_amominu)
-          & (packet_lo.op_v2 != e_remote_amomaxu)
-        ) else $error("[BSG_ERROR] Incoming packet has an unsupported amo type. op_v2=%d", packet_lo.op_v2);
-      end
+    if ((returned_v_r_o === 1'b1) && (returned_yumi_i != 1'b1) && returned_fifo_full_o) begin
+      $display("## Returned response will be dropped at YX=%d, %d (%m)", global_y_i, global_x_i);
```
taylor-bsg, line 406: Is this actually true? Or is it just that we are going to complain about it because endpoints are _supposed_ to sink?

v/vanilla_bean/mcsr.v
```diff
@@ -240,9 +244,19 @@ module mcsr
...
       endcase
     end
+  end
+
+
+  // credit limit
+  always_comb begin
+    credit_limit_n = credit_limit_r;
 
+    if (we_i & (addr_i == `RV32_CSR_CREDIT_LIMIT_ADDR) & (funct3_i == `RV32_CSRRW_FUN3)) begin
```
taylor-bsg, line 254: should have a warning that checks that outstanding credits is 0 when this set?

taylor-bsg (commented):
Nice! First round of review in

v/vanilla_bean/mcsr.v
taylor-bsg: I guess the question is, will anything break if we reduce the value of the credit limit, and have credits outstanding? Or will it naturally throttle back? Probably has to do with whether we have a <= or = comparator on credit limit.

v/bsg_manycore_endpoint_standard.v
tommydcjung: Not exactly. here you are checking rev_fifo_credit_available not rev_fifo_credit_r.

v/bsg_manycore_endpoint_standard.v
tommydcjung: Not exactly true. Better way of saying is that when the fifo full signal is on, you have to take the response, otherwise the rev packet that thought it was going into rev fifo will get lost.

v/vanilla_bean/mcsr.v
tommydcjung: It's not necessary to wait for credit to be zero, before modifying this CSR. Just don't set it to zero.

dpetrisko:
@tommydcjung @taylor-bsg What's the status on this? This change will break all accelerators not in this repo so I'd like this to be finalized before bringing up anything new

drichmond:
I think this PR is great. I have an incomplete fix for it in replicant that won't get finalized until after the DARPA deadline. No issues for me.

taylor-bsg:
Yeah after the DARPA deadline this will be reviewed and merged.

dpetrisko:
>@tommydcjung @taylor-bsg What's the status on this? This change will break all accelerators not in this repo so I'd like this to be finalized before bringing up anything new

Ping

taylor-bsg:
@drichmond @tommydcjung  is this ready to merge?

tommydcjung:
Nope

drichmond:
I think this is great. We don't have a long term solution in replicant yet but I don't think that's blocking. In short, the host should not use endpoint standard but fixing that should be pretty easy. Alternatively we could merge #492

taylor-bsg:
Seems like these are all things we want to merge? Can you guys figure out
what the steps are?

drichmond:
Sounds good to me! I have to adjust in replicant, but I'll work on top of Dan's PR. I think this PR is great and we should push to get it in tapeout (whether or not it is in RC0 makes no difference to me)

@tommydcjung @dpetrisko is there any thing I can help with?

tommydcjung:
I can't merge this PR now, because this requires the rev link that the host interface connects to be credit-ready interface. In the rtl testbench, we have an io router row that has the P-port as credit-ready, but in the tapeout version, I think the manycore adapter will need to directly connect to one of the links coming out of manycore tile or vcache, which is not credit-ready. @taylor-bsg

taylor-bsg:
So which leaf module needs to be modified to support this?

tommydcjung:
We can't modify the leaf module that contains the router (manycore tile/vcache) to support this. The previous assumption was that we would have a row of IO routers at the top, and those routers would have credit interface on the rev link. However, in the tapeout, we don't have IO routers, but the accelerators, and the manycore link adapter connects directly to the links that are coming out of tiles.

PR #483, opened 2021-03-02
Added parameters to top level

drichmond:
* hetero_type_vec_p - used in the network simulator
* Specific parameters to disable/enable different profilers

hetero_type_vec_p is necessary, and used in the network simulator. I didn't pull it out to the spmd_testbench since I don't think that's a use case that we envision. 

The specific profiler enables are necessary for the network simulator, too, but will also be used for SAIF generation. In the network simulator we want the bind statements for the caches and routers, but compilation will fail for the tiles because no tiles are instantiated.

### DIFF
--- a/v/bsg_manycore_pod_ruche.v
+++ b/v/bsg_manycore_pod_ruche.v
@@ -63,12 +63,6 @@ module bsg_manycore_pod_ruche
 
     , parameter wh_link_sif_width_lp = 
       `bsg_ready_and_link_sif_width(wh_flit_width_p)
-
-    // This is used to define heterogeneous arrays. Each index defines
-    // the type of an X/Y coordinate in the array. This is a vector of
-    // num_tiles_x_p*num_tiles_y_p ints; type "0" is the
-    // default. See bsg_manycore_hetero_socket.v for more types.
-    , parameter int hetero_type_vec_p [0:(num_tiles_y_p*num_tiles_x_p) - 1]  = '{default:0}
   )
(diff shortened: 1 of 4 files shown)


v/bsg_manycore_pod_ruche.v
```diff
@@ -272,6 +278,7 @@ module bsg_manycore_pod_ruche
         ,.addr_width_p(addr_width_p)
         ,.data_width_p(data_width_p)
         ,.ruche_factor_X_p(ruche_factor_X_p)
+        ,.hetero_type_vec_p(hetero_type_vec_p)
```
tommydcjung, line 281: You can't pass in the same hetero_type_vec_p into this module, because the subarray size is smaller

drichmond: Ah, ok. I need to update my arrays to reflect that.

drichmond: What if I take out the hetero parameter, and just do the profiler parameters?

drichmond: The hetero one isn't as important to me. I can work around it in my network simulator branch

tommydcjung: okay you can leave out the hetero type stuff.

drichmond: Fixed

tommydcjung (approved):

PR #484, opened 2021-03-03
Fixes header_print_p input in manycore testbench

mrutt92:


drichmond:
I think this is fine for now, but... will this work with multiple pods?

@tommydcjung your two cents?

drichmond (approved):

PR #487, opened 2021-03-05
CUDA test for checking memory map

mrutt92:


drichmond (approved):

tommydcjung:
Seems like a weird software engineering practice to have all the codes that are only used in other repo get piled up here? @drichmond @mrutt92

drichmond:
Yeah - something that needs to be fixed when we have free time.

PR #496, opened 2021-03-17
set wh dest by network packet

tommydcjung:


tommydcjung:
Added a 1-bit EPA for vcache for setting wh dest cordinate.

gaozihou (approved):
LGTM!

drichmond:
Did the mapping from cache NPA to DRAM get changed in this PR?

drichmond:
Oh, is this the PR we have to fix in software?

tommydcjung:
Yeah, the wh dest coordinates need to be set as part of initialization.

drichmond:
How do I do that? I see this https://github.com/bespoke-silicon-group/bsg_manycore/blob/faf7b30c58b0953cbeb038afab7ca9ca235d171c/v/bsg_manycore_link_to_cache.v#L251

We need to do a remote store to an address on initialization?

tommydcjung:
Yes, remote store as shown here https://github.com/bespoke-silicon-group/bsg_manycore/blob/ci_bigblade/software/py/nbf.py#L214

drichmond:
OK, so if I understand correctly: 

For each cache, send a packet
Address: Set the top two address bits of the EPA
Data: 1 if (x-pods == 1 and south cache) OR (cache-x-coordinate is "east" half), else 0

tommydcjung:
Doesn't matter if it's south or north. depends only on num_pods_x and x-cord of vcache.

PR #497, opened 2021-03-21
pod hierarchy refactor

tommydcjung:


tommydcjung:
- removing io rtr out of pod array
- adding pod row hierarchy
- removing bsg_tag out of pod

PR #500, opened 2021-03-25
imul mux order

tommydcjung:


v/vanilla_bean/fpu_float_fma.v
```diff
@@ -63,7 +63,15 @@ module fpu_float_fma
...
-    if (fp_v_i) begin
+    // FPU gets imul inputs only when there is imul in EXE.
+    // so that it does not cause spurious toggles in FPU by normal integer ops.
+    if (imul_v_i) begin
+      fma_a_li = {1'b0, imul_rs1_i};
+      fma_b_li = {1'b0, imul_rs2_i};
+      fma_c_li = 33'h0;
+      fma_op_li = eIMUL;
+    end
+    else begin
```
drichmond, line 74: Are there any drawbacks to removing the check on `fp_v_i`? I don't have context for the FPU, but I can imagine that when `~imul_v_i` and `~fp_v_i` there could still be toggling.

Do the outputs from the floating point register file, and fpu_float_op_i only change when there is a valid FPU instruction?

taylor-bsg: Seems good to document the assumptions here:
1) the inputs to this module come straight out of registers (usually indicated by _r_i prefix, but that may be too much change for this pull request)
2) if it is not a floating point operation, or there is a bubble, then all of these inputs are zerod (https://github.com/bespoke-silicon-group/bsg_manycore/blob/fa46a85a19bffc47d25f2957c89dc9cd347c5159/v/vanilla_bean/vanilla_core.v#L1500)
3) a zero'd fpu_float_op_i corresponds to eFADD (is that your intent?)

tommydcjung: I changed the fp_exe_n logic so that it invalidates the valid bits in the pipeline without changing other data, so that the toggle within FPU is minimal.

drichmond: I support what @taylor-bsg said. I think my questions have been answered regarding toggling, but it would be good to document that assumption in the code. 

If there are no functional issues with this code can I create a branch that Cornell can run with? Or will this get merged today?

v/vanilla_bean/vanilla_core.v
```diff
@@ -1493,12 +1493,17 @@ module vanilla_core
...
 
   always_comb begin
+    fp_exe_n = fp_exe_r;
     if (stall_all) begin
       fp_exe_n = fp_exe_r;
     end
     else begin
       if (flush | stall_id | ~id_r.decode.is_fp_op) begin
-        fp_exe_n = '0;
+        // put nop in fp_exe.
```
taylor-bsg, line 1502: // we hold the data inputs steady in the case of a stall,
// or if there is not a floating point operation
// to avoid unnecessarily toggling of the FP unit

v/vanilla_bean/fpu_float_fma.v
```diff
@@ -63,7 +63,15 @@ module fpu_float_fma
     fma_op_li = ePM_PB;
     is_fma_op = 1'b0;
 
-    if (fp_v_i) begin
+    // FPU gets imul inputs only when there is imul in EXE.
+    // so that it does not cause spurious toggles in FPU by normal integer ops.
+    if (imul_v_i) begin
```
taylor-bsg, line 68: // assumption: imul_v_i is coming straight of a register and does not glitch

taylor-bsg (approved):
This looks good to me, if @tommydcjung can add the comments I suggested.

PR #504, opened 2021-03-31
Parser fixes 

drichmond:
Parser fixes in master, now applied to bigblade. 

#428 fixes the cache profiler
#480 fixes an issue with statistics counting

The final commit fixes an issue with the header_print_p parameter. Otherwise, the header in vcache_stats is not printed

### DIFF
--- a/testbenches/common/v/bsg_nonsynth_manycore_testbench.v
+++ b/testbenches/common/v/bsg_nonsynth_manycore_testbench.v
@@ -656,7 +656,7 @@ if (enable_cache_profiling_p) begin
   bind bsg_cache vcache_profiler #(
     .data_width_p(data_width_p)
     ,.addr_width_p(addr_width_p)
-    ,.header_print_p("py[0].px[0].pod.north_vc_x[0].north_vc_row.vc_x[0].vc")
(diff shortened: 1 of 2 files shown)


testbenches/common/v/bsg_nonsynth_manycore_testbench.v
```diff
@@ -656,7 +656,7 @@ if (enable_cache_profiling_p) begin
   bind bsg_cache vcache_profiler #(
     .data_width_p(data_width_p)
     ,.addr_width_p(addr_width_p)
-    ,.header_print_p("py[0].px[0].pod.north_vc_x[0].north_vc_row.vc_x[0].vc")
+    ,.header_print_p({`BSG_STRINGIFY(`HOST_MODULE_PATH),".testbench.DUT.py[0].px[0].pod.north_vc_x[0].north_vc_row.vc_y[0].vc_x[0].vc.cache.vcache_prof"})
```
drichmond, line 659: This is the only new change in this PR. @tommydcjung I don't understand how the string match function works, so I just put the whole path.

tommydcjung: Did the cell path change? Not sure why the entire path is needed.

drichmond: The cell path changed, but I also didn't understand the string match function so I just put the whole path.

PR #414, opened 2020-10-26
Added LLVM Makefile commands to try for cmake3

drichmond:
This has been irritating for a while. On the ZeBu system I've been iterating on they only provide cmake, not cmake3, and I can't symlink. 

This solution attempts to find the cmake3 binary (standard only in RHEL), then the cmake binary, and fails (appropriately) if it can find either.

Issue #391, opened 2020-09-03
HOST_TOOLCHAIN documention


marcusnchow:
https://github.com/bespoke-silicon-group/bsg_manycore/blob/428100202eddc7035c24be4e3390237784fc80bc/software/mk/Makefile.llvminstall#L6

Can we add documentation thet HOST_TOOLCHAIN in [software/mk/Makefile.llvminstall](https://github.com/bespoke-silicon-group/bsg_manycore/blob/master/software/mk/Makefile.llvminstall) needs to be set to the... host's... toolchain
Or perhaps change the default to something more generic? 
Or make the error pop up earlier in the build process, cuz I always forget to set it and the error doesn't pop up until like a bit of time has passed.
 
Since this is really the build process it might be better if the documentation was in [bsg_bladerunner](https://github.com/bespoke-silicon-group/bsg_bladerunner)?

drichmond:
Similar to #334 

I think a long term solution would be good. I don't have one off the top of my head

marcusnchow:
Ah okay i didn't look if an issue already existed for it

drichmond:
No worries, this is a better description of the issues

PR #506, opened 2021-04-06
Update Makefile.llvminstall

dpetrisko:
Updates LLVM installation to use default GCC if it's new enough. Needed to compile on recent Ubuntu systems

### DIFF
--- a/software/mk/Makefile.llvminstall
+++ b/software/mk/Makefile.llvminstall
@@ -3,7 +3,14 @@ ifndef LLVM_DIR
 endif
 
 # devtoolset-8
-HOST_TOOLCHAIN ?= /opt/rh/devtoolset-8/root/usr/bin
+GCCVERSION = $(shell gcc -dumpversion)
+ifeq ($(shell expr $(GCCVERSION) \> 7), 1)
+HOST_TOOLCHAIN ?=
+$(info Using default GCCVERSION $(GCCVERSION))
+else
+HOST_TOOLCHAIN ?= /opt/rh/devtoolset-8/root/usr/bin/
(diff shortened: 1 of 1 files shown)


taylor-bsg:
I approve if Dustin does!

drichmond:
Love it. Can you run it on KK9 to make sure it doesn't break Centos?

drichmond:
Solves #391 #414

dpetrisko:
works on kk9!

drichmond:
Seems perfectly reasonable to me. I'm sure there's some corner case but I think the ease-of-use benefit of this outweighs some corner case we haven't thought of

Ship it. 📦

drichmond (approved):

dpetrisko:
The big corner case that broke a lot of other solutions was gcc 10.0, but I believe this works for that as well. 

Not that I checked :P

drichmond:
The remaining question I would ask is - do we want this in bigblade?

taylor-bsg:
@tommydcjung i am guessing Tommy would prefer it on big blade to avoid later merging headaches

dpetrisko:
If bigblade is the new master, it should just be master IMO

dpetrisko:
But less controversially, I sincerely doubt bigblade will touch the LLVM makefile...

tommydcjung:
I would merge into ci_bigblade...

tommydcjung:
image
you can click edit button and change the base

drichmond:
Get rid of the other ones. The second and third are already in bigblade (#504, I think)

dpetrisko:
Should be good to go now

PR #508, opened 2021-04-13
surround hetero type stuff so dc doesn't crash

tommydcjung:


tommydcjung:
Preventing design compiler crash due to function keyword.

drichmond (approved):

PR #511, opened 2021-04-18
Fixing up manycore accelerators

dpetrisko:
This PR fixes the compilation for manycore hetero accelerators. This is necessary for verilator compilation because it compiles files which are unused (which is why VCS doesn't pick up this breakage: a good reason to run Verilator in CI, but I digress).

Are there any hetero machine configs / programs to test this on? I would like to make sure the functionality is there, but if there's nothing available, we should settle for fixing the compilation.


drichmond:
I have one. Will take a day or so to bring back up to speed.

dpetrisko:
Thanks Dustin! Not incredibly urgent but would useful as one of many minor changes to fix Verilation

dpetrisko:
Any movement on this? Probably worth to merge for compilation fixes at least

tommydcjung (approved):

drichmond (approved):

PR #493, opened 2021-03-15
Updated DPI interface for endpoint credit refactor

drichmond:

PR #517, opened 2021-05-05
Updated DPI interface for endpoint credit refactor

drichmond:
* Change credit handling from count down to count up
* Add DPI methods for tracking response fifo capacity
* Add DPI response functionality

### DIFF
--- a/testbenches/dpi/bsg_nonsynth_dpi_manycore.v
+++ b/testbenches/dpi/bsg_nonsynth_dpi_manycore.v
@@ -9,37 +9,46 @@ module bsg_nonsynth_dpi_manycore
      ,parameter y_cord_width_p = "inv"
      ,parameter addr_width_p = "inv"
      ,parameter data_width_p = "inv"
-     ,parameter max_out_credits_p = "inv"
+     ,parameter credit_counter_width_p = `BSG_WIDTH(32)
      ,parameter ep_fifo_els_p = "inv"
-     ,parameter dpi_fifo_els_p = "inv"
+     ,parameter dpi_fifo_els_p = "inv" // Response capacity
      ,parameter rom_els_p = "inv"
      ,parameter rom_width_p = "inv"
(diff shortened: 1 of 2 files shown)


drichmond:
Replaces #493 (which had a lot of force pushes)

drichmond:
This passes regression in replicant on top of #479

dpetrisko:
Does this require changes in replicant host code? i.e. When you read credits for an endpoint to determine whether to send or not?

drichmond:
Yes. See https://github.com/bespoke-silicon-group/bsg_replicant/pull/698

dpetrisko (approved):

testbenches/dpi/bsg_nonsynth_dpi_manycore.v
```diff
@@ -146,40 +155,60 @@ module bsg_nonsynth_dpi_manycore
...
       // manycore link
-      .link_sif_i      (link_sif_i),
-      .link_sif_o      (link_sif_o),
-      .my_x_i          (my_x_i),
-      .my_y_i          (my_y_i),
-      .out_credits_o   (ep_out_credits_lo)
+      ,.link_sif_i(link_sif_i)
+      ,.link_sif_o(link_sif_o)
+      ,.my_x_i(my_x_i)
+      ,.my_y_i(my_y_i)
```
tommydcjung, line 210: The current naming convention is that my*_i means "subcoordinate within a pod". I think you should rename it global_*_i here.

tommydcjung (approved):

testbenches/dpi/bsg_nonsynth_dpi_manycore.v
drichmond: Good catch

It turns out that I no longer need my_x_i and my_y_i since I don't use the endpoint_standard anymore. I've removed these ports to address your feedback.

PR #522, opened 2021-05-10
deprecate dram enable bit

tommydcjung:


drichmond:
Looks good to me

One suggestion for thought: Consider printing out a warning when the DRAM Enable CSR is accessed. If we had other runtimes/users then I think this would be a good idea. However, since we have only one software stack I don't think it's necessary here.

dpetrisko:
What was the use case that this feature was solving? What's changed such that it's no longer relevant?

dpetrisko:
Got it!

PR #526, opened 2021-05-18
Increased timeout for testbenches

drichmond:
I also decreased the beebs iteration count to reduce runtime there. Is it critical to run them twice?

### DIFF
--- a/software/mk/Makefile.verilog
+++ b/software/mk/Makefile.verilog
@@ -1,7 +1,7 @@
 #######################################
 # Simulation run-time flags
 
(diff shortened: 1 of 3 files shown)


tommydcjung (approved):

tommydcjung:
Might want to repeat if it exercises icache misses meaningfully. It has been discussed before.

drichmond:
That's a good reason. It think it's a decision between tests running longer, and possibly more meaningful coverage. I'll leave this open for now. 

My opinion is - if we want to stress the icache we already have tests (or should write more) that stress it.

PR #531, opened 2021-05-25
Reducing runtime of simulation execution

drichmond:
* Replace printf with bsg_print_hexadecimal/int or heartbeats
* make SPMD loader print once every 1024 packets when !verbose_p

In the RC-ALPHA-RTLFREEZE-0 RTL the following tests were the longest of the SPMD suite: 

* quicksort 12961 sec
* vcache_atomic_histogram 17486 sec
* vcache_atomic_histogram2 17739 sec
* perf_test_reduction 20072 sec
* perf_test_conv3x3 21529 sec
* test_credit_limit_csr_multi 22100 sec
* amoadd_test 24099 sec
* bsg_barrier 24771 sec
* bsg_tile_group_barrier 26798 sec
* jal_rv32 28044 sec
* vcache_atomic_inc 28875 sec
* bsg_outbuf_full2 42535 sec

Of these, quicksort (fixed in #530), perf_test_reduction, perf_test_conv3x3, amoadd_test, bsg_barrier, bsg_tile_group_barrier, vcache_atomic_inc, bsg_outbuf_full2 use bsg_printf. Most of these can be replaced with print_hex, print_int, or heartbeat -- which actually provides better debugging, too, since it prints out the source of the integer/hex/heartbeat.

bsg_printf has a high cost in full-chip testing. When executing, tiles try to grab the spinlock at the host, which increases congestion and slows down outgoing print packets, which slows down the progress even more. On top of that, when printf is in the compiled binary the NBF size can double (or more). 

bsg_printf is tested in other tests that are shorter and faster: (nprimes, hello, kmean, fib, striped_hello, etc) that are significantly faster AND not in the tail of the distribution.

Therefore, I propose that we remove printf from these tests. 

Since stdout on linux can slow down simulation, but is a useful heartbeat, I also added a feature to the SPMD loader that prints out the address every 1000 packets when simulating instead of every packet.

### DIFF
--- a/software/spmd/Makefile
+++ b/software/spmd/Makefile
@@ -150,7 +150,7 @@ SCRAPE-TAIL=/usr/bin/sed 's@.*/@@g' | /usr/bin/sed 's@-run.log@@g' | /usr/bin/co
 
 summary: BSG_FINISH.scrape BSG_FINISH.scrape.i BSG_TIMEOUT.scrape BSG_FAIL.scrape BSG_ERROR.scrape
 	@echo -n "Free disk:"; df .
-	@echo -n "SIMVs running:"; /usr/bin/ps a | /usr/bin/grep simv
+	@echo  "SIMVs running:"; /usr/bin/ps a | /usr/bin/grep simv
 
 watch-summary:
 	watch make summary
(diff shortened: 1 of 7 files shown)


drichmond:
For example, replacing `bsg_printf("I am awake. id=%d\n", id);` with `bsg_heartbeat_iter(id)` reduced the runtime of bsg_outbuf_full2 from 42535 seconds to 4455 seconds.

tommydcjung (approved):

PR #533, opened 2021-05-31
Ci gate level sim

tommydcjung:


tommydcjung:
Adding skip_dram_instruction_load options to nbf generator to skip loading instruction to DRAM, if the binary fits in icache.

drichmond (approved):

PR #536, opened 2021-06-06
Toplevel BP <--> HB test

sripathi-muralitharan:
This PR adds a SPMD program and the corresponding BP code that performs matrix-matrix add on a 2x2 tile group in a single pod of the manycore.


tommydcjung:
Can you make your changes in nbf.py more maintainable by separating out bp-specific codes into a separate class/file (e.g. nbf_bp.py), which uses NBP as an abstract class?

tommydcjung (changes requested):
requesting nbf.py refactor

dpetrisko:
@sripathi-muralitharan

software/mk/Makefile.builddefs
```diff
@@ -295,8 +296,24 @@ $(BSG_MANYCORE_LIB): $(BSG_MANYCORE_LIB_OBJS)
...
+NBF_SCRIPT   ?= $(NBF_PY)
+BP_MEM_IMAGE ?=
+BP_DRAM_BASE ?=
+BP_DRAM_POD  ?=
+endif
+
 %.nbf: %.riscv
-	python $(NBF_PY) $*.riscv \
+	python $(NBF_SCRIPT) \
+	  $*.riscv \
```
tommydcjung, line 316: Can you make these parameters a separate makefile variable (e.g. $NBF_PARAM), and add the bp specific params if $HOST equals bp? so it doesn't break the existing nbf.py

software/spmd/bp_manycore_test/bp_main.c
```diff
@@ -0,0 +1,164 @@
+#include <stdint.h>
```
tommydcjung, line 1: Add bp_manycore_test to NO-RECURSE. https://github.com/bespoke-silicon-group/bsg_manycore/blob/eb132d1915350ac79df451b1a7bd37caf5874ce3/software/spmd/Makefile#L9

software/mk/Makefile.builddefs
sripathi-muralitharan: I can do that for the BP-specific parameters. But which script to run is going to remain a separate variable as it is now.

tommydcjung (approved):

PR #547, opened 2021-07-14
Machine configuration fixup

dpetrisko:
- Removes deprecated BSG_MACHINE_BRANCH_TRACE_EN + BSG_MACHINE_VCACHE_MISS_FIFO_ELS
- BSG_MACHINE_X/Y_CORD_WIDTH ->  BSG_MACHINE_GLOBAL_X/Y_CORD_WIDTH (and actually connect it to testbench)
- Adds BSG_MACHINE_POD_X/Y_CORD_WIDTH
- Calculate subcord based on x/y_cord_width_p-pod_x/y_cord_width_p rather than `BSG_SAFE_CLOG2(num_tiles_x/y_p). This is more flexible because there are a variety of reasons one would want to keep the same (overprovisioned) coordinate size. For instance, accelerators with more tiles than manycore pods or software that may be written with a specific coordinate width in mind
- Reorganizes the machines/Makefile VCS_DEFINES list. I'm fine reverting this, but I moved it to match the Makefile.machine.include and it was much easier to verify everything was lined up that way, so I leave it to you


taylor-bsg:
Maybe add some comments somewhere that gives precise semantics of the added variables?

@drichmond BSG_MACHINE_VCACHE_MISS_FIFO_ELS is this the FIFO that controls the v-cache output fifo size? I think this is a feature that we determined had some non-trivial impact on performance, right?

dpetrisko:
>BSG_MACHINE_VCACHE_MISS_FIFO_ELS is this the FIFO that controls the v-cache output fifo size? I think this is a feature that we determined had some non-trivial impact on performance, right?

Is only used for non-blocking cache

v/bsg_manycore_tile_compute_ruche.v
```diff
@@ -18,8 +18,8 @@ module bsg_manycore_tile_compute_ruche
     // Number of tiles in a pod
     , parameter num_tiles_x_p="inv"
     , parameter num_tiles_y_p="inv"
-    , parameter x_subcord_width_lp = `BSG_SAFE_CLOG2(num_tiles_x_p)
-    , parameter y_subcord_width_lp = `BSG_SAFE_CLOG2(num_tiles_y_p)
+    , parameter x_subcord_width_lp=x_cord_width_p-pod_x_cord_width_p
+    , parameter y_subcord_width_lp=y_cord_width_p-pod_y_cord_width_p
```
tommydcjung, line 22: I think you want to calculate it the other way to avoid inconsistent input parameters.  pod_x_cord_width_p = x_cord_width_p-x_subcord_width_lp, where x_subcord_width_lp=`BSG_SAFE_CLOG2(num_tiles_x_p).

tommydcjung (changes requested):
rc

v/bsg_manycore_tile_compute_ruche.v
dpetrisko: I believe decoupling the coordinate width from the number of tiles is better since:
- Manycore tile PD is identical for pods of different sizes. 
- Software written for pods of different sizes will be identical
- Changing the size of the manycore then does not change attached accelerator PD

dpetrisko:
>Maybe add some comments somewhere that gives precise semantics of the added variables?

Where's the best place to add this documentation @drichmond ?

v/bsg_manycore_tile_compute_ruche.v
tommydcjung: Manycore tile PD is not identical for pods of different sizes. For example. eva2npa mapping hardware will change. num_tiles_x/y_p determines the size of the pod, so it makes sense to have it derived from there.

tommydcjung:
> > Maybe add some comments somewhere that gives precise semantics of the added variables?
> 
> Where's the best place to add this documentation @drichmond ?

https://github.com/bespoke-silicon-group/bsg_manycore/blob/master/machines/README.md

v/bsg_manycore_tile_compute_ruche.v
taylor-bsg: Yeah the pod would not be identical. But I think Dan is saying that we could have more stability in terms of the pinouts of the tile? So the PD flow could be more uniform?

drichmond:
> Maybe add some comments somewhere that gives precise semantics of the added variables?
> 
> @drichmond BSG_MACHINE_VCACHE_MISS_FIFO_ELS is this the FIFO that controls the v-cache output fifo size? I think this is a feature that we determined had some non-trivial impact on performance, right?

No - BSG_MACHINE_VCACHE_MISS_FIFO_ELS is for the non-blocking cache. IIRC it is the number of misses that can be absorbed from the network before back-pressuring the network. The response fifo depth that you're referring to is a different parameter that has never been PR'd (it is called BSG_MACHINE_VCACHE_RSP_FIFO_ELS).

I can add that in a PR. I rebased the branch containing it on Monday.

drichmond:
BSG_MACHINE_BRANCH_TRACE_EN was used to debug beebs, I think. Every time a branch was taken the core would spit out the result. I thought it was still hooked up as per some discussion in the manycore/tapeout meeting a couple months ago.

tommydcjung:
can we leave BSG_MACHINE_X/Y_CORD_WIDTH without changing its name? It will break the testbench tapeout repo and other places. Just document its meaning well in the README

drichmond:
It looks to me like BSG_MACHINE_BRANCH_TRACE_EN was removed from the top level testbench, and that may have been accidental. I don't think that means it was deprecated.

I recall a long conversation at one of the meetings about keeping it. I would prefer to keep it and we can file an issue to decide whether to remove it, and if not, have a test that verifies it still works.

dpetrisko:
>can we leave BSG_MACHINE_X/Y_CORD_WIDTH without changing its name? It will break the testbench tapeout repo and other places. Just document its meaning well in the README

Sure, no problem

dpetrisko:
>It looks to me like BSG_MACHINE_BRANCH_TRACE_EN was removed from the top level testbench, and that may have been accidental. I don't think that means it was deprecated.

#548

v/bsg_manycore_tile_compute_ruche.v
dpetrisko: identical->mostly identical. The TCL scripts can stay the same.

Basically, it has effects on things independent of the manycore pod itself. Notably accelerators + software.

dpetrisko:
How's this looking?

dpetrisko:
@mrutt92 relevant to the 4x2 config?

mrutt92:
yeah, the ability to adjust pod_x/y_cord_width in the machine file is helpful for getting the 4x2 pod size to work.

v/bsg_manycore_tile_compute_ruche.v
mrutt92: As someone who interacts with the RTL more than the PD, I think I support @tommydcjung's suggestion for the reasons he mentioned. 

But if there's merit to @dpetrisko's approach (tile cord width is calculated from the pod cord width and full cord width) most of the confusion over inconsistent parameter settings could be handled with an elaboration time assertion.

dpetrisko: It seems like if we are set on calculating subcord width, then we should remove
BSG_MACHINE_X_CORD_WIDTH, BSG_MACHINE_Y_CORD_WIDTH as parameters and calculate them as well since they serve no purpose. Otherwise, they will become inconsistent.

tommydcjung: I agree with having measures to prevent users from plugging inconsistent sets of parameters. In the context of manycore pod and this testbench (the source code you are modifying here), it is assumed that num_tiles in a pod determines the subcord width. I would say the user should be plugging in x/y_cord_width (global cord width), because that's the coordinate width in the network packet that is universal to the things connected to the manycore network. And you can calculate pod_x/y_cord as x/y_cord_width - subcord_x/y.

dpetrisko: Works for me, that will at least fix the bug this PR was intended for.

tommydcjung (approved):

PR #553, opened 2021-07-20
Ci regfile latch

tommydcjung:


taylor-bsg:
To make naming consistent, I suggest renaming to sram, latch and ff wherever applicable. The select parameter can be named sel_sram_latch_rf_p; meaning, 2=sram, 1=latch, 0=RF; to avoid to awkward interdependence of the current latch and harden parameters.

PR #558, opened 2021-08-05
fix sp init value

tommydcjung:


tommydcjung:
Fixing the problem where the last word of DMEM is never used as stack.

drichmond (approved):

mrutt92 (approved):

PR #570, opened 2021-09-01
MCS mutex

mrutt92:
Fast and scalable mutex for the manycore.


software/bsg_manycore_lib/bsg_mcs_mutex.S
```diff
@@ -0,0 +1,81 @@
...
+bsg_mcs_mutex_acquire_ret:
+        ret
+
+        .globl bsg_mcs_mutex_release
+bsg_mcs_mutex_release:
+        // a0 = mtx         : bsg_mcs_mutex_t*
+        // a1 = lcl         : bsg_mcs_mutex_node_t*
+        // a2 = lcl_as_glbl : bsg_mcs_mutex_node_t*
+        li      t1, 1                                       // t1 = 1
+        lw      t0, 0(a1)                                   // next = lcl->next
```
drichmond, line 54: Reorder? It's a small change, but I think switching li and lw will hide the load-use distance for memory and remove a stall.

software/bsg_manycore_lib/bsg_mcs_mutex.S
```diff
@@ -0,0 +1,81 @@
...
+        // this is the case where there is a successor
+        // we need only unlock the successor and return
+        fence                                               // fence to implement release semantics
+        sw      t1, 4(t0)                                   // successor->unlocked = 1
+        ret
+bsg_mcs_mutex_release_no_successor:
+        // this is the case where there is no known successor
+        // attempt to swap out the tail pointer with 0
+        amoswap.w.rl t2, x0, 0(a0)                          // victim_tail = swap(&mtx, 0)
+        bne     t2, a2, bsg_mcs_mutex_release_exists_victim // victim_tail == lcl_as_glbl?
```
drichmond, line 65: The blind use of a2 depends on the compiler putting *next at offset 0 of the struct, right?

software/bsg_manycore_lib/bsg_mcs_mutex.S
```diff
@@ -0,0 +1,81 @@
...
+        // we need to put them back
+        amoswap.w t2, t2, 0(a0)                             // usurper = swap(&mtx, victim_tail)
+bsg_mcs_mutex_release_wait_on_successor:
+        lw      t3, 0(a1)                                   // next = lcl->next
+        bnez    t3, bsg_mcs_mutex_release_wait_on_successor // while (lcl->next != 0)
+        bnez    t2, bsg_mcs_mutex_release_exists_usurper    // was there an usurper?
+        // no usurper exists -- unlock our successor
+        sw      t1, 0(t3)                                   // next->unlocked = 1
+        ret
+bsg_mcs_mutex_release_exists_usurper:
```
drichmond, line 78: Four threads: Releaser, Victim, Usurper, and Naive

What happens if there's a victim, and the naive thread adds itself to the usurper's next before the releaser can fix the usurper's next?

software/bsg_manycore_lib/bsg_mcs_mutex.S
mrutt92: good point!

software/bsg_manycore_lib/bsg_mcs_mutex.S
mrutt92: I'll add a comment about the ordering of the struct members.

software/bsg_manycore_lib/bsg_mcs_mutex.S
```diff
@@ -0,0 +1,81 @@
...
+        .text
+        .globl bsg_mcs_mutex_acquire
+bsg_mcs_mutex_acquire:
+        // a0 = mtx         : bsg_mcs_mutex_t*
+        // a1 = lcl         : bsg_mcs_mutex_node_t*
+        // a2 = lcl_as_glbl : bsg_mcs_mutex_node_t*
+        sw      x0, 0(a1)                      // lcl->next = 0
+        sw      x0, 4(a1)                      // lcl->unlocked = 0
+        amoswap.w.aq t0, a2, 0(a0)             // predecessor = swap (&mtx, lcl_as_glbl)
+        addi    t1, a1, 4                      // t1 = &lcl->unlocked
```
tommydcjung, line 37: Is it possible to swap the location of next and unlocked in bsg_mcs_mutex_node_t, so you can get rid of this addi?

drichmond: I think this is the dual of the comment I made below - I asked about compiler reordering. If the struct is reordered, then there needs to be an addi in bsg_mcs_mutex_release_no_successor to access the next pointer. Other places may exist, but that's where I put my comment.

mrutt92: Well currently I don't do the load on broken reservation when spinning on the next pointer in the no successor case, so I can just use an offset load to access the next pointer. I commit the load-on-broken because we can expect that pointer to be updated imminently by the successor that's just added themselves to the list.

software/bsg_manycore_lib/bsg_mcs_mutex.S
mrutt92: actually in this case, no it doesn't depend. a0 is of type (bsg_mcs_mutex_node_t **) so it's a pointer-pointer.

mrutt92: a2 is a pointer, we are not dereferencing it in this function.

software/bsg_manycore_lib/bsg_mcs_mutex.S
```diff
@@ -0,0 +1,80 @@
+// MCS mutex
+// Author: Max
+//
+// This is an implementation of the MCS mutex described by Mellor-Crummey and Scott in their 1991 paper
```
taylor-bsg, line 4: inspired in part by

software/bsg_manycore_lib/bsg_mcs_mutex.S
```diff
@@ -0,0 +1,80 @@
...
+//
+// Cores atomically append their local memory region to the global list using an unconditional
+// amoswap operation. They then spin on their local memories for a predecessor in the queue
+// to notify them that they now hold the lock.
+//
+// Once a core has completed its critical region, it checks for a successor and updates releases the lock to them.
+//
+// The advantages of this mutex over a simple spin lock on the manycore are two fold:
+//
+// (1) It greatly reduces the number of memory reqeusts on the network and it mitigates the extent to which
```
taylor-bsg, line 18: requests

software/bsg_manycore_lib/bsg_mcs_mutex.S
```diff
@@ -0,0 +1,80 @@
...
+// (2) The lock approximates a FIFO-ish structure, which improves fairness. A simple spinlock on the manycore
+// will favor threads topologically closer to the memory bank in which the lock resides and can lead to
+// starvation of the other cores.
+//
+// This lock is by no means perfect. For locks with low contention, a simple spinlock may result in better performance.
+
+        .text
+        .globl bsg_mcs_mutex_acquire
+bsg_mcs_mutex_acquire:
+        // a0 = mtx         : bsg_mcs_mutex_t*
```
taylor-bsg, line 31: - definition and naming of input parameters?  
- probably a command to clarify that the global pointer would preferably a tile group global pointer, unless the synchronization is being done between tile groups; this would preserve virtualization of tile groups

taylor-bsg: indication that a0 must be in DRAM address space?

taylor-bsg: a1/a2 must be in the tile's local DMEM?

mrutt92: see the header bsg_mcs_mutex.h, I can add them here too.

mrutt92: I missed the dram address space bit though

software/bsg_manycore_lib/bsg_mcs_mutex.S
```diff
@@ -0,0 +1,80 @@
...
+        sw      x0, 4(a1)                      // lcl->next = 0
+        sw      x0, 0(a1)                      // lcl->unlocked = 0
+        amoswap.w.aq t0, a2, 0(a0)             // predecessor = swap (&mtx, lcl_as_glbl)
+        beqz    t0, bsg_mcs_mutex_acquire_ret  // return if predecessor = 0
+        sw      a2, 4(t0)                      // predecessor->next = lcl_as_glbl
+        lr.w    t0, (a1)                       // unlocked = lcl->unlocked
+        bnez    t0, bsg_mcs_mutex_acquire_ret  // return if unlocked
+bsg_mcs_mutex_acquire_loop:
+        lr.w.aq t0, (a1)                       // unlocked = lcl->unlocked
+        beqz    t0, bsg_mcs_mutex_acquire_loop // while !unlocked
```
taylor-bsg, line 43: // backwards predict not taken branch variant would help here

software/bsg_manycore_lib/bsg_mcs_mutex.S
taylor-bsg: Depending on whether you think this is the common or uncommon case, you can reverse the sense of the branch and branch forwards or backwards to somebody else's ret instruction, and save icache footprint

software/bsg_manycore_lib/bsg_mcs_mutex.S
```diff
@@ -0,0 +1,80 @@
...
+        // attempt to swap out the tail pointer with 0
+        amoswap.w.rl t2, x0, 0(a0)                          // victim_tail = swap(&mtx, 0)
+        bne     t2, a2, bsg_mcs_mutex_release_exists_victim // victim_tail == lcl_as_glbl?
+        ret                                                 // there really is no successor -- return
+bsg_mcs_mutex_release_exists_victim:
+        // someone added themselves to the queue and we have removed them
+        // we need to put them back
+        amoswap.w t2, t2, 0(a0)                             // usurper = swap(&mtx, victim_tail)
+bsg_mcs_mutex_release_wait_on_successor:
+        lw      t0, 4(a1)                                   // next = lcl->next
```
taylor-bsg, line 71: should be using LBR?

software/bsg_manycore_lib/bsg_mcs_mutex.S
taylor-bsg: you can have a comment that says 

// refer to bsg_mcs_mutex.h for detailed description of usage

software/bsg_manycore_lib/bsg_mcs_mutex.S
mrutt92: I guess I made a judgement call here. My reasoning was that this situation means a successor is in the process of acquiring the lock and an update to the next pointer is imminent (10s of cycles). Using LBR requires extra instructions and I thought it was likely not worth the additional instruction footprint given that this a hopefully uncommon corner case.

taylor-bsg: maybe you can document this in the code? for each of these synchronizations, what is the expectation?

software/bsg_manycore_lib/bsg_mcs_mutex.S
taylor-bsg: // note that this is also polling, since it does not reset the reservation
// what is the expected wait time in this case?

mrutt92: > note that this is also polling, since it does not reset the reservation
should it branch back to making the reservation, then?

> what is the expected wait time in this case?
the wait time can be arbitrarily long as it depends on both lock contention and the time required for the client's critical region. hopefully the client code does not require more than 100 cycles (possible if there's a cache miss) and contention is low (maybe no more than 2 cores waiting on the lock at once).

> is it possible that somebody would be writing to this register to do anything except unlockit?
memory location or register?
for the memory location: gosh I hope not. only two writes happen to this memory. the first is before a core tries to acquire the lock and it clears it. the second is when the lock owner is releasing the lock and unlocking its successor.
for the register: are we talking about a situation where we have context switched?

software/bsg_manycore_lib/bsg_mcs_mutex.S
mrutt92: done

software/bsg_manycore_lib/bsg_mcs_mutex.S
```diff
@@ -0,0 +1,103 @@
...
+        //
+        // The expected wait time for this load is arbitrarily long as it depends
+        // on (1) the time it takes client code to complete the critical region
+        // and (2) the contention on this lock.
+        // We expect the wait time to be on the order of 20-100 cycles in the
+        // case where there is low contention on the lock.
+        lr.w    t0, (a1)                       // unlocked = lcl->unlocked
+        bnez    t0, bsg_mcs_mutex_acquire_ret  // return if unlocked
+        lr.w.aq t0, (a1)                       // unlocked = lcl->unlocked
+        // MBT: backwards predict not taken branch variant would be helpful here
```
taylor-bsg, line 55: Just add a few notes here should we revisit:

// MBT: if we supported context switching the reservation register, then we would not even
//         need this branch (i.e. a blind synchronization); but if there was a context switch, 
//         then the reservation would be cleared and we would this need to go back to the lr.w 
//         to reprime the barrier.
//
// MBT: if lr.w.aq re-primed the reservation AND we supported context switching the reservation
// register, then we could just jump back to the lr.w.aq instruction

taylor-bsg (commented):
Looks great! Just had a few suggested comments to add.

mrutt92:
OK. Feedback has been addressed.

taylor-bsg:
super cool!

taylor-bsg (approved):
My name is taylor-bsg and I approve this message.

PR #574, opened 2021-09-07
Printf mcs mutex

mrutt92:


taylor-bsg:
is this one ready to go, max?

software/bsg_manycore_lib/bsg_tile_config_vars.h
```diff
@@ -29,5 +31,6 @@ extern int __bsg_grid_dim_y;      //The Y Dimensions of the grid of tile groups
 extern int __bsg_tile_group_id_x; //The X Cord of the tile group within the grid
 extern int __bsg_tile_group_id_y; //The Y Cord of the tile group within the grid
 extern int __bsg_tile_group_id;   //The flat ID of the tile group within the grid
-
+extern int __bsg_pod_x;           //The X Cord of the tile's pod (lives in DRAM)
+extern int __bsg_pod_y;           //The Y cord of the tile's pod (lives in DRAM)
```
tommydcjung, line 35: You could grab these values from cfg_pod CSR at the beginning. There is no need to have the host set these.

mrutt92: I guess that's true. It save's a trivial few instruction words in the initialization to have the host set it. That CSR resets to the current pod but it can be set by the core to something else (like if it wants to read another pod's DRAM).

mrutt92: I'm honestly not 100% sure which is better. I think it's better to have the host set it? It's just a couple extra packets in the nbf and we don't need to entangle it with the semantics of the pod csr. But I'm not sure, I'm open to be talked down from that.

tommydcjung: I don't think it makes sense to keep these variable in DRAM, because if you change the pod x/y CSR then now you are reading those values from different pods' DRAM space. I think it's better off to make global x/y readable by CSR instructions (read-only), then it doesn't take up space in neither DRAM or DMEM.

mrutt92: I think that makes a lot of sense. That'll also decouple these values from the semantics of the pod address register.

taylor-bsg: Why is it necessary to access the X and Y coordinate of the pod? As a general principle we would like to avoid code that depends on this, since it is a virtualization hazard...

mrutt92: It's a global lock.

taylor-bsg: Yes why do we need a global lock across pods..

mrutt92: Maybe we don't? If each pod has its own IO node then there's no need.

mrutt92: At the moment, there's only one IO node shared across pods. A global lock prevents the IO node from being slammed with packets from all pods.

taylor-bsg:
Seems okay though? With individual pod locks, the throughput is already
throttled to some extent?

taylor-bsg:
My thought is that the lock is presumably there because of lack of
re-entrancy in some library; which would not apply for multipod, since they
have their own DRAM spaces. Or is the lock there to reduce interspersing of
characters or ... ?

mrutt92:
Both. It prevents one IO node from being slammed and it also serves as keeping the messages from being garbled.

PR #577, opened 2021-09-09
Cache utilization parser

drichmond:
WIP: Needs more comments and testing

* Generates cache utilization graph for each cache, and all caches collectively
* Idle: No operations
* Stall: Cache was stalled (typically by response network)
* Miss: Cache was responding to a miss
* Active: Cache responded to an operation
* The sum of all fields is equal to 1

Examples for a cache resident workload: 

Overall: 
cache_utilization

Index 29: 
cache_29_utilization


taylor-bsg:
so cool!

taylor-bsg:
wouldn't be so hard to implement this in real HW, with some vcache CSRs
that we poll

PR #579, opened 2021-09-15
LSU refactor

tommydcjung:


v/vanilla_bean/lsu.v
```diff
@@ -58,8 +58,8 @@ module lsu
   logic [data_width_p-1:0] mem_addr;
   logic [data_width_p-1:0] miss_addr;
 
-  assign mem_addr = exe_rs1_i + mem_offset_i;
-  assign miss_addr = (pc_plus4_i - 'h4) | bsg_dram_npa_prefix_gp;
+  assign mem_addr = exe_rs1_i + {{20{mem_offset_i[11]}}, mem_offset_i};
+  assign miss_addr = {1'b1, {(data_width_p-pc_width_p-3){1'b0}}, (pc_width_p)'(pc_plus4_i - 'h1), 2'b00};
```
taylor-bsg, line 62: Seems like this is a little bit of a reversion, in that the npa_prefix_gp is now implicit. Is bsg_dram_npa_prefix_gp used anywhere else?  Should we just concatenate instead?

tommydcjung: bsg_dram_npa_prefix_gp is not used anywhere else. I can add a comment that miss_addr maps to DRAM address space.

taylor-bsg: Maybe another comment that says I-Cache EVA to Pod Data EVA translation, and a comment in the file https://github.com/bespoke-silicon-group/bsg_manycore/blob/master/v/bsg_manycore_eva_to_npa.v that says,  Note: I-cache EVA to Pod Data EVA translation occurs in lsu.v

PR #587, opened 2021-09-29
fcsr assertion fix

tommydcjung:
A CSR instruction in ID should stall (stall_fcsr) until all the float instructions that could update fflags (e.g. fpu_float, fpu_int, fdiv ops) and have already been issued, finish and update fflags. Currently, the exception in fcsr module is being triggered even when the CSR instruction modifies only frm (rounding mode). So the assertion is being fixed to prevent this false alarm.


drichmond (approved):
Oh, good! Thrilled to hear this wasn't a bug. 

This looks good to me. It resolves my issue. I will turn my branch into to a PR!

PR #588, opened 2021-09-29
Tune C libraries for manycore architecture

drichmond:
This adds the --with-tune flag to the GCC build flow -- ensuring that the standard libraries that we use are tuned for the manycore architecture. 

This optimization uncovered an issue with RTL simulation that was resoled in #587.

This PR also contains a test that verifies that fix has been applied.


software/spmd/float_logf/main.cpp
```diff
@@ -0,0 +1,30 @@
...
+#include <math.h>
+
+int main(){
+        bsg_set_tile_x_y();
+
+        if ((__bsg_x == 0) && (__bsg_y == 0)){
+                float data[4] = {-1.0f, -0.0f, 0.0f, 1.0f};
+
+                for (int i = 0; i < 4; i++){
+                        bsg_print_float(logf(data[i]));
```
tommydcjung, line 23: Can you also verify the result and send bsg_fail() if it fails? Also with  other non-trivial inputs like 2 or 3.

drichmond: yeah, good point. Will do.

drichmond: I did this, but another issue popped up. Regardless of whether or not I use --with-tune or not, isnan and isfinite return incorrectly (see test).

I don't know how those functions work, but maybe they check a CSR? I haven't had time to look into it completely yet.

The normal cases produce correct results: logf(1) returns 0, logf(e) returns 1.

drichmond: The culprit is -ffast-math in Makefile.builidefs. 

The solution is adding -fno-finite-math-only in the SPMD test makefile.

tommydcjung (approved):

PR #589, opened 2021-09-30
[RFC] Verilator support

dpetrisko:
This PR adds support for verilator compilation. Very naively, I just split the compilation makefile into Makefile.vcs and Makefile.verilator.

There are some random changes included which I'll clean up, but I wanted to get feedback on the preferred way of switching between the compilation tools. It doesn't have to be super easy to go back and forth. Presumably, VCS will continue to be the preferred BSG way with Verilator being used by external contributors. For instance, we could only allow either simv or simsc to be built at a given time, etc. Whatever makes this easy to reason about. To throw another wrench in, I plan to add Xcelium (Cadence) simulator support next, so whatever option we choose should scale to that as well.

(This builds on https://github.com/bespoke-silicon-group/bsg_manycore/tree/small-pod for the small manycore, which makes verilator compilation palatable)


dpetrisko:
Note that this requires a pretty recent verilator that incorporates the patch for DPI clock generator support. The version on our group server is recent enough, that's what I've been developing with

drichmond:
Cool

TODO: We should increment the verilator submodule in bladerunner to get this PR.

dpetrisko:
Yeah bladerunner verilator platform is the next step PR

software/mk/Makefile.paths
```diff
@@ -1,4 +1,4 @@
 BSG_IP_CORES_DIR ?= $(BSG_MANYCORE_DIR)/../basejump_stl
 RISCV_BIN_DIR ?= $(BSG_MANYCORE_DIR)/software/riscv-tools/riscv-install/bin
 CAD_DIR ?= $(BSG_MANYCORE_DIR)/../bsg_cadenv
-BSG_MACHINE_PATH ?= $(BSG_MANYCORE_DIR)/machines/pod_1x1
+BSG_MACHINE_PATH ?= $(BSG_MANYCORE_DIR)/machines/pod_1x1_4X2Y
```
tommydcjung, line 4: can we keep pod_1x1 as default?

testbenches/common/v/spmd_testbench.v
```diff
@@ -4,9 +4,10 @@
  */ 
 
 
-module spmd_testbench();
+module spmd_testbench
   import bsg_manycore_pkg::*;
   import bsg_manycore_mem_cfg_pkg::*;
+  (output bit reset_i);
```
tommydcjung, line 10: reset_o?

software/mk/Makefile.paths
dpetrisko: yeah definitely I'll revert all the defaults before merging

testbenches/common/v/spmd_testbench.v
dpetrisko: I'll change that. For context, some testbenches we've done have taken clk/reset from another level up if not DPI, so the DPI testbenches reversed the direction and kept the name. No reason to keep it like that though, since we're keeping the clock gens in here

software/mk/Makefile.verilog
```diff
@@ -40,18 +40,27 @@ endif
 #######################################
 # Simulator setup
 
+TOOL ?= VCS
```
dpetrisko, line 43: @drichmond better way to deal with this?

machines/Makefile
```diff
@@ -8,143 +8,16 @@ ifeq ($(wildcard $(BASEJUMP_STL_DIR)/imports/DRAMSim3/Makefile),)
...
-		$(VCS_INCLUDES) $(VCS_DEFINES) $(CSOURCES) $(VCS_SOURCES) $*/bsg_tag_boot_rom.v \
-		-l $*/build-coverage.log -Mdir=$*/csrc-coverage
-
-clean:
-	rm -rf */csrc*
-	rm -rf */*.log */*.daidir */simv */simv-debug */simv-profile */simv-coverage
-	rm -rf stack.info.*
-	rm -f */vc_hdrs.h
-	rm -f */*.tr */bsg_tag_boot_rom.v
+TOOL ?= VCS
```
dpetrisko, line 15: Better way to deal with this?

software/mk/Makefile.verilog
drichmond: My response is what we did in replicant. Use platform.mk as a single source of truth for the simulator. 

The downside is that with different makefiles, cleaning is a bit weird. Sometimes you end up with files floating around from the simulator you're not using.

That or have separate targets for VCS and Verilator. simv and simv.verilator or something. 

Since VCS and Verilator are different enough, there's not a good way to handle the two toolchains like Clang and GCC. You could try though (e.g. generate a .daidir etc for Verilator so that you have the same intermediate outputs)

testbenches/common/v/spmd_testbench.v
```diff
@@ -190,5 +192,15 @@ module spmd_testbench();
     ,.ctr_r_o(global_ctr)
   );
 
+  // Disable assertions until reset is lowered
```
tommydcjung, line 195: why is this necessary?

dpetrisko: Can remove if you want, but there are a bunch of spurious assertions that trigger before reset is lowered and this cleans that up

taylor-bsg:
Generally the best practice for spurious assertions and resets is  &&
(reset_i !== 0) which includes X and 1.

dpetrisko:
>Generally the best practice for spurious assertions and resets is  &&
(reset_i !== 0) which includes X and 1.

1. Verilator doesn't support X
2. Does not work for a pulsed reset which is 0 at the beginning (____|----|____), which is the case for many testbenches

testbenches/common/v/spmd_testbench.v
drichmond: > > Generally the best practice for spurious assertions and resets is  &&
> > (reset_i !== 0) which includes X and 1.
> 
> 1. Verilator doesn't support X
> 2. Does not work for a pulsed reset which is 0 at the beginning `(____|----|____)` which is the case for many testbenches

FWIW, we handle this in replicant in C.

tommydcjung: What are the warnings from VCS? I've never seen any.

dpetrisko: >FWIW, we handle this in replicant in C.

Right. Which is why it seems to me to make sense to handle it in an equivalent way in VCS, instead of relying on X-optimism. But it's not my hill, I reverted it

dpetrisko: > What are the warnings from VCS? I've never seen any.

I think they came up when I was messing with the reset input, but I don't see them anymore, so should be fine

dpetrisko:
I'll add a 1b. which is that I've seen companies use 2-state variants of simulators for speed when they're not concerned about X-prop (power+performance validation)

taylor-bsg:
I think the challenge is, this assumes you have uniform reset across the
chip. I guess you could turn off assertions after you have asserted reset?

Dustin, maybe worth benchmarking manycore simulation with 0/1 for the paper?


M

dpetrisko:
>I think the challenge is, this assumes you have uniform reset across the
chip. I guess you could turn off assertions after you have asserted reset?

Yeah that's true. AFAIK, most simulators support $assertoff(hier.path.to.reset.domain). So you could do it in stages as the chip comes out of reset for more complex situations. In the manycore case, there's only a single domain so we only care about propagated rather than heterogeneous reset. Therefore, enabling assertions when reset falls is sufficient assuming that reset=1 has propagated throughout the chip

v/vanilla_bean/vanilla_core.v
```diff
@@ -396,12 +396,15 @@ module vanilla_core
...
   );
 
+  // Sensitivity list like this is disliked by Verilator 4.213
+  `ifndef VERILATOR
    always @ (cfg_pod_y_o or cfg_pod_x_o)
      begin
 	$display("%m cfg_pod_r changing to y=%b x=%b"
 		 , cfg_pod_y_o
 		 , cfg_pod_x_o);
      end
```
drichmond, line 406: Couldn't this be separated into two sensitivity blocks?

dpetrisko: Possibly? It basically continuously triggers this block, so not sure if that’ll help. I’ll give it a shot

dpetrisko: Still triggers both continuously, I added the translate guards though

dpetrisko:
Any last comments?

PR #592, opened 2021-10-12
FP register file write banking.

tommydcjung:


drichmond:
Adding a note - I used this in SGEMM and it reduced write-back stall cycles from 5% to 0% without needing to write assembly. 

This impact will likely increase when I switch to sw pipelining.

v/vanilla_bean/regfile.v
```diff
@@ -15,7 +15,10 @@ module regfile
     , `BSG_INV_PARAM(els_p)
     , `BSG_INV_PARAM(num_rs_p)
     , `BSG_INV_PARAM(x0_tied_to_zero_p)
-    , parameter harden_p=0
+    // FF-based = 0
+    // Latch-based = 1
+    // SRAM-based = 2
+    , parameter sel_sram_latch_rf_p=1
```
dpetrisko, line 21: Just food for thought, do we want to expose this as a toplevel parameter? Or treat it as a hardened swap? Is Latch-based the correct default here?

v/vanilla_bean/fcsr.v
```diff
@@ -26,8 +26,12 @@ module fcsr
     , output logic data_v_o      // 1, if addr_i matches fcsr addr.
 
     // exception accrue interface
-    , input [1:0] fflags_v_i
-    , input [1:0][fflags_width_lp-1:0] fflags_i
+    // exception sources:
```
dpetrisko, line 29: Trying to understand why this additional source is coming from this change? Where did the fdiv/fsqrt fflags get accumulated before?

v/vanilla_bean/vanilla_core.v
```diff
@@ -1330,13 +1286,19 @@ module vanilla_core
   wire id_rs2_equal_wb_rd = (id_rs2 == wb_ctrl_r.rd_addr);
 
   // stall_depend_long_op (idiv, fdiv, remote_load, atomic)
+  //logic [float_rf_num_banks_p-1:0] id_rs2_match_sb_clear;
```
dpetrisko, line 1289: delete? Unless here for historical purposes

v/vanilla_bean/vanilla_core.v
```diff
@@ -1123,10 +1081,8 @@ module vanilla_core
     | stall_barrier;
 
   wire stall_all = stall_icache_store
-    | stall_idiv_wb
     | stall_remote_ld_wb
```
dpetrisko, line 1084: These signals are removed because the write is now transparent so we don't need to stall? Is the output latched on the idiv/fdiv unit outputs such that there's no critical path issue?

v/vanilla_bean/vanilla_core.v
```diff
@@ -570,17 +531,17 @@ module vanilla_core
...
     ? rs1_forward_val
     : int_rf_rdata[0];
   
   assign rs2_val_to_exe = id_r.decode.read_frs2
-    ? fsw_data
+    ? float_rf_rdata[1]
     : (rs2_forward_v
-      ? rs2_forward_val
-      : int_rf_rdata[1]);
+      ? {1'b0, rs2_forward_val}
```
dpetrisko, line 543: consider BSG_ZERO_EXTEND?

v/vanilla_bean/vanilla_core.v
```diff
@@ -265,18 +267,19 @@ module vanilla_core
...
+  logic [float_rf_num_banks_p-1:0][fpu_recoded_data_width_gp-1:0] float_rf_wdata;
  
   logic [2:0] float_rf_read;
   logic [2:0][fpu_recoded_data_width_gp-1:0] float_rf_rdata;
 
-  regfile #(
-    .width_p(fpu_recoded_data_width_gp)
-    ,.els_p(RV32_reg_els_gp)
+
+  bsg_mem_multiport_latch_write_banked_bypassing_sync #(
```
dpetrisko, line 278: This is my only major concern about this change. The integer regfile has the option to be ff, latch or sram-based which helps for portability between FPGA, ASIC and open-source tools. By baking in the latch requirements here, is it too restrictive to external users? How difficult would it be to add the banking to the regfile/ff implementations as well? If the answer is 'more than we're willing to spend', maybe we could have a wrapper here (regfile_banked?) and upon selecting other implementations than latch-based, we could $error("FF/SRAM multi-banking not implemented")

tommydcjung: The fact that this latch regfile is transparent during write (when clock is low), lets us to have high number of banks and minimize the writeback bank conflicts without adding tons of bypass paths. Because there are changes in the bypass logic/paths, there is no simple drop-in replacement solution here. Is it impossible to have latches on FPGA? I know there are lots of articles about how accidental uses of latches are bad, but the usage is very intentional in this case,

v/vanilla_bean/fcsr.v
tommydcjung: fflags_v_i[1] used to be where fdiv and fpu_float accumulate exceptions, since only one of them used to be able to write back per cycle. But now it is possible for both to writeback at the same time, as long as there is no bank conflict, this fflags_v_i needs to be separated.

v/vanilla_bean/vanilla_core.v
tommydcjung: I removed them , because stall_idiv_wb and stall_fdiv_wb are always zero.

tommydcjung:
merge with: https://github.com/bespoke-silicon-group/basejump_stl/pull/524

PR #604, opened 2021-11-10
turning off icache during reset

tommydcjung:
When vanilla core is in reset (by the freeze register), it should still allow remote packets to be written into the icache, but turn off the enable pin while not being written.

### DIFF
--- a/v/vanilla_bean/vanilla_core.v
+++ b/v/vanilla_bean/vanilla_core.v
@@ -1227,7 +1227,7 @@ module vanilla_core
     : 1'b1;
 
   assign icache_v_li = icache_v_i | ifetch_v_i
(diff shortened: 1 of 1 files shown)


taylor-bsg:
So the point is, we were burning energy when a tile was frozen; and now we are not; but we need to allow the special case when we are writing instructions in?

tommydcjung:
> So the point is, we were burning energy when a tile was frozen; and now we are not; but we need to allow the special case when we are writing instructions in?

Yes

taylor-bsg (commented):
Oh what happened -- did you find an issue, or integrate the code change elsewhere?

tommydcjung:
Yeah I integrated in the instruction expander branch that I'm currently working on.

PR #606, opened 2021-11-22
Replacing parameters with localparams

dpetrisko:
Some tools (genus) and flows (ILM) like to have all parameters in name to resolve, otherwise subtle differences in instantiation can cause name mismatches. This causes all default parameters to be included in the name and makes the name extremely long and complex. For instance, the manycore tile actually breaks the filename length limit for Unix.

This PR replaces "localparameters" with localparam. In BP this also helped compile times significantly, but that's a secondary benefit. @taylor-bsg I don't think there are any tools we care about that can't handle localparams in module headers. Icarus Verilog is the only one I would guess, and there's no way the rest of bsg_manycore is going through that anyway

### DIFF
--- a/v/bsg_cache_dma_to_wormhole.v
+++ b/v/bsg_cache_dma_to_wormhole.v
@@ -24,10 +24,10 @@ module bsg_cache_dma_to_wormhole
     , `BSG_INV_PARAM(wh_len_width_p)
     , `BSG_INV_PARAM(wh_cord_width_p)
     
-    , parameter data_len_lp = (vcache_data_width_p*vcache_block_size_in_words_p/vcache_dma_data_width_p)
+    , localparam data_len_lp = (vcache_data_width_p*vcache_block_size_in_words_p/vcache_dma_data_width_p)
 
-    , parameter dma_pkt_width_lp=`bsg_cache_dma_pkt_width(vcache_addr_width_p)
-    , parameter wh_link_sif_width_lp = 
+    , localparam dma_pkt_width_lp=`bsg_cache_dma_pkt_width(vcache_addr_width_p)
+    , localparam wh_link_sif_width_lp = 
       `bsg_ready_and_link_sif_width(wh_flit_width_p)
   )
   (
--- a/v/bsg_cache_dma_to_wormhole.v
+++ b/v/bsg_cache_dma_to_wormhole.v
@@ -26,8 +26,8 @@ module bsg_cache_dma_to_wormhole
     
     , localparam data_len_lp = (vcache_data_width_p*vcache_block_size_in_words_p/vcache_dma_data_width_p)
 
-    , localparam dma_pkt_width_lp=`bsg_cache_dma_pkt_width(vcache_addr_width_p)
-    , localparam wh_link_sif_width_lp = 
+    , dma_pkt_width_lp=`bsg_cache_dma_pkt_width(vcache_addr_width_p)
+    , wh_link_sif_width_lp = 
       `bsg_ready_and_link_sif_width(wh_flit_width_p)
   )
   (
--- a/v/bsg_manycore_accel_default.v
+++ b/v/bsg_manycore_accel_default.v
@@ -4,7 +4,7 @@
 module bsg_manycore_accel_default 
   import bsg_manycore_pkg::*;
   import bsg_vanilla_pkg::*;
-   #(parameter `BSG_INV_PARAM(x_cord_width_p   )
+   #(`BSG_INV_PARAM(x_cord_width_p   )
      , `BSG_INV_PARAM(y_cord_width_p )
      , `BSG_INV_PARAM(pod_x_cord_width_p )
      , `BSG_INV_PARAM(pod_y_cord_width_p )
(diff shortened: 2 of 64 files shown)


taylor-bsg:
Interesting -- how things have changed! So localparam in arguments is supported by DC, VCS, Surelog, Verilator, and Vivado?

dpetrisko:
And genus/xcelium ;)

BP uses them heavily, so if they weren’t supported it would have come up for anything we run BP through

taylor-bsg:
I think rather than placing lots of parameter keywords in many places, you can just use the parameter keyword in the next parameter after you add a localparam, to switch it back?

dpetrisko:
I like the explicitness of having the parameter keyword but it’s up to you. The error is very unintuitive, so not sure it’s worth the clarity hit for future maintainers

dpetrisko:
Think this is good to go once CI passes

v/bsg_cache_dma_to_wormhole.v
```diff
@@ -14,20 +14,20 @@ module bsg_cache_dma_to_wormhole
   import bsg_cache_pkg::*;
   import bsg_manycore_pkg::*;
   #(`BSG_INV_PARAM(vcache_addr_width_p)
```
tommydcjung, line 16: missing "parameter" on the first ones.

v/bsg_manycore_hor_io_router.v
```diff
@@ -17,26 +17,26 @@ module bsg_manycore_hor_io_router
...
+    , parameter `BSG_INV_PARAM(tieoff_west_p)
+    , parameter `BSG_INV_PARAM(tieoff_east_p)
     , parameter tieoff_proc_p=0
 
-    , parameter dims_lp=3 // only support 3
+    , localparam dims_lp=3 // only support 3
 
     , parameter fwd_use_credits_p = 7'b0000000
-    , parameter int fwd_fifo_els_p[dims_lp*2:0] = '{2,2,2,2,2,2,2}
+    , localparam int fwd_fifo_els_p[dims_lp*2:0] = '{2,2,2,2,2,2,2}
```
tommydcjung, line 33: This is not localparam

tommydcjung (changes requested):
Left feedback that must be addressed before merging.

taylor-bsg:
Yes, let's remove the redundant localparam and parameters. Unfortunately, it is things like this that kill our code size metrics versus other alternative HDL's (something that is particularly relevant for BlackParrot)

v/bsg_cache_dma_to_wormhole.v
dpetrisko: fixed

v/bsg_manycore_hor_io_router.v
dpetrisko: good catch, this file isn't compiled normally so it wasn't caught

v/bsg_manycore_accel_default.v
```diff
@@ -5,41 +5,41 @@ module bsg_manycore_accel_default
   import bsg_manycore_pkg::*;
   import bsg_vanilla_pkg::*;
    #(`BSG_INV_PARAM(x_cord_width_p   )
```
tommydcjung, line 7: There are all these first parameter in each module missing the keyword "parameter"

dpetrisko: technically, the first one isn't required because the list starts as parameter by default

taylor-bsg: I think it's okay to omit the first one. But can we go ahead and remove the redundant parameter and localparam statements? I still see a bunch in the code

tommydcjung: then we can remove parameter on the second one? Or put the parameter on the first one to be more clear.

dpetrisko: I just changed the parameter to localparam so no net change in keywords. If we want to code golf, we could have all parameters and then all localparams? Then it’s only 1 “localparam” per module

dpetrisko: @taylor-bsg Just pushed a version that's minified without grouping. I think anything further should be its own PR

v/bsg_manycore_tile_compute_array_ruche.v
```diff
@@ -64,14 +64,14 @@ module bsg_manycore_tile_compute_array_ruche
     // coordinate within a pod
     // my_x/y_i
```
tommydcjung, line 65: Can you also fix y/x_cord_width_p, pod_y/x_cord_width_p in this file?

tommydcjung:
these files are deprecated and no longer in use. These should be either reverted or deleted.
```
v/bsg_ruche_anti_buffer.v
v/bsg_manycore_rocc_wrapper.v
v/bsg_manycore_rocc_streambuf_wrapper.v
v/bsg_manycore_packet_streamer.v
v/bsg_manycore_mesh.v
v/bsg_manycore_links_to_fsb.v
v/bsg_manycore_link_to_rocc.v
```

tommydcjung (approved):

v/bsg_manycore_tile_compute_array_ruche.v
dpetrisko: fixed!

dpetrisko:
>these files are deprecated and no longer in use. These should be either reverted or deleted.

Reverted and added #609

PR #614, opened 2022-01-06
Added missing inline

drichmond:
If this file is included in multiple .c files that are independently
compiled into .o files, the linker will complain about duplicate
symbols.


mrutt92:
What you actually want here is to use "static"

Any function defined in a header file should be marked static for exactly this reason.

Use of inline is not guaranteed to work.

drichmond:
Should all of the functions in the header be static, then?

I just mimic'd the rest of them.

mrutt92:
10000000000% yes

mrutt92:
BTW: I'm not saying inline is wrong, you just also need to mark them static.

drichmond:
Fixed, waiting for CI re-run

mrutt92 (approved):

PR #615, opened 2022-01-06
WIP: Added amomin, amomax, amominu, amomaxu to Core & Caches

drichmond:
I thought I needed these for nbody, and did not. Work in progress
because these are not tested or validated. Perhaps someone can build
off of my work

WIP and DNM until someone needs them.


v/bsg_manycore_link_to_cache.v
```diff
@@ -157,7 +157,7 @@ module bsg_manycore_link_to_cache
       e_cache_op: begin
         return_pkt_type = e_return_credit;
       end
-      e_remote_amoswap, e_remote_amoor, e_remote_amoadd: begin
+      e_remote_amoswap, e_remote_amoor, e_remote_amoadd, e_remote_amomin, e_remote_amomax, e_remote_amominu, e_remote_amomaxu: begin
```
tommydcjung, line 160: You also need to assign opcode in the case statement starting at line 285

taylor-bsg:
These seem useful!

M

drichmond:
I agree, but I don't have an immediate use case (though I have an idea for how this might help the MCS lock). 

Do you want me to push it through? Happy to address the comment above and write a few test cases

v/bsg_manycore_link_to_cache.v
drichmond: Fixed

taylor-bsg:
Looks interesting, needs some tests too

drichmond:
Yep. I just addressed Tommy's comment while I had a moment.

PR #617, opened 2022-01-15
Icache instr 30b

tommydcjung:


v/vanilla_bean/bsg_vanilla_defines.vh
```diff
@@ -11,16 +11,16 @@
 
 `include "bsg_defines.v"
 
-`define declare_icache_format_s(tag_width_mp, block_size_in_words_mp) \
+`define declare_icache_format_s(tag_width_mp, block_size_in_words_mp, instr_width_mp) \
```
dpetrisko, line 14: You could also pass in instruction_s as an _mp if that would help with usage

v/vanilla_bean/vanilla_core.v
```diff
@@ -772,14 +776,13 @@ module vanilla_core
...
 
-  // In the icache, branch instruction has the direction of the branch encoded in the bit-0 of the instruction.
   // 0 = forward branch (always predict 'not taken')
   // 1 = backward branch (always predict 'taken')
   // 'branch underpredict' means that branch was predicted to be "not taken", but actually needs to be taken.
   // 'branch overpredict' means that branch was predicted to be "taken", but actually needs to be not taken.
   // In either cases, the frontend should be flushed. 
-  wire branch_under_predict = (alu_jump_now & ~exe_r.instruction[0]);
-  wire branch_over_predict  = (~alu_jump_now & exe_r.instruction[0]); 
+  wire branch_under_predict = (alu_jump_now & ~exe_r.branch_imm_sign);
```
dpetrisko, line 784: maybe call branch_imm_sign br_pred? In case a future scheme is not BTFNT?

v/bsg_manycore_pod_ruche_array.v
```diff
@@ -31,7 +31,7 @@ module bsg_manycore_pod_ruche_array
     , `BSG_INV_PARAM(icache_entries_p)
     , `BSG_INV_PARAM(icache_tag_width_p)
     , `BSG_INV_PARAM(icache_block_size_in_words_p)
-
+    , icache_instr_30b_p = 0
```
dpetrisko, line 34: maybe make a testbench parameter instead of default?

dpetrisko: also might be good to have a comment explaining the area / custom space tradeoff, either here or at the usage in I$

dpetrisko (approved):
LGTM! A few random suggestions

taylor-bsg:
Instead of threading the parameter through, can we just hide the "11" generation inside the implementation of the SRAM?
have we confirmed that we don't have any instructions that use those bits?

dpetrisko:
>Instead of threading the parameter through, can we just hide the "11" generation inside the /hard implementation of the SRAM?
have we confirmed that we don't have any instructions that use those bits?

I think it's good to have a parameter, since there should be a check -- once we use those bits for some instruction -- that you don't use that class of instruction while the parameter is set. Otherwise, you'll just get in a situation where RTL passes, but synth fails and that's never a good position to be in...

taylor-bsg:
In the /hard implementation of the RAM, we can verify that the low two bits are always written? That seems better than what we have now, where we actually don't check anything.

dpetrisko:
I think we’d be losing some valuable abstraction, since the hardened rams usually don’t have any “application logic” in them and can just be straight replaced by whatever flow the user likes. This would require somewhat esoteric knowledge of RISC-V and the system to take advantage of, whereas encoding it in the source RTL is more clear about the tradeoff.

taylor-bsg:
I think longer term, the solution is to pass a configuration enum in, just
like with BP, to avoid endless proliferation of parameters. But time is
short; I think we can do the 11 check hack for now.

M

taylor-bsg:
That is what makes sense to me..

tommydcjung:
That works for me too

dpetrisko:
new version LGTM

tommydcjung:
I checked that the software stack does not write any opcodes not ending in 2'b11, but there could be cases where the host is zeroing out the icache with 32'b0, which is considered as nop.

taylor-bsg (commented):
Check synth for latches =)

taylor-bsg:
@tommydcjung is 30'b0, 11 also a nop?

tommydcjung:
No, 0000011 is LOAD opcode.

dpetrisko:
So looks like 000000 is a nop because it doesn't match any opcodes in the system i.e. illegal instruction. We could make the initialization be 1111111 instead, which should similarly decode to nop?

taylor-bsg:
Is 0^30 11 effectively noop on manycore? I assume it maps to LW $0,0($0),
which hits in DMEM?

But actually, I thought 0 was illegal instruction, not noop, in RISC-V?

M

dpetrisko:
'0 is an illegal instruction, but manycore does not trap on illegal. Instead, the decoder simply doesn't match any instructions, so the decode falls through to '0 control signals

dpetrisko:
(Canonical nop is addi, x0, 0)

taylor-bsg:
If Tommy can confirm that 0^30 11 behaves as I said then we are good…

tommydcjung:
Yes it will just hit DMEM and write back to x0. But the program shouldn't be executing those zero instructions added  to align the code any way, so we are good.

taylor-bsg:
Okay sounds good.

PR #619, opened 2022-01-27
mem_addr_op2 width opt

tommydcjung:
Removing unnecessary flops that store sign-extended bits.

### DIFF
--- a/v/vanilla_bean/lsu.v
+++ b/v/vanilla_bean/lsu.v
@@ -34,7 +34,7 @@ module lsu
     , input [data_width_p-1:0] exe_rs1_i
     , input [data_width_p-1:0] exe_rs2_i
     , input [reg_addr_width_lp-1:0] exe_rd_i
-    , input [data_width_p-1:0] mem_offset_i
+    , input [11:0] mem_offset_i
     , input [data_width_p-1:0] pc_plus4_i
     , input icache_miss_i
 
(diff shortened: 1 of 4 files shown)


v/vanilla_bean/lsu.v
```diff
@@ -34,7 +34,7 @@ module lsu
     , input [data_width_p-1:0] exe_rs1_i
     , input [data_width_p-1:0] exe_rs2_i
     , input [reg_addr_width_lp-1:0] exe_rd_i
-    , input [data_width_p-1:0] mem_offset_i
+    , input [11:0] mem_offset_i
```
dpetrisko, line 37: Maybe use a RV32_imm_width_gp constant for clarity?

v/vanilla_bean/vanilla_core.v
```diff
@@ -480,16 +480,11 @@ module vanilla_core
...
-    | id_r.decode.is_amo_op;
-
-  wire [data_width_p-1:0] mem_addr_op2 = is_amo_or_lr_op
-    ? '0
-    : (id_r.decode.is_store_op
-      ? `RV32_signext_Simm(id_r.instruction)
-      : `RV32_signext_Iimm(id_r.instruction));
-
+  wire [11:0] mem_addr_op2 = id_r.decode.is_store_op
+    ? 12'(`RV32_signext_Simm(id_r.instruction))
```
dpetrisko, line 484: Since we're not using the signext here, we could define a macro without it. But if this synthesizes the same, not required

dpetrisko (approved):
LGTM, clarity suggestions

v/vanilla_bean/vanilla_core.v
```diff
@@ -480,16 +480,11 @@ module vanilla_core
...
+    : (id_r.decode.is_load_op
```
dpetrisko, line 485: Just to make sure, AMO/LR is not considered load or store ops in this decoder?

tommydcjung: AMO / lr has its own flag.

PR #620, opened 2022-01-28
Energy benchmark update

tommydcjung:
energy_lr_aq_test and other updates


drichmond:
LGTM, but what's the reason for using packets instead of an architecturally idempotent instruction?

tommydcjung:
Because you can't put a bind module on a post-apr netlist to detect this instruction being executed. Non post-apr approach is not accurate.

drichmond (approved):

drichmond:
Got it, SGTM.

There will be some loss in fidelity, but you know that.

tommydcjung:
That's why it runs many instructions to take an average.

PR #624, opened 2022-02-02
Moves File Open/Close Operations in Trace Profiler

mrutt92:
Currently the tracers call $fopen and $fclose inside of always blocks in modules that are instantiated many times.
When the trace is enabled this results in $fopen/$fclose being called redundantly many times per cycle.
I found that calling $fopen and $fclose just once at the start of the simulation reduced sim time by 1.6x (average) and as much as 2x.

### DIFF
--- a/testbenches/common/v/profiler.hpp
+++ b/testbenches/common/v/profiler.hpp
@@ -9,6 +9,7 @@ public:
     }
 
     // members
+    int is_init;
     int is_exit;
     int trace_file;
 };
@@ -18,30 +19,30 @@ public:
     extern "C" void profiler_name ## _init(             \
         int _trace_file                                 \
         ) {                                             \
-        if (profiler_name == nullptr) {                 \
-            profiler_name = new bsg_profiler::profiler; \
-            profiler_name->trace_file = _trace_file;    \
+        if (profiler_name.is_init == 0) {               \
+            profiler_name.trace_file = _trace_file;     \
         }                                               \
     }
 
 #define PROFILER_EXIT_FUNC(profiler_name)            \
     extern "C" void profiler_name ##_exit() {        \
+        profiler_name.is_exit = 1;                   \
         return;                                      \
     }
 
 #define PROFILER_IS_INIT_FUNC(profiler_name)        \
     extern "C" int profiler_name ## _is_init() {    \
-        return (profiler_name != nullptr);          \
+        return profiler_name.is_init;               \
     }
 
 #define PROFILER_IS_EXIT_FUNC(profiler_name)        \
     extern "C" int profiler_name ##_is_exit() {     \
-        return profiler_name->is_exit;              \
+        return profiler_name.is_exit;               \
     }
 
 #define PROFILER_TRACE_FD_FUNC(profiler_name)           \
     extern "C" int profiler_name ##_trace_fd() {        \
-        return profiler_name->trace_file;               \
+        return profiler_name.trace_file;                \
     }
 
 #define DEFINE_PROFILER(profiler_name)          \
--- a/testbenches/common/v/profiler.hpp
+++ b/testbenches/common/v/profiler.hpp
@@ -1,4 +1,5 @@
 #pragma once
+#include <mutex>
 namespace bsg_profiler {
 class profiler {
 public:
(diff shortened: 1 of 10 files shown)


testbenches/common/v/vcache_profiler.v
```diff
@@ -336,19 +336,17 @@ module vcache_profiler
 
 
         if (~reset_i & trace_en_i) begin
-          trace_fd = $fopen(tracefile_lp, "a");
-
           // If miss handler has finished the dma request and result is ready
           // for a missed request
           if (inc_miss_ld)
-            print_operation_trace(trace_fd, my_name, "miss_ld");
+            print_operation_trace($root.`HOST_MODULE_PATH.vcache_trace_fd, my_name, "miss_ld");
```
tommydcjung, line 342: Can we just make the fd not an argument to print_operation trace? Is there any reason to be different?

mrutt92: Not that I can tell. Seems reasonable to change.

testbenches/common/v/spmd_testbench.v
```diff
@@ -193,6 +193,39 @@ module spmd_testbench
     assign trace_en = (trace_arg == 1);
   end
 
+  // vanilla operations trace
```
tommydcjung, line 196: It just  seems weird that the headers are created  here.  Does that mean each testbench toplevel need to replicate this? What if the format changes?

mrutt92: I agree that it's awkward. Maybe a package that defines the header and the print task you mention in your last comment?

mrutt92: I guess the printing of the header itself could be moved back to inside the module, which has its own awkwardness but it's less than making testbenches duplicate more code than necessary I suppose.

The important change is moving the open and close.

tommydcjung: maybe you can just create a static variable on the DPI code.

mrutt92: What would the static variable be? The file descriptor?

tommydcjung: Yeah file descriptor and the variable to check if the header has already been printed.

mrutt92: k. see changes.

mrutt92:
@tommydcjung I addressed your comment, look good to you?

testbenches/common/v/profiler.hpp
```diff
@@ -0,0 +1,52 @@
...
+public:
+    // constructors
+    profiler()
+        : is_exit(0)
+        , trace_file(-1) {
+    }
+
+    // members
+    int is_exit;
+    int trace_file;
```
tommydcjung, line 13: Are we assuming these variables are atomically updated because the simulation is single-threaded?

tommydcjung (approved):

testbenches/common/v/profiler.hpp
mrutt92: I guess that is an assumption I made; I didn't make any of these function re-entrant.

drichmond: Dan might have an interesting input here, having worked with Verilator.

mrutt92: making the init and exit functions thread safe shouldn't be too much work

mrutt92: @dpetrisko are you using parallel simulation?

mrutt92: k. i made these functions thread safe. it doesn't sound like we have any parallel simulation but just in case we ever do, that's one less difficult bug to track down.

tommydcjung (approved):

testbenches/common/v/profiler.hpp
tommydcjung: I think it's fine as long as there is no significant performance overhead, but it seems like we don't actually know if there is a simulator where this mutex is required.

tommydcjung: Maybe it's possible to have one log file for each tile, then you kinda avoid all these problems.

mrutt92: I think there's value to having it be one trace file. It makes the scripts that read this in simpler, and changing it would require rewriting those scripts. 

Whatever overhead from the mutex exists at init and exit time is amortized over the life of simulation so I think the current solution is quite workable.

PR #626, opened 2022-02-04
setting vcache_wh_dest for all pods

tommydcjung:
vcache_wh_dest should be set for every vcache in the pod array, whether or not that its pod have tiles being launched.

### DIFF
--- a/software/py/nbf.py
+++ b/software/py/nbf.py
@@ -414,6 +414,13 @@ class NBF:
   # public main function
   # users only have to call this function.
   def dump(self):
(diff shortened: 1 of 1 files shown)


drichmond (approved):
LGTM!

I don't think we need a corresponding PR in replicant - I believe this code handles it: https://github.com/bespoke-silicon-group/bsg_replicant/blob/e6f11e2367c224d166bd41feb82c776478e4eeb2/libraries/bsg_manycore_vcache.cpp

PR #628, opened 2022-02-07
Credit pkt regifile corrupt test

tommydcjung:
This test recreates the bug where a residual credit pkt in the return FIFO corrupts the regfile state.


dpetrisko (approved):
LGTM!

PR #631, opened 2022-03-08
Add bsg_group_strider, and SPMD test

drichmond:
This is a C++ class for striding between tiles in a tile group.

It can be used to stride between tiles in the horizontal, vertical, or diagonal direction. This can be more than the size of the tile group, but must be static.

This is used in the optimized SGEMM implementation, but is more general than that. Hence, why it is being merged here.

### DIFF
--- a/software/bsg_manycore_lib/bsg_group_strider.hpp
+++ b/software/bsg_manycore_lib/bsg_group_strider.hpp
@@ -18,6 +18,12 @@
 //    bsg_tile_group_strider<bsg_tiles_X, 1, bsg_tiles_Y, 1, int> stride_y(foo, 0, 0);
 //
 // Use the stride() method to get an updated pointer after construction.
+//
+// When the strider reaches the end of a tile group it wraps back
+// around to 0 -- i.e. mod.
+
+// NOTE: THIS STRIDER ONLY WORKS WITH TILE GROUPS THAT ARE A POWER OF
+// TWO IN THE DIMENSION(S) OF STRIDING.
 
 #ifndef __BSG_GROUP_STRIDER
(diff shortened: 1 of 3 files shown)


software/bsg_manycore_lib/bsg_group_strider.hpp
```diff
@@ -0,0 +1,74 @@
+// The intent of this file is to provide a simple, templated method
```
mrutt92, line 1: Is this class an interface for a tile accessing memory that is S_X to the west and and S_Y to the south, where S_X and S_Y are static?

drichmond: East, but yes. 

You can make them negative to move in the other direction.

drichmond: It ends up turning into ~2-3 instructions to move between tile groups.

mrutt92: what happens if you're at the eastern or southern edge of a tile-group?

mrutt92: well... could if you use -1 for S_X, can't you access east not west?

drichmond: Positive moves in the South (vertical) and East (Horizontal) directions. They wrap around when the edge of the tile group is reached.

mrutt92: Neat!

mrutt92 (approved):

mrutt92 (approved):

PR #635, opened 2022-03-23
Refix verilator

dpetrisko:
These are minor fixes to get verilator back and happy. The barrier change is just to stifle errors since those modules are unused and probably broken as is. The profiler change is to get around a feature missing of declaring a local int variable within an initial/final block.

Merge with: https://github.com/bespoke-silicon-group/bsg_replicant/pull/752


drichmond (approved):
LGTM, on all the unused module changes. They are identical to what I did.

Someone else should verify the profiler change.

mrutt92 (approved):

tommydcjung (approved):

PR #643, opened 2022-04-06
Initial xcelium commit

dpetrisko:
This commit adds support for xcelium. It does not use multicore, plusperf, or other performance optimizations. However, it does add dumping and waveform viewing.

It also has a few fixes

### DIFF
--- a/machines/platform.mk
+++ b/machines/platform.mk
@@ -1,5 +1,5 @@
 
-BSG_PLATFORM ?= vcs
+BSG_PLATFORM ?= xcelium
 
 ifeq ($(BSG_PLATFORM),vcs)
 DEFAULT_MACHINES = pod_1x1 pod_1x1_hbm2 pod_4x4 pod_4x4_hbm2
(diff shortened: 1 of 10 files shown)


testbenches/common/v/bsg_manycore_crossbar.v
```diff
@@ -22,9 +22,9 @@ module bsg_manycore_crossbar
 
     // set using the credit interface for the local FIFO for fwd and rev network.
     , parameter `BSG_INV_PARAM(fwd_use_credits_p)
-    , parameter int fwd_fifo_els_p[num_in_lp-1:0] = -1
+    , parameter int fwd_fifo_els_p[num_in_lp-1:0] = '{-1}
```
dpetrisko, line 25: Xcelium doesn't like this default

software/mk/Makefile.verilog
```diff
@@ -21,7 +21,7 @@ ifneq ($(IGNORE_CADENV),)
   # select the tools
 
 # Checking the VCS settings
-else ifeq (,$findstring(cs.washington.edu,$(HOSTNAME)))
+else ifeq (,$(findstring cs.washington.edu,$(HOSTNAME)))
```
dpetrisko, line 24: This was buggy before ($f appended with indstring), should work properly now

machines/xcelium_dump.tcl
```diff
@@ -0,0 +1,4 @@
+database -open dump -shm
```
dpetrisko, line 1: Xcelium doesn't seem to have a command line interface, so we're forced to use TCL here

machines/platform.mk
```diff
@@ -1,11 +1,14 @@
 
-BSG_PLATFORM ?= vcs
+BSG_PLATFORM ?= xcelium
```
drichmond, line 2: I think you need to revert this?

tommydcjung (approved):

dpetrisko:
Thanks!

PR #644, opened 2022-04-15
Faster PC Histogram

mrutt92:
We have a PC histogram tool currently that parses a cycle-by-cycle trace file from simulation.
The functionality is very helpful, but it comes at ~3x cost to simulation time and the trace file can grown to GBs in size for even relatively short simulations.

This PR provides a fast-functional PC histogram. The histogram is stored in memory while simulation is running and dumped to a file at the end of simulation. Some level of detail is lost such as the instruction opcode (instead we see either 'instr' or 'fp_instr'). However, the stall type information is retained.

The overhead from using the fast PC histogram was never more than 20% over fast non-profiling simulation and as much as 3x faster than enabling the trace.

Note that this is unsuitable for blood graph generation. You still need to use tracing for that.


testbenches/common/v/vanilla_core_pc_histogram.v
```diff
@@ -0,0 +1,232 @@
...
+`define VANILLA_CORE_STALL(name, cond, pc) \
+      else if ((cond)) begin \
+        vanilla_core_pc_hist_increment \
+          (pc_hist_vptr \
+           ,pc \
+           ,e_``name \
+           ); \
+            end
+`include "vanilla_core_stall.def"
+`undef VANILLA_CORE_STALL
```
tommydcjung, line 220: Can we just write these out instead of using macros? Though it 's convenient, code readability is one issue. Different places might need different ordering. The stall model might change, and this makes it brittle. I see more problems lying ahead than the minor code reduction benefit.

mrutt92: Sure, I'll make that change before merging.

testbenches/common/v/vanilla_core_stall.def
```diff
@@ -0,0 +1,27 @@
+// stalls that stall entire pipeline
+`VANILLA_CORE_STALL(stall_icache_store, stall_all & stall_icache_store, exe_pc)
+`VANILLA_CORE_STALL(stall_remote_ld_wb, stall_all & stall_remote_ld_wb, exe_pc)
+`VANILLA_CORE_STALL(stall_remote_flw_wb, stall_all & stall_remote_flw_wb, exe_pc)
+`VANILLA_CORE_STALL(stall_ifetch_wait, stall_all & stall_ifetch_wait, exe_pc)
```
tommydcjung, line 5: For stall_ifetch_wait, the relevant pc is in mem stage.

testbenches/common/v/vanilla_core_stall.def
```diff
@@ -0,0 +1,27 @@
+// stalls that stall entire pipeline
+`VANILLA_CORE_STALL(stall_icache_store, stall_all & stall_icache_store, exe_pc)
+`VANILLA_CORE_STALL(stall_remote_ld_wb, stall_all & stall_remote_ld_wb, exe_pc)
+`VANILLA_CORE_STALL(stall_remote_flw_wb, stall_all & stall_remote_flw_wb, exe_pc)
```
tommydcjung, line 4: When there is a remote write back, the pc in exe might actually be invalid (exe has bubble), or some unrelated instruction that has nothing to do with the stall. The more accurate and relevant info would be the PC of the remote load, or the PC of the instruction that causing the contention in WB.

mrutt92: k... so would i use the scoreboard tracker to track this? 
is this tracked correctly in the current tracer?

mrutt92: so would this be the pc in the exe stage when a register is scored?

drichmond:
Very excited to see this merged in

testbenches/common/v/vanilla_core_stall.def
mrutt92: I would benefit from more input on what pc should be shown here. It looks like I need to track pcs associated with idiv and imul instructions?

testbenches/common/v/bsg_nonsynth_manycore_testbench.v
```diff
@@ -913,6 +915,22 @@ if (enable_vanilla_core_trace_p) begin
 end
 `endif
 
+  //////////////////
+  // PC Histogram //
+  //////////////////
```
dpetrisko, line 920: VERILATOR_WORKAROUND_DISABLE_PC_HISTOGRAM

testbenches/common/v/vanilla_scoreboard_tracker_pkg.v
```diff
@@ -0,0 +1,21 @@
...
+  } vanilla_isb_type_e;
+
+  // floating point scoreboard
+  typedef enum logic [31:0] {
+    e_vanilla_fsb_remote_group_load
+    ,e_vanilla_fsb_remote_global_load
+    ,e_vanilla_fsb_remote_dram_load
+    ,e_vanilla_fsb_fdiv_fsqrt
+    ,e_vanilla_fsb_n
+  } vanilla_fsb_type_e;
```
tommydcjung, line 19: Can we make these structs instead?

mrutt92: yes

testbenches/common/v/vanilla_core_stall.def
mrutt92: k, i commented them out and left a TODO comment

mrutt92:
@tommydcjung anything else? remove the macros and then good-to-go?

testbenches/common/v/vanilla_core_stall.def
mrutt92: it's counted as an unknown stall at exe_pc

drichmond: OK. Works for me. 

Is there some message that says "unknown stalls were encountered, this may be because ..."?

mrutt92: Not so easy to print it once. Not super hard to print it one per instance. 

I think it makes more sense to report the stall type with a not super useful pc than it does to report it as 'unknown.'
I'm happy to iterate on finding a better pc to report for wb stalls, but i think the merits of this PR easily meet the bar for merging. It's output is at least as correct as the current tracer's and it's 3x faster.

Finding a better pc for write back stalls seems to me like enough work to merit it's own PR.

mrutt92: I guess my pc reporting comment was more directly in response to feedback from @tommydcjung. 

But following that feedback results in more of these 'unknown' stalls, so it's not unrelated to your suggestion of a print message. It sounds like you agree with me that reporting the stall type with the current exe_pc is more helpful than reporting unknown?

testbenches/common/v/vanilla_scoreboard_tracker.v
```diff
@@ -0,0 +1,97 @@
...
+      // clear scoreboards
+      if (int_sb_clear) begin
+        int_sb_r[int_sb_clear_id] <= '0;
+      end
+      if (float_sb_clear) begin
+        float_sb_r[float_sb_clear_id] <= '0;
+      end
+
+    end
+  end // always_ff @ (posedge clk_i)
```
tommydcjung, line 92: Make sure to copy and paste this logic  from the latest source. there has been a fix to this recently.

mrutt92: I saw that: I also made a change to address the scoreboard clearing. The changes I saw looked like more code to accomplish the same thing? 

Question: it's a bug in the scoreboard module to clear and score the same register in the same cycle right?

tommydcjung: Yeah the issue is that the ordering is not very explicit with the way these non-blocking assignments are written. If both conditions are met which one is supposed to execute? In the fix, the set is supposed to override the clear.

tommydcjung:
> @tommydcjung anything else? remove the macros and then good-to-go?

Can we remove the macros so we can look at the resulting code? we might find more issues then

mrutt92:
> >  @tommydcjung anything else? remove the macros and then good-to-go?
> Can we remove the macros so we can look at the resulting code? we might find more issues then
Are we done reviewing everything else?

testbenches/common/v/vanilla_core_stall.def
drichmond: Yes. I do want to know which PC caused it, but I think the data of where the stall happened is more useful. I don't know where the stall happened if I just know what PC caused it.

When I think of SGEMM, where these stalls happen, I want to answer the question: are my stalls happening at the tail of my FMA sequence? At the start? At the end? Those questions tell me something fundamental about read latencies in the network for a given access pattern and kernel. To me, that is actionable data.

Is the compiler spreading across ports correctly? What matters is that loads were coming back and conflicting with other writes to the register file. 

Using this information I could adjust the schedule and register allocation of my code. Potentially.

drichmond: We can add a caveat to the wb stalls that they are not necessarily caused by the PC that the stall is being attributed to. Instead, this PC is just is where it was observed.

drichmond: For posterity, in SGEMM at the moment, 10% of our stalls are these wb stalls. I would love more insight into why.

tommydcjung: I'm not saying we should report it as unknown. I think it's a good idea to report which remote load PC caused the stall, but  there is also a PC that is affected by it, which can be either exe_pc or exe_bubble_pc. I think remote load PC is more useful info for developers, but that could depend on situation.

mrutt92: What I'm proposing is that we merge this PR reporting the stall types with exe_pc, and then I'll start a new branch to resolve this issue quickly (hopefully within a day or two). I know of a couple people that are waiting on this tool and I don't think this type of stall is a major concern for them.

testbenches/common/v/vanilla_scoreboard_tracker.v
mrutt92: done

testbenches/common/v/vanilla_core_pc_histogram.v
mrutt92: done

mrutt92:
i went ahead and removed the macros.

testbenches/common/v/vanilla_core_stall.def
mrutt92: Let's address this in a different PR.

mrutt92:
Ok good to go?

testbenches/common/v/vanilla_scoreboard_tracker.v
```diff
@@ -0,0 +1,136 @@
...
+        else if (int_sb_clear & (int_sb_clear_id == i)) begin
+          int_sb_r[i].remote_global_load <= 1'b0;
+        end
+        // remote ld group
+        if (~stall_id & ~stall_all & ~flush & remote_ld_group_in_id & (id_rd == i)) begin
+          int_sb_r[i].remote_group_load <= 1'b1;
+        end
+        else if (int_sb_clear & (int_sb_clear_id == i)) begin
+          int_sb_r[i].remote_global_load <= 1'b0;
+        end
```
tommydcjung, line 96: remote_global_load -> remote_group_load

testbenches/common/v/vanilla_scoreboard_tracker_pkg.v
```diff
@@ -0,0 +1,19 @@
+package vanilla_scoreboard_tracker_pkg;
+
+  // integer scoreboard
+  typedef struct packed {
+    logic remote_group_load;
+    logic remote_global_load;
+    logic remote_dram_load;
+    logic idiv;
+  } vanilla_isb_info;
```
tommydcjung, line 9: vanilla_isb_info -> vanilla_isb_info_s

testbenches/common/v/vanilla_scoreboard_tracker.v
```diff
@@ -0,0 +1,136 @@
...
+          float_sb_r[i].remote_group_load <= 1'b0;
+        end
+      end // for (integer i = 0; i < RV32_reg_els_gp; i++)
+    end // else: !if(reset_i)
+  end // always_ff @ (posedge clk_i)
+
+  assign int_sb_o = int_sb_r;
+  assign float_sb_o = float_sb_r;
+endmodule // vanilla_scoreboard_tracker
+
```
tommydcjung, line 136: missing BSG_ABSTRACT_MODULE()?

mrutt92:
good-to-go?

testbenches/common/v/vanilla_core_pc_histogram.v
```diff
@@ -0,0 +1,331 @@
...
+    end
+    else begin
+      if (~stall_all & (exe_bubble_type == e_exe_no_bubble)) begin
+        mem_pc_r <= exe_pc;
+      end
+    end
+  end
+
+  // event signals
+  wire instr_inc = ~(stall_all) * (exe_r.instruction != '0) & ~exe_r.icache_miss;
```
tommydcjung, line 227: * -> &

tommydcjung (approved):
LGTM

testbenches/common/v/vanilla_core_pc_histogram.v
mrutt92: oh man, nice catch

PR #646, opened 2022-05-13
Add amoor and lw prefetch commands

drichmond:
Admittedly one of the more odd PRs that I've submitted.

This PR adds two inline functions: `bsg_amo_prefetch` and `bsg_lw_prefetch`. These functions can be used to prefetch data into the cache for a tile to read later.

`bsg_amo_prefetch` uses `amoor.w`, but with an operand of 0 so it doesn't disturb data. However, the cache doesn't know this so it pollutes the dirty bit. Therefore, `bsg_amo_prefetch` should only be used to verify prefetching behavior. In the statistics parser, correct use of `bsg_amo_prefetch` will drive the load/store miss rate to 0

After verifying behavior with `bsg_amo_prefetch` replace all instances with `bsg_lw_prefetch`

* bsg_amo_prefetch and bsg_lw_prefetch both target x0 to avoid architectural dependencies
* Use amo to verify behavioral functionality of preloads, then lw in performance code since amo marks cache lines as dirty

### DIFF
--- a/software/bsg_manycore_lib/bsg_manycore.h
+++ b/software/bsg_manycore_lib/bsg_manycore.h
@@ -167,6 +167,18 @@ inline void bsg_fence()      { __asm__ __volatile__("fence" :::); }
 
 #define bsg_compiler_memory_barrier() asm volatile("" ::: "memory")
 
+
+// These functions are a simple way to issue pre-fetch commands to the
+// caches.  amo prefetch should not be used in practice since it marks
+// the cache line as dirty. However, it is a great way to verify that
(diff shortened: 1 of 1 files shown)


mrutt92 (commented):
you could have `bsg_prefetch` wrapped by a and ifdef PREFETCH_PROFILE or something

drichmond:
To switch back and forth?

There are times I want to use both. If I have multiple phases of a program, I want to use amo for one phase, and lw for the other.

mrutt92 (approved):

PR #651, opened 2022-05-17
Remote profiler functionality

drichmond:
* Added remote store tracking functionality
* Added (unconnected) wires to network_tx to track returning store credits
* Added print statements to remote_load_trace
* Separated remote load bind instantiation from tracer instantiation

### DIFF
--- a/v/vanilla_bean/network_tx.v
+++ b/v/vanilla_bean/network_tx.v
@@ -95,7 +95,11 @@ module network_tx
     , output logic invalid_eva_access_o
   );
 
+  // Unused wires
   wire unused = reset_i;
+  // These are used for bind/profiling, not synthesis
+  wire unused0 = returned_credit_v_i;
+  wire [RV32_reg_addr_width_gp-1:0] unused1 = returned_credit_reg_id_i;
 
   // manycore packet struct
   //
--- a/v/vanilla_bean/network_tx.v
+++ b/v/vanilla_bean/network_tx.v
@@ -55,6 +55,10 @@ module network_tx
     , input bsg_manycore_return_packet_type_e returned_pkt_type_i
     , input returned_fifo_full_i
     , output logic returned_yumi_o
+
+    // Write response coming back. Unused for synthesis, used in bind statements
+    , input                               returned_credit_v_i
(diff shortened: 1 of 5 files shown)


v/vanilla_bean/network_tx.v
```diff
@@ -55,6 +55,10 @@ module network_tx
     , input bsg_manycore_return_packet_type_e returned_pkt_type_i
     , input returned_fifo_full_i
     , output logic returned_yumi_o
+
+    // Write response coming back. Unused for synthesis, used in bind statements
+    , input                               returned_credit_v_i
+    , input [RV32_reg_addr_width_gp-1:0]  returned_credit_reg_id_i
```
tommydcjung, line 61: Can you add to prevent lint warning?
```
wire unused0 = returned_credit_v_i;
wire [RV32_reg_addr_width_gp-1:0] unused1 = returned_credit_reg_id_i;
```

tommydcjung (approved):

dpetrisko:
Fun trick I like to use:

wire unused = &{returned_credit_v_i, returned_credit_reg_id_i}

v/vanilla_bean/network_tx.v
drichmond: Fixed, and yes, it is used. It is used to track stores.

tommydcjung: Well , it's still using returned_reg_id_i below, but I think they are the same signal anyway

drichmond: Fair. Good catch. That's a mistake

I fixed that, too.

tommydcjung:
What's the reg_id sent by the vanilla core though? sw instruction doesn't have rd field, so I don't think it's possible to track each request uniquely.

drichmond:
I thought sw uses the reg_id field. I thought black parrot required it. 

https://github.com/bespoke-silicon-group/bsg_manycore/blob/17ac63116106598bb98622c8b171d6b842b7a033/v/bsg_manycore_link_to_cache.v#L186

drichmond:
I've removed the write latency tracking code, and removed the unused signals that I added

tommydcjung (approved):

PR #652, opened 2022-05-19
2d mesh pod hierarchy

tommydcjung:
Adding BSG_MACHINE_NETWORK_CFG again to support 2D mesh.


dpetrisko:
Code looks good. Does it make sense to have ruche_factor = 0 decompose into a regular mesh? Could remove some of these mostly identical hierarchies

dpetrisko (approved):

dpetrisko:
Ah, that makes sense. zero-width signals strike again

PR #659, opened 2022-07-05
Update regfile_synth for ultrascale plus fpga mapping

gaozihou:
This PR resolves the regfile mapping problem on the Ultrascale Plus FPGA. currently regfile is mapped as registers instead of LUTRAMs. The mapping problem can be resolved by unpacking the arrays and tweaking the combinational logics.

1. It is functionally identical to the original version (verified with randomly-generated testbench in both VCS and Vivado simulators)
2. ASIC resource utilization change is minimal (-0.1% area) after synthesis in DC
3. LUT utilization on FPGA reduce by 15% in Vivado

### DIFF
--- a/v/vanilla_bean/regfile_synth.v
+++ b/v/vanilla_bean/regfile_synth.v
@@ -42,12 +42,10 @@ module regfile_synth
 
   if (x0_tied_to_zero_p) begin: xz
     // x0 is tied to zero.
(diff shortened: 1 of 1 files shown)


tommydcjung (approved):

gaozihou:
Thanks Tommy!

PR #660, opened 2022-07-13
barrier perf tests

tommydcjung:
Tests to measure AMOADD, HW, tile-group barrier latencies.


mrutt92 (commented):
How are you reading the cycle time with these changes?

tommydcjung:
bsg_print_int prints the time, and the diff was taken manually for the interest of time, but I can add some scripts to do a sweep and parsing.

mrutt92:
Ah got it. Works differently in cuda. Your call on the script. I would do it if you expect someone else to reproduce or update what you've done.

mrutt92 (approved):

PR #662, opened 2022-07-28
bsg_cache word tracking wormhole change

tommydcjung:
- Enabling write-validate cache.
- Updating wormhole test mem, and the testbench,
- Added a new test 'memtest2022'


mrutt92:
what's up with this pr? looks like it's necessary to keep up to date with master in basejump.

dpetrisko (approved):
LGTM! 

- Is it worth it to add a parameter for word tracking, or do we feel the hardware overhead is always justified?
- Have we tested with this same changeset except for word_tracking_p(0), from a functionality perspective?

taylor-bsg:
Hi,

I think it would be helpful to have the parameter that allows us to
enable/disable it, so that bsg_cache can still be used even when word
writes are not available in the memory backing store.

tommydcjung:
- Added a machine parameter 'BSG_MACHINE_VCACHE_WORD_TRACKING'.

PR #667, opened 2022-10-05
Some fixes to vary pod size.

tommydcjung:
- Smallest pod size = X=4, Y=2 (need at least 4 vcaches on each side to make use of all the wormhole links)
- For 7-bit, xy global cord, the largest pod size is X=32, Y=32.
- Upgrading barriers in some tests.


drichmond:
LGTM, assuming CI passes?

drichmond (approved):
merge after CI

PR #669, opened 2022-10-19
Cadenv fix

dpetrisko:
This PR extends a guard for bsg_cadenv (IGNORE_CADENV) to other files. It's useful for building with verilator without bsg_cadenv existing at all. It also describes the flow in a bit more detail for users unfamiliar with git

This goes after #662


dpetrisko:
@tommydcjung Rebased after #662, this good to go?

tommydcjung (approved):

PR #678, opened 2022-11-10
Update bsg_manycore_arch.h

### LINKED ISSUE #661: Logical vs Physical Pod Software Discussion
Hi, we're running into issues when trying to index subpods that are not part of the main manycore array.

Our setup is (simplified):

```
                    [ vcache (0,1)     ]
[foo subpod (1, 0)] [ compute (1, 1) ]
                    [ vcache (2, 1)    ]

```

So to index foo subpod CSR 0x800, we use bsg_global_pod_ptr(px,py,x,y,local_addr) and do bsg_global_pod_ptr(0,1,15,0,0x800). However, this ended up giving an unexpected output address. Looking closer at the macro:

https://github.com/bespoke-silicon-group/bsg_manycore/blob/911ca3ab4c393cf89a320487edc95f2b3a979a0c/software/bsg_manycore_lib/bsg_manycore_arch.h

```
// pod remote pointer
// px = pod id x
// py = pod id y
// x = subcord x
// y = subcord y
#define bsg_global_pod_ptr(px,py,x,y,local_addr) \
  ((bsg_remote_int_ptr) ( (1 << GLOBAL_PREFIX_SHIFT)    \
                        | ((((1+(py*2))*bsg_global_Y)+y) << GLOBAL_Y_CORD_SHIFT)  \
                        | ((((px+1)*bsg_global_X)+x) << GLOBAL_X_CORD_SHIFT)  \
                        | ((int) local_addr)))
```

It appears that bsg_global_pod_ptr indexes some higher abstraction of px,py == the coordinate of compute in the [vcache, compute, vcache unit]. From offline discussion it seems like there is some nomenclature about "logical pod" vs "physical pod" that is missing from the reference doc. Can we clarify that? Secondly, we'll define a macro that just uses the pod coordinates passed directly as:

```
// pod remote pointer
// px = pod id x
// py = pod id y
// x = subcord x
// y = subcord y
#define bsg_global_physical_pod_ptr(px,py,x,y,local_addr) \
  ((bsg_remote_int_ptr) ( (1 << GLOBAL_PREFIX_SHIFT)    \
                        | ((((py)*bsg_global_Y)+y) << GLOBAL_Y_CORD_SHIFT)  \
                        | (((px*bsg_global_X)+x) << GLOBAL_X_CORD_SHIFT)  \
                        | ((int) local_addr)))
```

Whatever the correct nomenclature is, can we have both in the manycore library? 

Minor, but I also suggest refactoring bsg_io_* macros to use the physical pod ptr as well. Currently they use hardcoded macros BSG_IO_X_CORD, and it'll be a bit more clear to use the physical pod ptr specifically?

dpetrisko:
This PR clarifies MC compute "logical" pods vs actual "physical" pods for coordinates. I kept the name the same for bsg_global_pod_ptr to be less controversial, but maybe makes sense to rename/alias to bsg_global_compute_pod_ptr. But I'll leave that up to the team.

Fixes #661

### DIFF
--- a/software/bsg_manycore_lib/bsg_manycore_arch.h
+++ b/software/bsg_manycore_lib/bsg_manycore_arch.h
@@ -58,38 +58,38 @@
 
 // tile-group address pointer 
 #define bsg_remote_ptr(x,y,local_addr) \
-  ((bsg_remote_int_ptr) ( (1 << REMOTE_PREFIX_SHIFT) \
-                        | (y << REMOTE_Y_CORD_SHIFT) \
-                        | (x << REMOTE_X_CORD_SHIFT) \
+  ((bsg_remote_int_ptr) ( (1 << REMOTE_PREFIX_SHIFT)   \
+                        | ((y) << REMOTE_Y_CORD_SHIFT) \
+                        | ((x) << REMOTE_X_CORD_SHIFT) \
                         | ((int) local_addr)))
 
 #define bsg_remote_flt_ptr(x,y,local_addr) \
-  ((bsg_remote_float_ptr) ( (1 << REMOTE_PREFIX_SHIFT) \
-                          | (y << REMOTE_Y_CORD_SHIFT) \
-                          | (x << REMOTE_X_CORD_SHIFT) \
+  ((bsg_remote_float_ptr) ( (1 << REMOTE_PREFIX_SHIFT)   \
+                          | ((y) << REMOTE_Y_CORD_SHIFT) \
+                          | ((x) << REMOTE_X_CORD_SHIFT) \
                           | ((int) local_addr)))
 
 #define CREATE_POINTER_BY_TYPE(type) bsg_remote_ ## type ## _ptr
 
 #define bsg_tile_group_remote_ptr(type,x,y,local_addr) \
-  ((CREATE_POINTER_BY_TYPE(type)) ( (1 << REMOTE_PREFIX_SHIFT) \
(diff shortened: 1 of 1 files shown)


software/bsg_manycore_lib/bsg_manycore_arch.h
```diff
@@ -104,6 +104,16 @@
...
 
+// physical pod remote pointer, corresponds to all pods for px, py
+// px = pod id x
+// py = pod id y
+// x = subcord x
+// y = subcord y
+#define bsg_global_physical_pod_ptr(px,py,x,y,local_addr) \
+  ((bsg_remote_int_ptr) ( (1 << GLOBAL_PREFIX_SHIFT)    \
+                        | ((((py)*bsg_global_Y)+y) << GLOBAL_Y_CORD_SHIFT)  \
+                        | (((px*bsg_global_X)+x) << GLOBAL_X_CORD_SHIFT)  \
```
tommydcjung, line 115: Maybe this hasn't become a problem yet, but these arguments need extra parentheses. if px was "1+1", that expression becomes (1+1*bsg_global_X), which is not the same as 2*bsg_global_X

dpetrisko: kk will fix! I'll go through the file and see if any other macros have similar mistakes

tommydcjung (approved):

dpetrisko:
Waiting for CI, then will merge

PR #686, opened 2023-01-03
Fix host module paths in saif dumper

dpetrisko:
https://github.com/bespoke-silicon-group/bsg_manycore/blob/8b450d997095c84dbfe5a44b83c7ea00713025b3/testbenches/common/v/bsg_nonsynth_manycore_testbench.v#L168

This path broke as a result of this change: #652

### DIFF
--- a/machines/xcelium_dump.tcl
+++ b/machines/xcelium_dump.tcl
@@ -1,4 +1,4 @@
 database -open dump -shm
-probe -create spmd_testbench.testbench.DUT -depth all -all -shm -database dump
+probe -create spmd_testbench.testbench.fi1.DUT -depth all -all -shm -database dump
 run
 exit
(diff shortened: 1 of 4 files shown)


dpetrisko:
Would be nice to keep separate genvar names for halfruche and mesh, but it makes the binds much more complicated. I think fi1 is a good compromise

tommydcjung:
Is it okay if the label names are duplicated?
Can you fix the path for the vcache_profiler as well?
https://github.com/bespoke-silicon-group/bsg_manycore/blob/master/testbenches/common/v/bsg_nonsynth_manycore_testbench.v#L923

dpetrisko:
Yup, got vcache profiler as well. It's fine to have the same label for all branches of an if -- just can't have the same label for two separate ifs. The labels have to uniquely identify modules

tommydcjung (approved):

PR #690, opened 2023-01-26
Fixing enum typecast

dpetrisko:
Minor change, VCS accepts it but Verilator and Xcelium do not. 

By IEEE, I think the expression gets demoted to logic by the intermediate values, but I haven't delved in the specs about it


taylor-bsg:
Hm … what are the error messages? What are the types of each of the inputs
to the ternary?

dpetrisko:
```
%Error-ENUMVALUE: <PATH>/zynq-parrot/cosim/import/bsg_manycore/v/vanilla_bean/vanilla_core.v:1638:17: Illegal implicit conversion to enum 'bsg_vanilla_pkg::frm_e' from 'logic[2:0]' (IEEE 1800-2017 6.19.3)
                                                                                                                                : ... In instance top.top_fpga_inst.DUT.py[0].podrow.px[0].pod.mc_y[0].mc_x[0].mc.y[0].x[0].tile.proc.h.z.vcore
                                                                                                                                : ... Suggest use enum's mnemonic, or static cast
 1638 |   assign fpu_rm = (id_r.instruction.funct3 == eDYN)
```

```  
frm_e fpu_rm;
  assign fpu_rm = frm_e'((id_r.instruction.funct3 == eDYN) // logic [2:0]
    ? frm_r // frm_e
    : id_r.instruction.funct3); // logic [2:0]
```

FWIW, I have had to make similar changes to BP to fix XSIM compatibility over the years, where old versions would segfault with no error (very fun). enum implicit casting seems to be an edge case for lots of tools

PR #694, opened 2023-03-06
[profiler] Dram Stall stats

tommydcjung:
- Adding the counter whether the dram access was sequential (consecutive access based on the same rs1) vs non-sequential.
- Splitting stall_depend_dram_load into atomics, sequential loads, non-sequential loads.


taylor-bsg (approved):
Great feature, did not review carefully, leaf code =)

PR #707, opened 2023-10-28
optimize script for `make tools`

Jasonyanyusong:
1) add paramater `COMPILING_THREADS` to specify how many thread to use

2) spike version changed from v1.0.0 to v1.1.0 (aarch64/arm64 support) current spike tag can not compile on aarch64/arm64 machines since the script in the "old" spike have no idea when checking the machine's arch with result of aarch64 or arm64. In spike v1.1.0, spike fixed this mistake

### DIFF
--- a/software/riscv-tools/Makefile
+++ b/software/riscv-tools/Makefile
@@ -41,7 +41,10 @@ SPIKE_REPO        := riscv-isa-sim
 SPIKE_URL         := https://github.com/riscv/riscv-isa-sim.git
 SPIKE_PATCH       := spike.patch
 SPIKE_GCC_PATCH   := spike-gcc.patch
-SPIKE_TAG         := v1.0.0
+SPIKE_TAG         := v1.1.0
+
+# How many thread you want to use compiling tools
+COMPILING_THREADS := 6
 
 LLVM_DIR          := $(CURDIR)/llvm
 export PATH:=$(DEPENDS_DIR)/bin:$(PATH)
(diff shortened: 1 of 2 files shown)


software/riscv-tools/Makefile
```diff
@@ -113,7 +128,7 @@ checkout-spike:
 	@rm -rf $(SPIKE_REPO);
 	git clone --recursive $(SPIKE_URL);
 	cd $(SPIKE_REPO) && git checkout tags/$(SPIKE_TAG)
-	cd $(SPIKE_REPO) && git apply ../$(SPIKE_PATCH) && git apply ../$(SPIKE_GCC_PATCH)
+	#cd $(SPIKE_REPO) && git apply ../$(SPIKE_PATCH) && git apply ../$(SPIKE_GCC_PATCH)
```
dpetrisko, line 131: I think we still want the SPIKE_PATCH for PK modifications

software/riscv-tools/Makefile
```diff
@@ -41,7 +41,22 @@ SPIKE_REPO        := riscv-isa-sim
...
+        NUMPROC := $(shell grep -c ^processor /proc/cpuinfo)
+else ifeq ($(OS),Darwin)
+        NUMPROC := $(shell sysctl hw.ncpu | awk '{print $$2}')
+endif
+
+# Only take half as many processors as available
+COMPILING_THREADS := $(shell echo "$(NUMPROC)/2"|bc)
+
+ifeq ($(COMPILING_THREADS),0)
+        COMPILING_THREADS = 1
```
dpetrisko, line 58: I would do 

COMPILING_THREADS ?= $(shell echo "($(NUMPROC)+1)/2"|bc)


which will 1. allow users to override from command line and 2 handle the 1 case more elegantly

dpetrisko (changes requested):

dpetrisko:
Superceded by #725

PR #709, opened 2023-11-10
Update bsg_manycore_link_to_cache.v

dpetrisko:
The current code throws an error if we write with a store word instead of store mask


dpetrisko:
rebased!

tommydcjung (approved):

PR #711, opened 2023-11-17
v to sv migration

BrendenPage:
Transitions all file extensions to sv and svh from v and vh, and updates the internal references to these files.


dpetrisko (approved):
LGTM!

PR #712, opened 2023-11-19
Rv64

Jasonyanyusong:
1) Migrated ISA, rv32imaf -> rv64imaf
2) Migrated ABI, ilp32f -> lp64f
3) Compile and build program using the 64 bit toolchain
4) Toolchain configured to strictly use the medium-any memory model


software/py/bsg_manycore_link_gen.py
```diff
@@ -235,7 +235,7 @@ def script(self):
       # .text section virtual address starts at 0x0 but
       # loaded at 0x80000000
       if sec == ".text.dram":
-        vaddr = "0x0"
+        vaddr = "0x80000000"
```
dpetrisko, line 238: should be parameterized based on glen

dpetrisko: with a comment about PC width requirements

software/riscv-tools/Makefile
```diff
@@ -27,8 +27,8 @@
 #
 .DEFAULT_GOAL := help
 
-TARGET_ARCH       := rv32imaf
-TARGET_ABI        := ilp32f
+TARGET_ARCH       := rv64imaf
```
dpetrisko, line 30: need to build with multilib I assume? I don't believe you can compile rv32 with an rv64 toolchain otherwise

dpetrisko:
Think this is missing NBF.py modifications -- have you tested the 32b configurations after these changes?

software/py/bsg_manycore_link_gen.py
Jasonyanyusong: This is changed in a new commit

machines/pod_1x1_4X2Y/Makefile.machine.include
```diff
@@ -1,6 +1,8 @@
 BSG_MACHINE_PODS_X									  = 1
 BSG_MACHINE_PODS_Y									  = 1
 
+BSG_MACHINE_XLEN                      = 32
```
dpetrisko, line 4: Set this for all other machines

Jasonyanyusong: Done, please look at the latest commit, thanks!

PR #715, opened 2023-12-07
IPOLY hashing

tommydcjung:
- Deprecating num_vcache_rows_p since there is no proven worth and is difficult to maintain
- Add IPOLY hashing parameter


dpetrisko:
Maybe makes sense to change 
BSG_MACHINE_IPOLY_HASHING
to
BSG_MACHINE_HASHING_TYPE = 0 (LINE), 1 (IPOLY), 2 (ETC)
?
Otherwise looks good to me

tommydcjung:
If there is more than one hashing that is much better, it might make sense to add an enum for it. But ipoly is shown to be pretty well-rounded, so I think this is fine for now.

dpetrisko (approved):
Kk!

PR #725, opened 2024-07-27
External tools for bsg_manycore

dpetrisko:
Adding ability to use external tool install as well as more robust cloning


dpetrisko:
I was overriding that path for my testing, good catch. Fixed!

tommydcjung (approved):

PR #731, opened 2024-11-22
Linker move read-only data from DRAM to DMEM

gaozihou:
This PR moves `.rodata','.rodata*','.srodata.cst16','.srodata.cst8', '.srodata.cst4', '.srodata.cst2','.srodata*'` from DRAM to DMEM. The `hb_hammerhead/blackscholes` test used to use this linker.

- Pros: Useful when we want read-only data stored locally, or when DRAM is unavailable
- Cons: Will take up more space in DMEM, especially when rodata is large

@tommydcjung Do you think it is okay to move all rodata to DMEM on the master branch?

### DIFF
--- a/software/py/bsg_manycore_link_gen.py
+++ b/software/py/bsg_manycore_link_gen.py
@@ -155,7 +155,9 @@ class bsg_manycore_link_gen:
       # <output section>: [<input sections>]
       ['.text.dram'        , ['.text.interrupt', '.crtbegin','.text','.text.startup','.text.*']],
       # bsg-tommy: 8 bytes are allocated in.dmem.interrupt for interrupt handler to spill registers.
-      ['.dmem'             , ['.dmem.interrupt', '.dmem','.dmem.*']],
+      ['.dmem'             , ['.dmem.interrupt', '.dmem','.dmem.*',
+                              '.rodata','.rodata*','.srodata.cst16','.srodata.cst8',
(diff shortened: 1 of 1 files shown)


dpetrisko:
Probably should have a flag since it’s benchmark-specific?

gaozihou:
> Probably should have a flag since it’s benchmark-specific?

I guess this flag is more of platform/machine specific? Since all software running on the real system would use this linker. The benchmark I mentioned is just for reference, it is not the motivation of this PR :)

gaozihou:
@dpetrisko Currently all the small read-only data are stored on DRAM.. I am not sure whether this is the preferred behavior? For example, whenever I specify a uint32 address that never changes (w/ or w/o const), this uint32 data would be forced to store onto DRAM..

dpetrisko:
I say benchmark-specific because if there's benchmark with large RO data that doesn't fit in the DMEM, we'd want to link into DRAM?

dpetrisko:
I guess it's also a matter of how often the RO data is read. If it's just once at the beginning of the program DRAM is probably fine

taylor-bsg:
This is likely to break a lot of stuff... DMEM is supposed to be explicitly owned by the programmer. If you have strings etc, this is going to fill the DMEM up with them. 

So I agree with Dan, this is probably a benchmark specific kind of thing.

gaozihou:
See PR #735

PR #733, opened 2024-11-26
switch between amoadd and hw barrier

tommydcjung:


tommydcjung:
Should the function name changed from "bsg_barrier_hw_tile_group_sync" to "bsg_barrier_tile_group_sync"?

gaozihou:
Thanks @tommydcjung !

- Do we need to add the flag to this Makefile as well? https://github.com/bespoke-silicon-group/bsg_manycore/blob/master/software/mk/Makefile.builddefs
- I guess we may keep the `bsg_barrier_hw_tile_group_sync` name as-is for now? Seems like the name change would incur 200+ changes on the bsg_replicant and hb_hammerbench repos.. The dependencies among repos would be hard to manage

dpetrisko:
Maybe make bsg_barrier_hw_tile_group_sync a deprecated wrapper for bsg_barrier_tile_group_sync? That way people could make a PR to update if they want, but it won't break existing code?

gaozihou (approved):
I confirm this PR is working in both cases (with and without ENABLE_HW_BARRIER defined) in RTL simulation

PR #734, opened 2024-12-02
Use bash shell in riscv-tools Makefile

jackhumphries:
The Makefile currently uses &> which works in bash, but not other shells. On Ubuntu 20.04.6 LTS, `make` uses /bin/sh, which is a symlink to dash. dash does not support this syntax.
    
This PR explicitly sets the shell to bash in the Makefile.

Issue reported in https://github.com/black-parrot-hdk/zynq-parrot/issues/106.

### DIFF
--- a/software/riscv-tools/Makefile
+++ b/software/riscv-tools/Makefile
@@ -73,7 +73,7 @@ export CC=gcc
 export CXX=g++
 export SED=sed
 export PATH
-export SHELL:=$(SHELL)
+export SHELL:=/bin/bash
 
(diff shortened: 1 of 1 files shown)


dpetrisko:
Another option is adding SHELL := /bin/bash in the manycore makefiles. Likely there are other assumptions about bash built-in

jackhumphries:
That's a better solution. I updated this PR accordingly (though only did it for the riscv-tools Makefile).

jackhumphries:
Is it ok to merge this?

dpetrisko (approved):

PR #737, opened 2025-02-04
fix wh dest cid

tommydcjung:


gaozihou (approved):
Looks good! SPMD vcache test is passing with this fix

PR #739, opened 2025-02-05
adding global_xy_i ports

tommydcjung:
Adding global_x/y_i ports to the pod module;
So that coordinates are injected at the top-level, and avoid creating extra space above IO router tiles to place tieoff cells;


gaozihou (approved):
LGTM

PR #742, opened 2025-02-27
[WIP] Add uncached IO instructions

Yuan-Mao:
This PR aims to provide two new custom RISC-V instructions - uncached read and uncached write.

* The proxy vcaches for the uncached IO requests are currently the whatever the Vcache a tile is nearest to. This is reflected in network_tx.sv.

* This work has been tested with a simple newly added spmd test that does an uncached write and then uncached read. Since it hasn't been interfaced with the Vcache side of work, module bsg_manycore_link_to_cache hasn't been changed and the current behavior is mapping the uncached read/write back to regular read and write cache op in bsg_manycore_link_to_cache. This module will be changed once we start interfacing with the Vcache work.

### DIFF
--- a/software/spmd/uncached_io_test/Makefile
+++ b/software/spmd/uncached_io_test/Makefile
@@ -1,0 +1,17 @@
+bsg_tiles_X ?= 16
+bsg_tiles_Y ?= 8
+
+
+all: main.run
+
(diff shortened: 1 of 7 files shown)


taylor-bsg:
@Yuan-Mao can you resubmit this pull request against main?

PR #743, opened 2025-03-04
Sets sense to 1 in init for amoadd barrier

mrutt92:
Before this fix: if a cuda kernel invokes the barrier an odd number of times when using the amoadd barrier, then a second kernel launch will hang since barrier is always initialized to zero by the host.

After this fix: the sense is initialized to 1 and the barrier is set to 0.


tommydcjung (approved):
okay looks good!

PR #747, opened 2025-05-17
Allow llvm installation with gcc11

gaozihou:
Allow llvm installation with gcc11. This would be useful for OS that does not support devtoolset-8 (like Rocky 9 with gcc11 as default)

Tested with gcc 9.3.1, 10.2.1, 11.2.1 and 11.5.0 (llvm installation finished)


dpetrisko (approved):
LGTM!

PR #751, opened 2025-06-04
Some fixes

Yuan-Mao:
Fix incorrect flit count in nonsynth mem for uncached IO operations. This PR only changes nonsynth part of RTL.
This change allows passing the uncached IO test with 16 tiles enabled.
@tommydcjung  @gaozihou  Thanks.


gaozihou (commented):
I confirm this fixes the test hanging problem. @Yuan-Mao Please update the spmd software to enable 4x4 tiles by default

gaozihou (approved):

PR #755, opened 2025-06-06
pc_profilier sw fix

tommydcjung:
Fixing the issue reading out-of-range value from the array 'expectation'


gaozihou (approved):
I confirm this is an effective fix

PR #762, opened 2025-06-16
Fix address unit

Yuan-Mao:
The previous RTL assumes addr_r is word address. However, it is actually byte address.

Also the total IO address space should be 29-bit.

@tommydcjung @gaozihou Thanks

### DIFF
--- a/testbenches/common/v/mini_testbench.sv
+++ b/testbenches/common/v/mini_testbench.sv
@@ -257,7 +257,7 @@ module mini_testbench
     ,.wh_flit_width_p(wh_flit_width_p)
     ,.wh_cord_width_p(wh_cord_width_p)
     ,.wh_len_width_p(wh_len_width_p)
-    ,.mem_size_p((2**29))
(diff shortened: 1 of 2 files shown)


testbenches/common/v/mini_testbench.sv
```diff
@@ -257,7 +257,7 @@ module mini_testbench
     ,.wh_flit_width_p(wh_flit_width_p)
     ,.wh_cord_width_p(wh_cord_width_p)
     ,.wh_len_width_p(wh_len_width_p)
-    ,.mem_size_p(2*mem_size_lp)
+    ,.mem_size_p((2**29))
```
gaozihou: Maybe ```2**vcache_addr_width_p```

gaozihou:
I think this should go onto the RC2 branch as well, since the RC1 has already been signed off and frozen

testbenches/common/v/mini_testbench.sv
Yuan-Mao: Fixed!

PR #763, opened 2025-06-16
Miniblade rc2

Yuan-Mao:
The major changes is in the last commit. The first two commits are unmerged fixes from RC1.

In this PR, the uncached IO address has the following format: {vcache y cord (1 bit), vcache x cord (2 bits), IO address}. With that, programmers can now use the 3 MSBs to decide which Vcache the IO request would go to.

@tommydcjung @gaozihou Thanks!

### DIFF
--- a/v/vanilla_bean/network_tx.sv
+++ b/v/vanilla_bean/network_tx.sv
@@ -122,7 +122,7 @@ module network_tx
 //    ? {(pod_y_cord_width_p)'(pod_y_i + 1'b1), {y_subcord_width_lp{1'b0}}}
 //    : {(pod_y_cord_width_p)'(pod_y_i - 1'b1), {y_subcord_width_lp{1'b1}}};
 
-  wire [x_cord_width_p-1:0] proxy_x_cord_lo = {pod_x_i, io_addr[30:29]};
+  wire [x_cord_width_p-1:0] proxy_x_cord_lo = {pod_x_i, io_addr[30-:x_subcord_width_lp]};
   wire [y_cord_width_p-1:0] proxy_y_cord_lo = proxy_is_in_the_south
     ? {(pod_y_cord_width_p)'(pod_y_i + 1'b1), {y_subcord_width_lp{1'b0}}}
     : {(pod_y_cord_width_p)'(pod_y_i - 1'b1), {y_subcord_width_lp{1'b1}}};
@@ -177,7 +177,7 @@ module network_tx
     if (is_uncached_op) begin
       out_packet.y_cord = proxy_y_cord_lo;
       out_packet.x_cord = proxy_x_cord_lo;
-      out_packet.addr = (addr_width_p)'(io_addr[28:2]);
+      out_packet.addr = (addr_width_p)'(io_addr[30-x_subcord_width_lp:2]);
     end
     else begin
       out_packet.y_cord = y_cord_lo;
--- a/v/vanilla_bean/network_tx.sv
+++ b/v/vanilla_bean/network_tx.sv
@@ -112,15 +112,21 @@ module network_tx
   wire [x_cord_width_p-1:0] src_x_cord = {pod_x_i, my_x_i};
   wire [y_cord_width_p-1:0] src_y_cord = {pod_y_i, my_y_i};
 
-  wire proxy_is_in_the_south = src_y_cord[y_subcord_width_lp-1];
+  wire [31:0] io_addr = remote_req_i.addr;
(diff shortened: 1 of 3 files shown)


gaozihou:
Resolves #762

v/vanilla_bean/network_tx.sv
```diff
@@ -112,15 +112,21 @@ module network_tx
...
+  wire proxy_is_in_the_south = io_addr[31];
 
   // Uncached requests will be sent to the nearest vcache
-  wire [x_cord_width_p-1:0] proxy_x_cord_lo = (x_cord_width_p)'(src_x_cord);
+//  wire [x_cord_width_p-1:0] proxy_x_cord_lo = (x_cord_width_p)'(src_x_cord);
+//  wire [y_cord_width_p-1:0] proxy_y_cord_lo = proxy_is_in_the_south
+//    ? {(pod_y_cord_width_p)'(pod_y_i + 1'b1), {y_subcord_width_lp{1'b0}}}
+//    : {(pod_y_cord_width_p)'(pod_y_i - 1'b1), {y_subcord_width_lp{1'b1}}};
+
+  wire [x_cord_width_p-1:0] proxy_x_cord_lo = {pod_x_i, io_addr[30:29]};
```
gaozihou: Should be `io_addr[30:30-x_subcord_width_lp+1]`

tommydcjung: `io_addr[30-:x_subcord_width_lp]`

v/vanilla_bean/network_tx.sv
```diff
@@ -171,7 +177,7 @@ module network_tx
     if (is_uncached_op) begin
       out_packet.y_cord = proxy_y_cord_lo;
       out_packet.x_cord = proxy_x_cord_lo;
-      out_packet.addr = io_addr;
+      out_packet.addr = (addr_width_p)'(io_addr[28:2]);
```
gaozihou: Should be `out_packet.addr = io_addr[addr_width_p+1:2];`

tommydcjung: `io_addr[30-x_subcord_width_lp:2]`

gaozihou: @tommydcjung It's better to send the full `addr_width_p` bits to the vcache, the testbench is not using x-cord and cid to reconstruct the address

tommydcjung: okay I see, so the IO address space limited to be 28-bit word address?

gaozihou: For now it is 27-bit word address (since vcache has 29-bit byte address, the MSB is dropped in the link_to_cache adapter). In the future we may extend it in the testbench

tommydcjung: although if x_subcord_width_p was wider, it would overlap with io_addr. From the programmer's perspective, they would be interfaced like `bsg_uncached_store(bank_id, io_addr)`, it might  be more natural to make them not overlap...

gaozihou: @tommydcjung Oh I get your point. Yeah if we don't want overlap then `io_addr[30-x_subcord_width_lp:2]` is better, and we can set the mini_testbench io mem size to ```2**(32-1-`BSG_SAFE_CLOG2(num_tiles_x_p))```

tommydcjung: also kinda related, we might wanna reorder the order of if-else:
```
if (addr == 11xxxx... && ~uncached) begin
    // maps to wh_dest and notification_en registers;
end
else if (addr == 1xxxxx... && ~uncached) begin
    // maps to vcache tags;
end
else begin
    // everything else
end

-----------------------------------------------------

if (uncached) begin
  // uncached op;
end
else begin
    if (addr == 11xxxx...) begin
      // maps to wh_dest and notification_en registers;
    end
    else if (addr == 1xxxxx...) begin
      // maps to vcache tags;
    end
    else begin
      // everything else;
    end
end
```

Maybe the below approach is better? Something to think about...

tommydcjung: Well maybe this doesn't solve the problem of io addr space being restricted by vcache_addr_width_p. I guess the main problem is that vcache_addr_width_p is tied with the vcache tag width, but uncached/IO operation does not need to be tied to that. Adding a new parameter uncached_addr_width_p sounds like a more work though. So maybe we should keep the way it is now.

tommydcjung (approved):

PR #768, opened 2026-09-05
Support non-x86 hosts in the 128-bit DPI packet wrapper

taylor-bsg:
## Summary

The DPI wrapper currently exposes every 128-bit packet as the Intel intrinsic type `__m128i` and includes `xmmintrin.h` unconditionally, which prevents it from compiling on Apple Silicon.

This change keeps `__m128i` on x86 and uses an aligned two-word container on other hosts. The DPI FIFO copies packet bytes and does not require SIMD operations. A static assertion preserves the 16-byte ABI.

## Validation

- Compiled in the complete Replicant host runtime with Apple Clang 21 on arm64
- Elaborated and linked a 492-module 4-by-2 HammerBlade model with Verilator v5.050
- Passed the 4096-word memcpy execution and profiling runs
- Rebased cleanly onto the current `master` branch

## AI assistance

Prepared with OpenAI Codex assistance. The submitter remains responsible for review and validation.

### DIFF
--- a/testbenches/dpi/bsg_nonsynth_dpi_manycore.hpp
+++ b/testbenches/dpi/bsg_nonsynth_dpi_manycore.hpp
@@ -9,9 +9,17 @@
 #include <svdpi.h>
 #include <cstring>
 #include <cstdint>
-// We use __m128i so that we can pass a 128-bit type between Verilog
-// and C.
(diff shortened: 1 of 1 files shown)


taylor-bsg:
Simple change.

